PR #10, opened 2022-08-15
Support for passive servers + imx8 hardware

lucypa:
The below commits enable a pd to be "passive" by signalling the monitor to then remove a pd's scheduling context and bind it with that pd's notification objects. 
Users can specify a pd is passive by setting adding a boolean to the system xml. In order for the passive server to run, it needs a notification object and cap. 
In order to properly initialise, the passive server requires a scheduling context to start with. Once initiliased, the passive server then signals to the monitor that initialisation is complete, and the monitor will then unbind the passive server's scheduling context, and instead bind it to the passive server's notification cap. This is required so that the passive server will be scheduled when it gets a notification.


tool/sel4coreplat/__main__.py
```diff
@@ -1486,6 +1503,7 @@ def build_system(
     for pd in system.protection_domains:
         # Could use pd.elf_file.write_symbol here to update variables if required.
         pd_elf_files[pd].write_symbol("sel4cp_name", pack("<16s", pd.name.encode("utf8")))
+        pd_elf_files[pd].write_symbol("passive", pack("?", pd.passive))
```
bennoleslie, line 1506: What does `?` do in this context?

example/imx8mm/passive_server/client.c
```diff
@@ -0,0 +1,20 @@
...
+
+    /* message the server */
+    sel4cp_mr_set(0, 0);
+    (void) sel4cp_ppcall(SERVER_CH, sel4cp_msginfo_new(1, 1));
+}
+
+void
+notified(sel4cp_channel ch)
+{
+    sel4cp_dbg_puts("Client recieved a notification on an unexpected channel\n");
```
bennoleslie: please make the debug output consistent suggest:

```client: PD init function invoked```

and

```client: received a notification on an unexpected channel\n```

File: example/imx8mm/passive_server/passive_server.system [hunk 3]
```diff
@@ -0,0 +1,21 @@
...
+    <protection_domain name="client" priority="99">
+        <program_image path="client.elf" />
+    </protection_domain>
+
+    <channel>
+        <end pd="server" id="0" />
+        <end pd="client" id="0" />
+    </channel>
+
+</system>
```
bennoleslie: Add a newline at end of line for consistency for other system files.

example/imx8mm/passive_server/server.c
```diff
@@ -0,0 +1,29 @@
+
+#include <sel4cp.h>
+
+sel4cp_msginfo 
+protected(sel4cp_channel ch, sel4cp_msginfo msginfo)
+{
+    switch (sel4cp_msginfo_get_label(msginfo)) {
+        case 0:
+            sel4cp_dbg_puts("Server is running on clients scheduling context\n");
```
bennoleslie: similar to comment on the `client` side. Please use `server:` consistently for the printing.

bennoleslie (changes requested):
There are some minor comments, but overall this looks good.

Overall there are a few other things that need to be done:

1. The `manual.md` needs to be update with the new interfaces.
2. I think that merge request needs to be a bit more descriptive of the implementation. It is mostly self-explanatory, but the MR will be there for people learning in the future, so if possible describe the changes to the implementation in some detail. Something like:

> The passive server needs a notification object and cap because .....
> The passive server has the monitor unbind the scheduler context at time x so that ...
> The bootstrap size is increase from 90 -> 120 because ...

Should take no more than an hour, but will be great help for others in the future.

I don't want to change this MR at this point, but in future I think it would be ideal so split the "add hardware support" and "add new feature" changes in to separate MRs.

(Of course, relies on me doing MR reviews in a much more timely manner, so understand why these are batched up right now!).

tool/sel4coreplat/__main__.py
lucypa: `?` is for a boolean type with a size of a single byte (as per https://docs.python.org/3/library/struct.html#format-characters )

PR #11, opened 2022-08-19
Benchmarking

lucypa:
This PR enables users to specify a 'benchmark' config in order to use seL4's benchmarking system calls.

### DIFF
--- a/build_sdk.py
+++ b/build_sdk.py
@@ -72,6 +72,19 @@ SUPPORTED_BOARDS = (
         examples = {
             "hello": Path("example/zcu102/hello")
         }
+    ),
+    BoardInfo(
+        name="imx8mm",
+        gcc_cpu="cortex-a53",
+        loader_link_address=0x41000000,
+        kernel_options = {
+            "KernelPlatform": "imx8mm-evk",
+            "KernelIsMCS": True,
+            "KernelArmExportPCNTUser": True,
+        },
+        examples = {
+            "ethernet": Path("example/imx8mm/passive_server")
+        }
     )
 )
 
--- a/build_sdk.py
+++ b/build_sdk.py
@@ -81,6 +81,7 @@ SUPPORTED_BOARDS = (
             "KernelPlatform": "imx8mm-evk",
             "KernelIsMCS": True,
             "KernelArmExportPCNTUser": True,
+            "KernelArmExportPMUUser": True,
         },
         examples = {
             "ethernet": Path("example/imx8mm/passive_server")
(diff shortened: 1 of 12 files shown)


tool/sel4coreplat/__main__.py
```diff
@@ -1460,6 +1462,10 @@ def build_system(
     for tcb_obj, schedcontext_obj, pd in zip(tcb_objects, schedcontext_objects, system.protection_domains):
         system_invocations.append(Sel4TcbSetSchedParams(tcb_obj.cap_addr, INIT_TCB_CAP_ADDRESS, pd.priority, pd.priority, schedcontext_obj.cap_addr, fault_ep_endpoint_object.cap_addr))
 
+    # Copy the PD's TCB cap into their address space for development purposes.
```
bennoleslie, line 1465: I *think* that this is only needed for development purposes, and following POLA we'd ideally avoid having this cap available in production.

We probably need to think about the best way to support development cap needs vs production cap needs. Could be topic for a Monday meeting discussion. I'll think some more about it.

build_sdk.py
```diff
@@ -102,7 +111,7 @@ class ConfigInfo:
             "KernelPrinting": True,
             "KernelVerificationBuild": False
         }
-    ),
+    )
```
bennoleslie, line 114: Let's avoid unrelated changes. In this case the trailing comma was intentional.

What we will do here in future is use: https://black.readthedocs.io/en/stable/index.html to auto-format all the Python code, which should avoid any of these issues.

bennoleslie (changes requested):
Overall need to think about how we have benchmarking builds and changes.

Code is very good, no issues there. We just need to think about bigger picture.
I really don't like the idea of requiring different builds for benchmarking vs production. 

I think it leads to potential (unintentional) cheating / fooling ourselves by having potential different code running for benchmarks vs production, which I think we should be able to find a good alternative to.

More general comments: with an MR please try to make the MR title a bit longer than just "Benchmarking", and then provide a bit more detail in the MR description.

for example here it could be something like:

"Add support to enable system benchmarking"

and the description would be something like:

> Add a benchmark kernel that includes xxxx so that yyyy.
> Add TCB cap to PD to enable the code to call xxx so that we get benchmark yyyy.

We should also update the manual appropriately as well.

Let's get the passive server MR complete first, and then rebase this branch off master once the passive server change is merged.

(And apologies, this is my fault for not being able to review in a more timely manner).

bennoleslie:
Please review conflicts (ideally rebase from master and push)

Ivan-Velickovic:
Re-done this work in https://github.com/seL4/microkit/pull/157 since lots has changed since this PR.

PR #17, opened 2023-03-02
Fix order of attributes and rights fields in Sel4PageMap invocation

Ivan-Velickovic:
The `seL4_ARM_Page_Map` invocation takes in the page attributes after the page rights. The seL4CP tool that creates the invocations has the correct order so it does not need to change, but because the invocation's fields in the wrong order, the actual invocations that were being generated were wrong.


bennoleslie:
Good find!

PR #20, opened 2023-03-02
Fix value of seL4_MaxUntypedBits when emulating kernel boot

Ivan-Velickovic:
I believe the variable `max_bits` corresponds to `seL4_MaxUntypedBits`. The value of this on AArch64 is 47 (see `seL4/libsel4/sel4_arch_include/aarch64/sel4/sel4_arch/constants.h`.

I have not run into issues because this value is not correct yet, but I thought I would try to fix it before having to debug something going wrong.

### DIFF
--- a/tool/microkit/util.py
+++ b/tool/microkit/util.py
@@ -66,7 +66,7 @@ class MemoryRegion:
     end: int
 
     def aligned_power_of_two_regions(self) -> List["MemoryRegion"]:
-        max_bits = 40
+        max_bits = 47
         # Align
         # find the first bit self
         r = []


bennoleslie:
I need to spend a little more time on this one. (Very simple change!).

I'm not 100% sure that it is intended to match `seL4_MaxUntypedBits` (it probably is, but I just want to check, from memory there is a few different things going on there). 

Ideally these constants come from a kernel build process artefact that would specify all of these in JSON or something.

Ivan-Velickovic:
I think this code is trying to emulate `create_untypeds_for_region`.

I agree that these constants should come from the kernel build output. This isn't currently done by the kernel, but I'll look into changing it to do so.

PR #32, opened 2023-05-24
Add option to declare IRQ as "edge" triggered

Ivan-Velickovic:
Currently all interrupts are registered as level-triggered, this patch enables users to specify a interrupt in the system description as edge-triggered instead.

One alternative solution would be to look at the device tree for the interrupt trigger type instead.


Ivan-Velickovic:
One issue I've just found with this current patch is that it assumes all ARM platforms have the kernel configuration `HAVE_SET_TRIGGER` set which is not true.

Ivan-Velickovic:
This PR can be completed once https://github.com/seL4/seL4/issues/1073 is resolved.

PR #33, opened 2023-05-24
Add error checking for write-only memory region mappings

Ivan-Velickovic:
On all architectures that seL4 currently supports, write-only mappings are not allowed. It is preferable to catch these at build-time rather than having an error at run-time.


bennoleslie:
LGTM.

PR #36, opened 2023-06-23
Add notes about the budget and interrupts

podhrmic:
Clarifications.

### DIFF
--- a/docs/manual.md
+++ b/docs/manual.md
@@ -158,7 +158,36 @@ A budget that equals the period (aka. a "full" budget) behaves like a traditiona
 The **priority** determines which of the runnable PDs to schedule. A PD is runnable if one of its entry points have been invoked and it has budget remaining in the current period.
 Runnable PDs of the same priority are scheduled in a round-robin manner.
 
-The **passive** determines whether the PD is passive. A passive PD will have it's scheduling context revoked after initialisation and then bound instead to the PD's notification object. This means the PD will be scheduled on receiving a notification, whereby it will run on the notification's scheduling context, or when the PD receives a *protected procedure* by another PD, whereby the passive PD will run on the scheduling context of the callee. 
+The **passive** determines whether the PD is passive. A passive PD will have it's scheduling context revoked after initialisation and then bound instead to the PD's notification object. This means the PD will be scheduled on receiving a notification, whereby it will run on the notification's scheduling context, or when the PD receives a *protected procedure* by another PD, whereby the passive PD will run on the scheduling context of the callee.
+
+If a PD runs out of its budget, it is preempted and resumed the next time period. An example of a task like that is below - the `for` loop will be preempted several times, depending on your budget:
+
+```C
+void
+notified(sel4cp_channel ch)
+{
+	switch (ch) {
+		case START_BIG_TASK: {
+				sel4cp_dbg_puts("Starting a big task\n");
+				int i = 0;
+				for (i = 0; i<100000000; i++) {
+					if (i % 10000000 == 0) {
+						sel4cp_dbg_puts(".");
+					}
+					asm("nop");
+				}
+				sel4cp_dbg_puts("Big task done\n");
+				break;
+			}
+		default: {
+			sel4cp_dbg_puts("received notification on unexpected channel\n");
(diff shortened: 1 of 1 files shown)


docs/manual.md
```diff
@@ -158,7 +158,36 @@ A budget that equals the period (aka. a "full" budget) behaves like a traditiona
...
+			}
+		default: {
+			sel4cp_dbg_puts("received notification on unexpected channel\n");
+			break;
+			}
+		}
+}
+```
+
+On the other hand, if a PD has `budget` available, but receives no notification, it is not run at all. Returning from `notified()` function and calling `seL4_Yield()` during the PD execution has the same effect, and donates the remaining budget of the PD.
```
Ivan-Velickovic, line 190: "Returning from `notified()` function and calling `seL4_Yield()` during the PD execution has the same effect, and donates the remaining budget of the PD." Has the same effect as what exactly? I'm not sure what you mean by calling `seL4_Yield` after returning from the notified entry point?

docs/manual.md
```diff
@@ -263,6 +292,8 @@ sel4cp does not provides timers, nor any *sleep* API.
 After initialisation, activity in the system is initiated by an interrupt causing a `notified` entry point to be invoked.
 That notified function may in turn notify or call other protection domains that cause other system activity, but eventually all activity indirectly initiated from that interrupt will complete, at which point the system is inactive again until another interrupt occurs.
 
+An interrupt must be acknowledged with `sel4cp_irq_ack()` in the `notified()` function, otherwise it will not be fired again.
```
Ivan-Velickovic: An interrupt doesn't necessarily have to be acknowledged in the `notified()` function. My understanding is that the interrupt will continue to be fired, it just won't be delivered to seL4. Perhaps it would be better to say "otherwise the interrupt will not be delivered again".

Ivan-Velickovic:
Just a general question I have is how did you interpret seL4-specific terminology such as "scheduling context" which isn't really explained in the seL4CP manual? Did you refer to the seL4 manual or are you already familiar with the concepts from previous seL4 experience?

docs/manual.md
podhrmic: This is based on the discussion [here](https://github.com/au-ts/camkes_to_sel4cp_guide/issues/1#issuecomment-1507922309). In my understanding, you can either return from `notified()` or call `seL4_Yield()` somewhere in your `notified()` function, and that will end the execution of that particular PD in the given period. Is that accurate?

docs/manual.md
podhrmic: Good point, I updated the wording.

podhrmic:
> Just a general question I have is how did you interpret seL4-specific terminology such as "scheduling context" which isn't really explained in the seL4CP manual? Did you refer to the seL4 manual or are you already familiar with the concepts from previous seL4 experience?

I read the seL4 manual, and have some previous experience, but I am by no means an seL4 expert.

docs/manual.md
Ivan-Velickovic: While seL4CP doesn't stop anyone from doing this in their user-code, I am kind of hesitant to have this in the manual as I don't think there should be any need to call `seL4_Yield` and having a non-event driven PD seems to go against the principles of seL4CP.

Ivan-Velickovic:
> > Just a general question I have is how did you interpret seL4-specific terminology such as "scheduling context" which isn't really explained in the seL4CP manual? Did you refer to the seL4 manual or are you already familiar with the concepts from previous seL4 experience?
> 
> I read the seL4 manual, and have some previous experience, but I am by no means an seL4 expert.

Cool, was just wondering. We should probably link to the seL4 manual or explain what a scheduling context is in the seL4CP manual but that can be done in another PR.

podhrmic:
@Ivan-Velickovic I am happy to close this PR - the existing documentation and tutorial is IMHO pretty good, if there is a need to further explain scheduling, perhaps a separate dedicated chapter with examples is better.

Ivan-Velickovic:
There are some useful changes here, I just think the code example is not necessary. You can close the PR and I can take your patch and adjust in a new PR, if that's okay with you.

PR #38, opened 2023-08-16
Added maaxboard support with hello example

josh-felm:


build_sdk.py
```diff
@@ -85,7 +85,20 @@ class ConfigInfo:
...
     #         "passive_server": Path("example/imx8mm/passive_server")
     #     }
-    # )
+    # ),
+    BoardInfo(
+        name="maaxboard",
+        gcc_cpu="cortex-a53",
+        loader_link_address=0x40480000,
+        kernel_options = {
+            "KernelPlatform": "imx8mq-evk",
```
Ivan-Velickovic: Shouldn't this be `maaxboard`? I assume that you've put it as `imx8mq-evk` as the version of seL4 that seL4CP uses doesn't have `maaxboard`. The solution is to update the the seL4 version used by seL4CP then.

However, if the `maaxboard` works with `"KernelPlatform": "imx8mq-evk"`, why does it even exist as a separate platform in seL4?

axel-h: Basically there is the Core, the SoC and the Board (and then also the board variant ....) that you can target. If board use the same SoC, the binaries might work on the other boards also more or less But it's good practice to really select the most specific target at the moment. As far as I know, having generic images that are configured at runtime is not a goal the moment for the kernel. Not sure how the micokit plans to handle this, though.

Ivan-Velickovic:
A couple requests:
* There seems to be two entries for the maaxboard in the `SUPPORTED_BOARDS` list now?
* Probably just have the hello world example for the maaxboard.
* The example code uses seL4CP instead of Microkit, if you just do `cp -r example/zcu102/hello examples/maaxboard/hello` and commit that, it should update everything...

Sorry for the confusion regarding the seL4 branch right now. I'll update the `microkit` branch and so it'll contain the MaaXBoard support and then we can merge this.

loader/src/loader.c
```diff
@@ -154,6 +154,8 @@ putc(uint8_t ch)
 {
     while (!(*UART_REG(STAT) & STAT_TDRE)) { }
     *UART_REG(TRANSMIT) = ch;
+    if (ch == '\n')
+        *UART_REG(TRANSMIT) = '\r';
```
axel-h: This will print "\n\r", but actually it should be "\r\n" that gets printed. This is a common bug in many UART drivers from seL4, we should really stop spreading it further.

loader/src/loader.c
```diff
@@ -153,9 +153,9 @@ static void
 putc(uint8_t ch)
 {
     while (!(*UART_REG(STAT) & STAT_TDRE)) { }
-    *UART_REG(TRANSMIT) = ch;
     if (ch == '\n')
         *UART_REG(TRANSMIT) = '\r';
+    *UART_REG(TRANSMIT) = ch;
```
axel-h, line 163: The UART output is much much slower than the core executing instructions. Thus you have to check here again that FIFO has space in case you wrote the `r` there. I'm not sure what the `STAT_TDRE` tells you exactly, usually it that the FIFO has space for at least one char and not that the FIFO is empty and thus can take more than one char. So better use this:
```
    if (ch == '\n') {
        // ensure FIFO has space
        while (!(*UART_REG(STAT) & STAT_TDRE)) { }
        *UART_REG(TRANSMIT) = '\r';
    }     
    
    // ensure FIFO has space
    while (!(*UART_REG(STAT) & STAT_TDRE)) { }
    *UART_REG(TRANSMIT) = ch;
```

josh-felm: Thanks for this

Bill-James-Ellis:
Hi, as per, https://github.com/seL4/seL4/pull/1164 is this pull request effectively blocked until sel4 (kernel) "microkit" branch is updated to have maxxboard support, as seen on the sel4 (kernel) master branch? (Is there a link or pull-request for that dependency change?)

Ivan-Velickovic:
> Hi, as per, [seL4/seL4#1164](https://github.com/seL4/seL4/pull/1164) is this pull request effectively blocked until sel4 (kernel) "microkit" branch is updated to have maxxboard support, as seen on the sel4 (kernel) master branch? (Is there a link or pull-request for that dependency change?)

That's correct. This has been done in https://github.com/seL4/microkit/pull/93.

There were a couple things that needed changing, some minor style fixes but mainly the documentation in the manual for how to boot the image and also that the `putc` function didn't work initially, so I've fixed that myself.

Were the changes tested? Just because I couldn't get the hello world to work without making the change:
```diff
- #define TRANSMIT 0x49
+ #define TRANSMIT 0x40
```

Thanks for your patience.

Bill-James-Ellis:
Hi, I'm pretty sure Josh tested this, but I'm not sure when, and in what configuration.

I have tried building from scratch with the merged code, using the head of main branch of microkit repo, and the documented commit (7008430d4432c71a74b2a1da0afae58f7a8658df) of microkit branch of sel4 repo, and it worked for me! I got "hello, world" over the serial TTY.

Note I used "debug" for MICROKIT_CONFIG -- I think the printing "hello" bit might depend on the kernel provided serial driver, which is only present in debug?

Note I can't find the "TRANSMIT 0x49" line anywhere? Do you know where that came from? The pull request, and our separate sel4-cap branch, all seem to have "TRANSMIT 0x40". Since the merged "0x40" seems to work, I guess that's fine, I'm just curious if there's another pull-request/branch/something I'm missing?

Ivan-Velickovic:
Glad to hear it works for you as well!

> Note I can't find the "TRANSMIT 0x49" line anywhere?

Weird, I looked at the original PR changes and `0x49` is not there, although when I force-pushed you can see in the diff at some point it was `0x49` https://github.com/seL4/microkit/compare/09003055e48d17f88725152c2922a49162959d4f..1dff9010c8f699f2575c5695135dab4999a8e78f. Not quite sure what's going on, but at least it works now.

PR #41, opened 2023-09-13
The big rename (seL4 Core Platform -> seL4 Microkit)

Ivan-Velickovic:
The seL4 Core Platform project is being renamed to seL4 Microkit. This is a breaking change and will affect **all** current users of Microkit.

**Why is this happening?**

The seL4 Core Platform name, especially the "Core" part, indicates to users of seL4 that it is *the* framework to build systems with on seL4. This is not true. It is a collection of tools, abstractions and a library to build *certain kinds* of systems on top of seL4, it is definitely not the only way to build on seL4.

**How to refer to seL4 Microkit**

In an seL4 context, it is referred as simply "Microkit". In a non-seL4 context it is referred to as "seL4 Microkit".

In code, it is referred to as `microkit`.

**Affects to users**

If you are using the Microkit already, the main renames to be aware of are the following:
* The library is now called `libmicrokit` instead of `libsel4cp`. This means that when you include the library the header is now `microkit.h` instead of `sel4cp.h`.
    * Subsequently, all library API functions are prefixed with `microkit_` instead of `sel4cp_`.
* The tool is now called `microkit` instead of `sel4cp`.

Apologies for any inconvenience, but the earlier we do the rename, the better.


docs/README.md
```diff
@@ -19,23 +19,23 @@ The initial images will need to be packed into the initial tasks image, and then
...
-## seL4 Core Platform Library
+## Microkit Library
 
-The header and source files for the library are in the `libsel4cp` directory.
+The header and source files for the library are in the `libmicrokit` directory.
 
 To compile against the library you need flags such as:
 
-    -I$(LIBSEL4CP)/include -L$(LIBSEL4CP) -llibsel4cp
+    -I$(LIBMICROKIT)/include -L$(LIBMICROKIT) -llibmicrokit
```
wom-bat, line 28: `-l` prepends `lib`, so it should be -lmicrokit

example/imx8mm/passive_server/server.c
```diff
@@ -18,12 +18,12 @@ protected(sel4cp_channel ch, sel4cp_msginfo msginfo)
...
+    microkit_dbg_puts("server: server protection domain init function running\n");
     /* Nothing to initialise */
 }
 
 void
-notified(sel4cp_channel ch)
+notified(microkit_channel ch)
 {
-    sel4cp_dbg_puts("server: recieved a notification on an unexpected channel\n");
+    microkit_dbg_puts("server: recieved a notification on an unexpected channel\n");
```
wom-bat, line 28: For later: fix spelling of 'received'

wom-bat (approved):
Looks generally good.  Not sure that 'mk_' wouldn't have made a nicer prefix in code; and I spotted a couple of issues in teh manual that will need a separate fix for.

libmicrokit/include/microkit.h
```diff
@@ -3,7 +3,8 @@
  *
  * SPDX-License-Identifier: BSD-2-Clause
  */
-/* seL4 Core Platform interface */
+
```
wom-bat, line 6: Here's your extra insertion btw.

docs/README.md
Ivan-Velickovic: Yes good point, I will make a separate PR for fixes as this is supposed to just be a straight find and replace kind of patch.

I've noted some other issues I found as well which can be in the same subsequent PR.

example/imx8mm/passive_server/server.c
Ivan-Velickovic: Will make a note of it thanks.

PR #50, opened 2023-09-13
Use specific AArch64 toolchain when compiling seL4

Ivan-Velickovic:
The seL4 build system is configured to find installations of compilers on the host, meaning that the `aarch64-none-elf-` toolchain that we use might not be used when building seL4 depending on what other toolchains are installed.

### DIFF
--- a/build_sdk.py
+++ b/build_sdk.py
@@ -29,6 +29,10 @@ ENV_BIN_DIR = Path(executable).parent
 
 MICROKIT_EPOCH = 1616367257
 
+# Due to only supporting AArch64 (for now), we hard-code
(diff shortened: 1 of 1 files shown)


lsf37 (approved):
Compatible tool chain selection is one of the things we should figure out in the medium term, but no issues with this in the short term.

Ivan-Velickovic:
> Compatible tool chain selection is one of the things we should figure out in the medium term, but no issues with this in the short term.

Yes, this is just to make sure that we’re using the same toolchain for all components of Microkit as well as seL4. We can definitely change it later on if necessary.

PR #54, opened 2023-09-13
basic foundation CI setup

lsf37:
This PR adds the basic standard checks for foundation repos:

- licenses
- trailing whitespace
- gitlint
- code style
- portable shell check
- link check (for links in .md and .html files)

It also adds missing license headers and texts so that the licence check can succeed on new pull requests.

The style check is likely to fail initially if the repo hasn't used the same `astyle` config before. We can leave that check as optional for a while of that leads to friction.

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -1,0 +1,34 @@
+# Copyright 2020, Data61, CSIRO (ABN 41 687 119 230)
+#
+# SPDX-License-Identifier: BSD-2-Clause
+
+# Actions to run on pull requests
+
+name: PR
+
+on: [pull_request]
+
+jobs:
+  gitlint:
+    name: Gitlint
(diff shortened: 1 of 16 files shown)


Ivan-Velickovic (approved):

Ivan-Velickovic:
Only comment I have is that *somewhere* in the repository, it should show you how to run the checks yourself. E.g a README in `.github`

Ivan-Velickovic:
For the CI tests I was going to add a `ci` directory, if we do that then the instructions for the style checks etc could also go there.

lsf37:
> For the CI tests I was going to add a `ci` directory, if we do that then the instructions for the style checks etc could also go there.

That's a good idea, and we should probably add general instructions on how to run those checks somewhere central as well (and maybe then link from here or something like that).

PR #57, opened 2023-09-13
LICENSE file adjustments

lsf37:
- adjust build_sdk for file rename
- adjust license file text so that it works for both repo and SDK

### DIFF
--- a/LICENSE.md
+++ b/LICENSE.md
@@ -5,13 +5,13 @@
 
 # License
 
-The files in this repository are released under standard open source licenses,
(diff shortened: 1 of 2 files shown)


Ivan-Velickovic:
What about the `LICENSES` directory?

lsf37:
> What about the `LICENSES` directory?

Leaving that one for a separate PR, because I'm currently not set up to test the script properly. But it should also be copied into the SDK, yes.

Ivan-Velickovic (approved):

PR #60, opened 2023-09-14
monitor: adhere to CI style checks

Ivan-Velickovic:
I don't agree with these style checks (particularly the switch case style) but I don't have the energy to fight the system on something that is ultimately meaningless.


Ivan-Velickovic:
Cool, they are the same.

PR #62, opened 2023-09-14
libmicrokit: fix to successfully link on LLD

Ivan-Velickovic:
LLD does not recognise the STARTUP directive which works with the GNU linker. Given that the crt0.s entry point is already specified to be at the start of the _text section when linking with libmicrokit, all we need to do is slightly change the Makefile to also include crt0.o when linking to produce libmicrokit.a

### DIFF
--- a/build_sdk.py
+++ b/build_sdk.py
@@ -312,13 +312,6 @@ def build_lib_component(
     # Make output read-only
     dest.chmod(0o444)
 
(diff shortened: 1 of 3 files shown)


Ivan-Velickovic:
Should have an example that shows using LLVM tools to compile PD ELFs for both education reasons and as well as testing that it actually works.

PR #64, opened 2023-09-25
fix: set have_signal to false after NBSendRecv

isubasinghe:


Ivan-Velickovic (approved):
I’ll make a separate PR fixing the style checks before merging this.

We should also add comments about why variables like ‘passive’ are fine to be uninitialised in the code, but again that can be done separately.

PR #67, opened 2023-09-25
tool: adhere to CI style checks

Ivan-Velickovic:
Ran `seL4_tools/misc/style-all.sh` on the whole repository, so hopefully this is the last PR of this kind we have to make.

### DIFF
--- a/tool/microkit/__main__.py
+++ b/tool/microkit/__main__.py
@@ -108,7 +108,7 @@ from microkit.sel4 import (
     SEL4_OBJECT_TYPE_NAMES,
 )
 from microkit.sysxml import ProtectionDomain, xml2system, SystemDescription, PlatformDescription
-from microkit.sysxml import SysMap, SysMemoryRegion # This shouldn't be needed here as such
+from microkit.sysxml import SysMap, SysMemoryRegion  # This shouldn't be needed here as such
 from microkit.loader import Loader
 
 # This is a workaround for: https://github.com/indygreg/PyOxidizer/issues/307
@@ -120,8 +120,9 @@ if argv[0] is None:
 
 
 default_platform_description = PlatformDescription(
-    page_sizes = (0x1_000, 0x200_000)
+    page_sizes=(0x1_000, 0x200_000)
 )
+
 
 @dataclass
 class MonitorConfig:
@@ -135,6 +136,7 @@ class MonitorConfig:
     def max_untyped_objects(self, symbol_size: int) -> int:
         return (symbol_size - self.untyped_info_header_struct.size) // self.untyped_info_object_struct.size
 
+
 # The monitor config is fixed (unless the monitor C code
 # changes the definitions of struct, or the name.
 # While this is fixed, we dynamically determine the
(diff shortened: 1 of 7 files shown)


tool/microkit/__main__.py
```diff
@@ -1258,7 +1258,7 @@ def build_system(
 
     final_cap_slot = cap_slot
 
-    ## Minting in the address space
```
Ivan-Velickovic, line 1271: Will have to revert these kinds of changes since these comments are intentionally formatted as so.

Ivan-Velickovic:
The style script managed to break the SDK, interesting.

Ivan-Velickovic:
The style checker is now crashing, I am not sure why this is happening. The only thing I can think of is that it looks like the style checker uses Python 3.10 and Microkit uses Python 3.9.

lsf37:
It should be working fine with python 3.9, but I have seen the style checker crashing before on some code (long time ago, I don't remember the details).

Ivan-Velickovic:
The version of autopep8 that the CI uses is older so maybe it's a bug that has been fixed by autopep8 by now. I installed the latest version of autopep8 and that doesn't crash locally. I'll test the version in the container to see if I can at least reproduce the Ci failure locally.

Ivan-Velickovic:
I was able to reproduce it:
```
ivanv@in-container:/host$ autopep8 -i --max-line-length 100 tool/microkit/docs.md tool/microkit/elf.py tool/microkit/__init__.py tool/microkit/loader.py tool/microkit/__main__.py tool/microkit/sel4.py tool/microkit/sysxml.py tool/microkit/util.py
Traceback (most recent call last):
  File "/usr/local/bin/autopep8", line 8, in <module>
    sys.exit(main())
  File "/usr/local/lib/python3.9/dist-packages/autopep8.py", line 4161, in main
    ret = fix_multiple_files(args.files, args, sys.stdout)
  File "/usr/local/lib/python3.9/dist-packages/autopep8.py", line 4063, in fix_multiple_files
    ret = _fix_file((name, options, output))
  File "/usr/local/lib/python3.9/dist-packages/autopep8.py", line 4043, in _fix_file
    return fix_file(*parameters)
  File "/usr/local/lib/python3.9/dist-packages/autopep8.py", line 3423, in fix_file
    fixed_source = fix_lines(fixed_source, options, filename=filename)
  File "/usr/local/lib/python3.9/dist-packages/autopep8.py", line 3403, in fix_lines
    fixed_source = fix.fix()
  File "/usr/local/lib/python3.9/dist-packages/autopep8.py", line 587, in fix
    self._fix_source(filter_results(source=''.join(self.source),
  File "/usr/local/lib/python3.9/dist-packages/autopep8.py", line 531, in _fix_source
    modified_lines = fix(result)
  File "/usr/local/lib/python3.9/dist-packages/autopep8.py", line 717, in fix_e225
    pycodestyle.missing_whitespace_around_operator(fixed, ts))
AttributeError: module 'pycodestyle' has no attribute 'missing_whitespace_around_operator'
```

Ivan-Velickovic:
Okay I used a newer version of `autopep8` locally and pushed the changes it produced. Now the CI style check does not crash anymore but we have another problem which is that it wants to apply this diff:
```

diff --git a/tool/microkit/sysxml.py b/tool/microkit/sysxml.py
index 7bcf4ca..b231878 100644
--- a/tool/microkit/sysxml.py
+++ b/tool/microkit/sysxml.py
@@ -3,6 +3,7 @@
 #
 # SPDX-License-Identifier: BSD-2-Clause
 #
+import xml.etree.ElementTree as ET
 from microkit.util import str_to_bool, UserError
 from typing import Dict, Iterable, Optional, Set, Tuple
 from dataclasses import dataclass
@@ -11,7 +12,6 @@ from pathlib import Path
 # Force use of Python elementtree to avoid overloading
 import sys
 sys.modules['_elementtree'] = None  # type: ignore
-import xml.etree.ElementTree as ET
```

which will cause the tool to not compile/interpret because `import xml.etree.ElementTree as ET` is needed *after* the `sys.modules['_elementtree'] = None  # type: ignore` line.

lsf37:
Does the newer version also want this diff or only the CI version? I could look into upgrading the CI version.

Ivan-Velickovic:
> Does the newer version also want this diff or only the CI version? I could look into upgrading the CI version.

I think it's some Python PEP8 rule probably. So different versions are also enforcing it. Not sure what to do since I'm not that familiar with this stuff.

tool/microkit/__main__.py
Ivan-Velickovic: Seems that this is enforced and the style checker does not allow freedom in how you format comments so will have to keep these changes.

Ivan-Velickovic:
Sorted it out finally.

Issue #70, opened 2023-09-25
Syntax/Semantics for the system manifest


podhrmic:
There was a brief discussion at the summit that maybe using XML for the system description is not the best ([hjson](https://hjson.github.io/) anyone?:-) ). Either way, there should be proper grammar (EBNF?) and associated rules for the manifest language.

Why?
1. then switching between different manifest languages should be trivial
2. generating the manifest from other tools should be simpler (think SysMLv2/AADL -> microkit)
3. with proper rules we can ensure the manifest is well formed (I know you are already doing a number of checks, but we should be explicit about it)

Ivan-Velickovic:
In my opinion I think all the system description formats are going to be equally as bad. I have not noticed significant friction in using the current format except for large systems. However, I think any format would struggle when dealing with large systems, which is why a format that can be auto-generated has been chosen.

Regarding auto-generating the system description XML, yes we will need some kind of description of the rules and expected format for each kind of element. It is something I have started to think about. I have not done something like this before though, so will have to look into it thoroughly first. It's a relatively low-priority for now, but is definitely something I want to do in the future.

Ivan-Velickovic:
Something I forgot to mention at the Summit is that because of the system description to CapDL translation validation, in the future it will not matter whether we trust the XML parser or not, as we have a proof that the CapDL outputted is valid.

gernotheiser:
> Regarding auto-generating the system description XML, yes we will need some kind of description of the rules and expected format for each kind of element. It is something I have started to think about. 

Given that we have an Isabelle import, doesn't this already exist?

Ivan-Velickovic:
> > Regarding auto-generating the system description XML, yes we will need some kind of description of the rules and expected format for each kind of element. It is something I have started to think about.
> 
> Given that we have an Isabelle import, doesn't this already exist?

It may, but it needs to be in an accessible format which I am not sure if it is.

lsf37:
> Something I forgot to mention at the Summit is that because of the system description to CapDL translation validation, in the future it will not matter whether we trust the XML parser or not, as we have a proof that the CapDL outputted is valid.

This is not correct -- the XML is the top-level specification that the capDL output and system image is proved against. If the XML parser consistently produces an unintended interpretation of the input and it is used as frontend for all steps then all tools will generate that unintended system and prove that it is indeed that unintended system. That is, the XML parser does remain trusted.

What the proof will catch are parser crashes or malformed specs, but it can't make sure that the system is the one that you want.

[...]

PR #77, opened 2023-10-24
Added XSD and a validation script

podhrmic:
Partially resolves https://github.com/seL4/microkit/issues/70

**TL;DR;** I added an XSD schema for validating the system XMLs. You are currently using numbers with underscores, such as `period="100_000"` which is *not* a valid XML number. I parse these as strings, which is fine as long as you trust your build tool to resolve these correctly.

**The long story**
I asked ChatGPT to generate XSD schema from the few Microkit examples that are out there, and it did pretty well. I manually edited the XSD in a few places, and it certainly can be improved further, but IMHO it is good enough as it is. I recommend validating all `.system` files in a CI.


Ivan-Velickovic:
Thanks for doing this @podhrmic. I'm not familiar with XSD, do you know if there's more sanitisation we can do, such as a minimum and maximum value for the period/budget for example?

For #70 I was hoping for some kind of specification that would be the input for the Microkit tool as well as any other tools that generate or consume the system description. One thing to consider is that with this XSD file, there is a possibility for the tool to be out-of-sync with the XSD or vice-versa. Something to look into would be whether we can consume the XSD file in Microkit to avoid that, or do you know of anything else in the XML space that would achieve this?

Ivan-Velickovic:
And sorry for taking so long to review the PR.

podhrmic:
No worries, I figured you were busy!

You can certainly set min/max for any integer attribute, similar to what I have done for the priority values [here](https://github.com/GaloisInc/microkit/blob/xsd/system.xsd#L53). It just requires the budget etc. to be treated as integers, not strings (see my initial comment).

You can certainly generate datatypes from XSD, ([xmlschema](https://pypi.org/project/xmlschema/), [xsdata](https://xsdata.readthedocs.io/en/latest/), [xml-schema](https://github.com/media-io/xml-schema),  [xgen](https://github.com/xuri/xgen), ... ). Perhaps a good start is to explicitly write the requirements for the  `.system` format, implement them in XSD and then use the XSD to generate code for Microkit tools.

If this is interesting to you, I can look into it further!

vjackson725:
See my comment on #430, but the nesting of `<sequence>` and `<choice>` does not correctly enforce the per element bounds on the `<irq>` element. A different approach will be needed. Ideally, we could use XSD 1.1, which supports unordered children with numerosity constraints via `<all>`. But `xmllint` does not support 1.1 yet, and there are very few open source validators for XSD 1.1 that I can find.

I suggest this pull request is closed, and discussion on schemas is moved to an issue instead.

PR #79, opened 2023-10-25
Omit assembler debug level to improve toolchain compatability

nspin:
More recent versions of `aarch64-none-elf-as` don't support passing a debug level to `-g`.

Specifically, GNU Binutils 2.40 (associated with GCC 12.3) doesn't support this.

### DIFF
--- a/libmicrokit/Makefile
+++ b/libmicrokit/Makefile
@@ -22,7 +22,7 @@ $(BUILD_DIR)/%.o : src/%.S
 	$(TOOLCHAIN)gcc -x assembler-with-cpp -c -g3  -mcpu=$(GCC_CPU)  $< -o $@
 
 $(BUILD_DIR)/%.o : src/%.s
-	$(TOOLCHAIN)as -g3 -mcpu=$(GCC_CPU) $< -o $@
+	$(TOOLCHAIN)as -g -mcpu=$(GCC_CPU) $< -o $@
 
 $(BUILD_DIR)/%.o : src/%.c
 	$(TOOLCHAIN)gcc -c $(CFLAGS) $< -o $@
--- a/loader/Makefile
+++ b/loader/Makefile
@@ -32,7 +32,7 @@ $(BUILD_DIR)/%.o : src/%.S
 	$(TOOLCHAIN)gcc -x assembler-with-cpp -c -g3  -mcpu=$(GCC_CPU)  $< -o $@
 
 $(BUILD_DIR)/%.o : src/%.s
-	$(TOOLCHAIN)as -g3 -mcpu=$(GCC_CPU) $< -o $@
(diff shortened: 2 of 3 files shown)


Ivan-Velickovic:
I guess the reason we pin a specific version of the toolchain is to avoid issues like this.

I'm not hesitant to upgrade the toolchain version if there's some specific reason for it, but the README does mention a specific toolchain version for a reason.

nspin:
I noticed that the debug level parameter to `-g` in `as` is undocumented even in Binutils 2.35.1, the version included in the toolchain pinned in the Microkit README, which accepts the paramter. I took a quick look at `as.c` in Binutils 2.35.1 and Binutils 2.40, the version I was using that does not accept that parameter, along with the [commit responsible for this change](https://github.com/bminor/binutils-gdb/commit/6ea673e2d643b7b2283aa73d35b02dfc9aa7115f), which confirmed my suspicion that the debug level parameter doesn't actually do anything in the `as` version that does accept it.

So, there is no cost to dropping the debug level parameter to `-g`.

Ivan-Velickovic:
> I noticed that the debug level parameter to `-g` in `as` is undocumented even in Binutils 2.35.1, the version included in the toolchain pinned in the Microkit README, which accepts the paramter. I took a quick look at `as.c` in Binutils 2.35.1 and Binutils 2.40, the version I was using that does not accept that parameter, along with the [commit responsible for this change](https://github.com/bminor/binutils-gdb/commit/6ea673e2d643b7b2283aa73d35b02dfc9aa7115f), which confirmed my suspicion that the debug level parameter doesn't actually do anything in the `as` version that does accept it.
> 
> So, there is no cost to dropping the debug level parameter to `-g`.

Thanks for looking into it, I have come across this issue before and I did find documentation on it here https://gcc.gnu.org/onlinedocs/gcc/Debugging-Options.html (search for `-glevel`).

I know that the toolchain provided by ARM for GCC 12 and above does not allow specifying the level and for adding macOS support to the SDK we would need to upgrade the toolchain version anyways as the macOS ARM64 host toolchain is only available for 12 and above.

This kind of inconsistency between C compilers is quite... annoying. What I might do is take your patch and see if the hash of all the ELFs is the same, that way we'll know if specifying the level is doing anything.

nspin:
GCC supports a debug level parameter, but the issue here is the assembler `$(TOOLCHAIN)-as` not supporting it.

Ivan-Velickovic:
Ah my bad, I thought you were removing `-g3` from GCC as well. Apologies.

Ivan-Velickovic:
No difference in the binaries. Merged in this case because we don't have to do any special handling to get both the ARM provided toolchain and the Nix ARM toolchain working and I want to upgrade to GCC 12 soon anyways for macOS support.

Issue #49, opened 2023-09-13
Create documentation on the internals and design of Microkit


Ivan-Velickovic:
It is important to outline the philosophy and goals of Microkit in a transparent way. This has been partially done in various places, but not in any single central place like this repository.

This documentation should contain the general goals as well as an overview of the design of Microkit and all the components that make up Microkit (e.g loader, initial task, libmicrokit).

PR #80, opened 2023-10-25
monitor: Increase BOOTSTRAP_INVOCATION_DATA_SIZE

nspin:
To enable protection domains with larger memory footprints.

https://github.com/seL4/rust-microkit-http-server-demo, for example, requires at least 122 `seL4_Word`s.

### DIFF
--- a/monitor/src/main.c
+++ b/monitor/src/main.c
@@ -79,7 +79,7 @@
  *
  * FIXME: This can be smaller once compression is enabled.
  */
-#define BOOTSTRAP_INVOCATION_DATA_SIZE 110
+#define BOOTSTRAP_INVOCATION_DATA_SIZE 150
 
 seL4_IPCBuffer *__sel4_ipc_buffer;
 


monitor/src/main.c
```diff
@@ -79,7 +79,7 @@
  *
  * FIXME: This can be smaller once compression is enabled.
  */
-#define BOOTSTRAP_INVOCATION_DATA_SIZE 110
+#define BOOTSTRAP_INVOCATION_DATA_SIZE 256
```
axel-h: I think the comment could also get a few more words then, where the magic value roughly comes from.

Ivan-Velickovic:
Lowered to 150 as internally at TS we've been able to build fairly large systems without increasing it above 150. Not that it matters much as this part of the monitor will be removed soon once the capDL initialiser is used. I just didn't want to increase the number by too much since I'd like to do a comparison of memory usage before and after the capDL changes.

monitor/src/main.c
Ivan-Velickovic: If this part of the monitor was sticking around it would be part of https://github.com/seL4/microkit/issues/49 but it's going to be removed in the transition to capDL so I think we'll leave the comment as it is for now.

PR #81, opened 2023-10-25
tool: Handle unaligned PD segments

nspin:
Correctly handle PD segments whose start vaddrs do not lie on page boundaries.


Ivan-Velickovic:
Surprised this is a simple one line change, but I couldn't find any other instances where it wasn't considering the segment alignment. Thanks!

PR #84, opened 2023-11-26
added concat of headers as a build output

isubasinghe:
# Motivation
These artefacts are needed as part of the verification pipeline, these concatenated files are cleaned and then fed into the C parser.

### DIFF
--- a/libmicrokit/Makefile
+++ b/libmicrokit/Makefile
@@ -17,6 +17,7 @@ CFLAGS := -std=gnu11 -g3 -O3 -nostdlib -ffreestanding -mcpu=$(GCC_CPU) -Wall -Wn
 LIBS := libmicrokit.a
 OBJS := main.o dbg.o
 OTHER_OBJS :=  crt0.o
+VERIFICATION := main_verification.c
 
 $(BUILD_DIR)/%.o : src/%.S
 	$(TOOLCHAIN)gcc -x assembler-with-cpp -c -g3  -mcpu=$(GCC_CPU)  $< -o $@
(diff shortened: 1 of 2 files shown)


libmicrokit/Makefile
```diff
@@ -17,19 +17,23 @@ CFLAGS := -std=gnu11 -g3 -O3 -nostdlib -ffreestanding -mcpu=$(GCC_CPU) -Wall -Wn
 LIBS := libmicrokit.a
 OBJS := main.o dbg.o
 OTHER_OBJS :=  crt0.o
+ALLS := main_all.c
```
Ivan-Velickovic:
```suggestion
# For verification, we concatenate all the C source files.
ALLS := microkit_all.c
```

We should probably follow the same style as the kernel with `*_all` and `*_all_pp` etc.

monitor/Makefile
```diff
@@ -16,6 +16,7 @@ CFLAGS := -std=gnu11 -g3 -O3 -nostdlib -ffreestanding -mcpu=$(GCC_CPU) -Wall -Wn
 
 PROGS := monitor.elf
 OBJECTS := main.o crt0.o debug.o util.o
+ALLS := main_all.c
```
Ivan-Velickovic: Same here with my comment above

libmicrokit/Makefile
isubasinghe: Makes sense

Ivan-Velickovic:
> No opinions from me on that one. Ideally, the concatenated C file that the verification sees should also be what the compiler gets

Okay yes we should definitely do that then. Thanks.

Ivan-Velickovic:
Going to close this for now and re-open/make a new PR once the verification for Microkit is caught up to speed with the latest Microkit and we start working on adding verification checks for libmicrokit to the CI.

Issue #30, opened 2023-05-03
ZCU102 hello world example


podhrmic:
Hello,
I am running intro problems running the ZCU102 example. I am using the [Xilinx QEMU fork](https://xilinx-wiki.atlassian.net/wiki/spaces/A/pages/821428302/Chapter+1+-+Introduction+to+QEMU), and I can boot the [sel4 test](https://docs.sel4.systems/Hardware/ZCU102.html) on the emulated board from Uboot using tftpboot (so I am reasonably confident that the emulator is set up correctly).

I am building the example with `dev_build.py --board=zcu102 --example=hello`

However, when I attempt to run the `loader.img` I get the following error:

```
# Starting application at 0x08000000 ...
LDR|INFO: altloader for seL4 starting
LDR|ERROR: mismatch on loader data structure magic number
...
Resetting CPU ...
```

The docs say that:

> The ZCU102 can run on a physical board or on an appropriate QEMU based emulator.

Can you advise how to run the hello world example on ZCU102?

Ivan-Velickovic:
The address that you load the system image to must match the `loader_link_address` specified, in the case of the ZCU102 it is `0x40000000` (https://github.com/BreakawayConsulting/sel4cp/blob/main/build_sdk.py#L65).

This is also why the QEMU command in the manual has this argument: `-device loader,file=[SYSTEM IMAGE],addr=0x40000000,cpu-num=0` in order to load the system at that address.

PR #87, opened 2024-01-11
Minor manual fixes

Ivan-Velickovic:
Fixes spelling mistakes and adds instructions for booting on ZCU102 hardware. Closes https://github.com/seL4/microkit/issues/30.

In the future we need to find out how to integrate the documentation for each board with the official seL4 documentation. Including how to setup U-Boot etc is too much for the Microkit manual most likely but I still believe that the manual should include instructions specific to Microkit should remain in the manual.

### DIFF
--- a/docs/manual.md
+++ b/docs/manual.md
@@ -576,13 +576,19 @@ When debugging is enabled the kernel will use the same UART as U-Boot.
 
 ## ZCU102
 
-Initial support is available for the ZCU102.
+Initial support is available for the Xilinx ZCU102.
 
 **FIXME:** Additional documentation required here.
 
 The ZCU102 can run on a physical board or on an appropriate QEMU based emulator.
 
-An QEMU command line:
+Microkit produces a raw binary file, so when using U-Boot you must execute the image using:
+
+    => go 0x40000000
+
+Note that the loading address must be `0x40000000`.
+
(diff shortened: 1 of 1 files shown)


docs/manual.md
```diff
@@ -576,13 +576,19 @@ When debugging is enabled the kernel will use the same UART as U-Boot.
...
 **FIXME:** Additional documentation required here.
 
 The ZCU102 can run on a physical board or on an appropriate QEMU based emulator.
 
-An QEMU command line:
+Microkit produces a raw binary file, so when using U-Boot you must execute the image using:
+
+    => go 0x40000000
+
+Note that the loading address must be `0x40000000`.
```
axel-h, line 589: If you add this note, you should also say, why it must be this address? The note's wording feels a bit odd now to me.

Ivan-Velickovic: I can remove the note or say something like 'The image expects to be loaded at `0x40000000`'. I added the note because I wanted to make it clear that it's a fixed address. The reasoning for the fixed address is explained in the section about adding platform support, it will also be explained once https://github.com/seL4/microkit/issues/49 is resolved.

I think I will probably just remove the note to make it consistent with the TQMa8XQP documentation.

Thanks.

PR #94, opened 2024-01-27
Add macOS support

Ivan-Velickovic:
It would be sad to force macOS users to have to dual boot or use a Docker container in order to use the Microkit SDK. These patches add macOS support to the Microkit SDK and have received internal use at TS for a while now so I think it's fair to say that the macOS support will be stable.

Couple of things left to do:
- [x] Update README to mention macOS is supported too
- [x] Use the same toolchain version across macOS/Linux (ARM only supplies macOS ARM64 toolchains for GCC 12.2 and above so we should make it consistent across environments when building Microkit).
- [x] I have noticed the ARM GCC toolchains are *really* slow to download sometimes, causing timeouts in the CI. I've been annoyed by this to the point that I just use my own web server on my fork of Microkit but it's probably not a good idea to do that here. Not sure what to do about this, let's start with using ARM's website for a start and see what happens (ideally we'd have backups of these toolchains etc anyways).

### DIFF
--- a/.github/workflows/sdk.yaml
+++ b/.github/workflows/sdk.yaml
@@ -37,9 +37,9 @@ jobs:
             python3.9 python3-pip python3.9-venv musl-tools
       - name: Install AArch64 GCC toolchain
         run: |
-          wget -O aarch64-toolchain.tar.gz https://developer.arm.com/-/media/Files/downloads/gnu-a/10.2-2020.11/binrel/gcc-arm-10.2-2020.11-x86_64-aarch64-none-elf.tar.xz\?revision\=79f65c42-1a1b-43f2-acb7-a795c8427085\&hash\=61BBFB526E785D234C5D8718D9BA8E61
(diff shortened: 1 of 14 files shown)


Ivan-Velickovic:
Right, well looks like the timeout I mentioned happened on the very first CI run for building the macOS SDK.

lsf37:
If you're referencing a specific file, then a faster mirror would be fine. It's an additional step when you update the dependency, but you could have a script for that. We could put the file on a foundation's AWS S3 bucket, for instance. Cost would be negligible for CI traffic.

You could also try caching the download with http://github.com/actions/cache, but that first needs a successful download.

Ivan-Velickovic:
Getting the toolchains from the AWS bucket seems to be working well. I kept the README links to the ARM website since it probably looks a bit dodgy to newcomers to download some specific toolchain off an AWS bucket. @lsf37 what do you think?

lsf37:
> Getting the toolchains from the AWS bucket seems to be working well. I kept the README links to the ARM website since it probably looks a bit dodgy to newcomers to download some specific toolchain off an AWS bucket. @lsf37 what do you think?

Agreed, that's a good idea. Also reduces traffic to AWS :-)

Issue #13, opened 2022-12-22
The SDK build fails with "crypt.h: No such file or directory"


podhrmic:
Hello!

I am very excited about the new core platform! I tried to build the SDK following the instructions, but I am getting this error:

```
generating custom link library containing Python...
deriving custom config.c from 101 extension modules
compiling custom config.c to object file
running: "musl-gcc" "-O0" "-ffunction-sections" "-fdata-sections" "-fPIC" "-m64" "-I" "/tmp/libpython58Y2ML" "-Wall" "-Wextra" "-std=c99" "-DNDEBUG" "-DPy_BUILD_CORE" "-o" "/tmp/pyoxidizer-build-exe-packagingQ0Ofjv/config.o" "-c" "/tmp/libpython58Y2ML/config.c"
cargo:warning=In file included from /tmp/libpython58Y2ML/config.c:1:
cargo:warning=/tmp/libpython58Y2ML/Python.h:44:10: fatal error: crypt.h: No such file or directory
cargo:warning=   44 | #include <crypt.h>
cargo:warning=      |          ^~~~~~~~~
cargo:warning=compilation terminated.
exit status: 1


error occurred: Command "musl-gcc" "-O0" "-ffunction-sections" "-fdata-sections" "-fPIC" "-m64" "-I" "/tmp/libpython58Y2ML" "-Wall" "-Wextra" "-std=c99" "-DNDEBUG" "-DPy_BUILD_CORE" "-o" "/tmp/pyoxidizer-build-exe-packagingQ0Ofjv/config.o" "-c" "/tmp/libpython58Y2ML/config.c" with args "musl-gcc" did not execute successfully (status code exit status: 1).
```

Not quite sure where is the problem coming from, `libcrypt-dev` is installed, and a simple call to `import crypt` succeeds:
```
# ./pyenv/bin/python                           
Python 3.10.6 (main, Nov 14 2022, 16:10:14) [GCC 11.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import crypt
>>> 
```

I am building this on `ubuntu:latest`:
```
# cat /etc/lsb-release 
DISTRIB_ID=Ubuntu
DISTRIB_RELEASE=22.04
DISTRIB_CODENAME=jammy
DISTRIB_DESCRIPTION="Ubuntu 22.04.1 LTS
```

Ivan-Velickovic:
[...]

PR #102, opened 2024-02-14
Add dependencies to README.md

wom-bat:
Make the list of dependencies for building the microkit tool more complete.

This list was created by making a minimal Debian stable chroot environment, and adding packages until the build worked

It addresses Issue #6

### DIFF
--- a/docs/manual.md
+++ b/docs/manual.md
@@ -479,6 +479,7 @@ The `irq` element has the following attributes:
 
 * `irq`: The hardware interrupt number.
 * `id`: The channel identifier. Must be at least 0 and less than 63.
+* `trigger`: (optional) Whether the IRQ is edge triggered ("edge") or level triggered ("level"). Defaults to "level".
 
 The `setvar` element has the following attributes:
 
--- a/tool/microkit/__main__.py
+++ b/tool/microkit/__main__.py
@@ -64,7 +64,7 @@ from microkit.sel4 import (
     Sel4CnodeMint,
     Sel4CnodeCopy,
     Sel4UntypedRetype,
-    Sel4IrqControlGet,
+    Sel4IrqControlGetTrigger,
     Sel4IrqHandlerSetNotification,
     Sel4SchedControlConfigureFlags,
     emulate_kernel_boot,
@@ -1173,9 +1173,10 @@ def build_system(
         for sysirq in pd.irqs:
             cap_address = system_cap_address_mask | cap_slot
             system_invocations.append(
-                Sel4IrqControlGet(
+                Sel4IrqControlGetTrigger(
                     IRQ_CONTROL_CAP_ADDRESS,
                     sysirq.irq,
+                    sysirq.trigger.value,
                     root_cnode_cap,
                     cap_address,
                     kernel_config.cap_address_bits
--- a/tool/microkit/sel4.py
+++ b/tool/microkit/sel4.py
@@ -158,6 +158,11 @@ def calculate_rootserver_size(initial_task_region: MemoryRegion) -> int:
     return size
 
 
+class Sel4ArmIrqTrigger(IntEnum):
+    Level = 0
+    Edge = 1
+
+
 class Sel4Aarch64Regs:
     """
 typedef struct seL4_UserContext_ {
@@ -528,13 +533,14 @@ class Sel4AsidPoolAssign(Sel4Invocation):
 
 
 @dataclass
-class Sel4IrqControlGet(Sel4Invocation):
+class Sel4IrqControlGetTrigger(Sel4Invocation):
     _object_type = "IRQ Control"
     _method_name = "Get"
     _extra_caps = ("dest_root", )
-    label = Sel4Label.IRQIssueIRQHandler
+    label = Sel4Label.ARMIRQIssueIRQHandlerTrigger
     irq_control: int
     irq: int
+    trigger: int
     dest_root: int
     dest_index: int
     dest_depth: int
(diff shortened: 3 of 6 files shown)


README.md
```diff
@@ -75,6 +97,11 @@ https://developer.arm.com/-/media/Files/downloads/gnu/12.2.rel1/binrel/arm-gnu-t
 On macOS Intel/x86-64 the following version is used:
 https://developer.arm.com/-/media/Files/downloads/gnu/12.2.rel1/binrel/arm-gnu-toolchain-12.2.rel1-darwin-x86_64-aarch64-none-elf.tar.xz?rev=09b11f159fc24fdda01e05bb32695dd5&hash=6AAF4239F28AE17389AB3E611DFFE0A6
 
+Recent Linux systems may not be compatible with the desired version of
```
Ivan-Velickovic: This is no longer true, I upgraded the version of pyoxidizer recently.

README.md
```diff
@@ -39,7 +39,29 @@ Please file an issue if additional packages are required.
 * python3.9
 * python3.9-venv
 * musl-1.2.2
-* ARM GCC compiler; 10.2-2020.11
+* cmake
+* ninja-build
+* ARM GCC compiler for none-elf; 10.2-2020.11
+* aarch64 Linux gcc compiler
```
Ivan-Velickovic: This is not necessary anymore either. `aarch64-none-elf-` is used for everything.

README.md
```diff
@@ -39,7 +39,29 @@ Please file an issue if additional packages are required.
 * python3.9
 * python3.9-venv
 * musl-1.2.2
-* ARM GCC compiler; 10.2-2020.11
+* cmake
+* ninja-build
+* ARM GCC compiler for none-elf; 10.2-2020.11
```
Ivan-Velickovic: This should be version 12.2.1 20221205, my bad I forgot to update it in https://github.com/seL4/microkit/pull/94

README.md
wom-bat: It breaks without it.

Ivan-Velickovic:
In general, I was hesitant to add this myself as with this approach we are recommending people use their system's package manager and so we are at the mercy of whatever they do like in https://github.com/seL4/microkit/issues/13.

@bennoleslie originally wrote "Note: It is a high priority of the authors to ensure builds are self-contained and repeatable. A high value is placed on using specifically versioned tools. At this point in time this is not fully realised, however it is a high priority to enable this in the near future."

I don't know exactly how he meant to realise this though...

README.md
wom-bat: Cool I can fix that.

README.md
Ivan-Velickovic: Can you show me the error message? The dependencies installed for the CI are here https://github.com/seL4/microkit/blob/89d1e621da862a0cbe33e0ac333eae61b579286c/.github/workflows/sdk.yaml#L33 and no longer use `aarch64-linux-gnu-`. The use of it was removed in https://github.com/seL4/microkit/commit/cb849f7695a39d3e0a9e2b34f7c9e360992a8883.

wom-bat: This may have been because I was testing on a different branch.  Repeating on master now

wom-bat:
One other note: python3.9 is no longer in the repos.  I'm using 3.11 without issues...

wom-bat:
Also, musl 1.2.3 is the one currently available (and 1.2.4 in Debian sid).  The changes from 1.2.2 are very minor, and will not be visible in our usecases

wom-bat:
I think wrt the reproducibility, the only issue is libmuslc  I think nothing  else that's variable will end up visible in the Monitor or Loader.

README.md
```diff
@@ -39,7 +39,28 @@ Please file an issue if additional packages are required.
...
+* texlive-latex-extra
+
+On a Debian-like system you can do:
+
+    $ sudo apt install build-essential git cmake ninja-build \
+	    device-tree-compiler libxml2-utils \
+		pandoc texlive-latex-base texlive-latex-recommended \
+		texlive-fonts-recommended texlive-fonts-extra \
+		python3 python3-venv \
+		musl-dev musl-tools
```
Ivan-Velickovic: There's been multiple issues when using the apt packaged musl, I'm wondering whether we should recommend building from source here. We can keep the apt install for now maybe and see how many people run into issues.

`musl-dev` is not needed.

README.md
```diff
@@ -39,7 +39,28 @@ Please file an issue if additional packages are required.
...
+* texlive-latex-recommended
+* texlive-fonts-recommended
+* texlive-fonts-extra
+* texlive-latex-extra
+
+On a Debian-like system you can do:
+
+    $ sudo apt install build-essential git cmake ninja-build \
+	    device-tree-compiler libxml2-utils \
+		pandoc texlive-latex-base texlive-latex-recommended \
```
Ivan-Velickovic: `texlive-latex-recommended` is not needed

README.md
Ivan-Velickovic: Thanks

README.md
Ivan-Velickovic: actually I might be wrong here - the CI doesn't install it but it might have it already available. I'll check on a fresh install.

README.md
wom-bat: musl-dev includes the actual library; musl-tools has musl-gcc that uses it.  musl-dev will be pulled in when musl-dev is installed; this make the dependency explicit.

README.md
wom-bat: I installed it to get xcolor.sty amongst other things

README.md
Ivan-Velickovic: Yes you are correct, my bad. I'm fine with having it explicit.

Ivan-Velickovic:
Argh, was trying to push to the branch and made a mistake... Apologies

PR #113, opened 2024-03-19
Add bzero

wom-bat:
Use the compiler builtin if we can.  This will work for GCC and CLANG; we usually build with GCC


Ivan-Velickovic:
Wrong repo I think.

PR #126, opened 2024-05-21
Rewrite the tool

Ivan-Velickovic:
This PR rewrites the entire Microkit tool from Python to Rust.

This was done for the following reasons:
* The current tool was difficult to build from source. It took too long,
  did not have any cache for rebuilds, and sometimes would not reliably
  build/link as a static binary.
* While perhaps not as extensive as C, Rust does have some verification
  tooling surrounding it, meaning it is possible for us to verify the
  Microkit tool in the future. Although at this time there are no
  concrete plans to verify the tool.

This rewrite, while not significantly reducing our dependencies in
general, does improve the situation. We are no longer dependent on
pyoxidizer, which was causing most of the issues with the tool.

We introduce a dependency on the Rust toolchain. The tool itself has
two dependencies (for XML parsing). No other dependencies are needed,
and it was a conscious decision to re-invent the wheel where the Rust
standard library was lacking (compared to the Python standard library).

The other programming languages considered for the rewrite were Zig and
Go. While they both have their benefits and costs (as does Rust), they were not
chosen due to them not being in any other seL4 Foundation supported projects
meaning that it could have been difficult to continue maintaining them
in the future.

The main cost of rewriting the tool in Rust is that it has
become slightly more complicated and harder to understand. This could potentially
result in a higher barrier to entry for contributors. However, most
contributions such as adding platform support will not need to touch the
tool.


Ivan-Velickovic:
It should be noted that for the example programs that are in the repository, the rewritten tool produces the same binary as the current tool.

The only observable differences between the new tool and the current one is various bug fixes and improvements to error messages.

Other than that, there should be *no* observable differences to users.

PR #129, opened 2024-05-29
Fix some bugs in the new tool

nspin:
- Fix handling of PD names longer than 16 bytes
- Fix some cases of incorrect constants being used

### DIFF
--- a/tool/microkit/src/main.rs
+++ b/tool/microkit/src/main.rs
@@ -1801,7 +1801,9 @@ fn build_system(kernel_config: &Config,
 
     for (i, pd) in system.protection_domains.iter().enumerate() {
         let elf = &mut pd_elf_files[i];
-        elf.write_symbol("microkit_name", pd.name.as_bytes())?;
+        let name = pd.name.as_bytes();
+        let name_length = min(name.len(), PD_MAX_NAME_LENGTH);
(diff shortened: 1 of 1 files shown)


tool/microkit/src/main.rs
```diff
@@ -1801,7 +1801,8 @@ fn build_system(kernel_config: &Config,
 
     for (i, pd) in system.protection_domains.iter().enumerate() {
         let elf = &mut pd_elf_files[i];
-        elf.write_symbol("microkit_name", pd.name.as_bytes())?;
+        let pd_name_bytes = pd.name.as_bytes();
+        elf.write_symbol("microkit_name", &pd_name_bytes[..pd_name_bytes.len().min(PD_MAX_NAME_LENGTH)])?;
```
Ivan-Velickovic: Can you do `let name_length = min(pd_name_bytes.len(), PD_MAX_NAME_LENGTH);` like below to make it a bit more readable.

Ivan-Velickovic:
Sorry about that. I'm away for the next couple of weeks but when I get back I'll work on runtime tests which would have caught at least some of these mistakes.

Will merge once the comment I made is addressed.

tool/microkit/src/main.rs
nspin: Done

PR #131, opened 2024-06-07
build_sdk.py: add support for minimal build

nspin:
This PR adds four new options to the `build_sdk.py` CLI:

- `--only-board <board>`: only artifacts for `<board>` will be built
- `--only-config <config>`: only artifacts for `<config>` (either `"debug"` or `"release"`) will be built
- `--skip-docs`: docs will not be built
- `--skip-source-tarball`: source tarball will not be built

These options are useful in cases where the SDK is being built by a build system that doesn't have any use for artifacts other than those necessary for building a particular system.

### DIFF
--- a/build_sdk.py
+++ b/build_sdk.py
@@ -411,7 +411,33 @@ def main() -> None:
     parser = ArgumentParser()
     parser.add_argument("--sel4", type=Path, required=True)
     parser.add_argument("--tool-target-triple", default=get_tool_target_triple(), help="Compile the Microkit tool for this target triple")
+    parser.add_argument("--boards", metavar="BOARDS", help="Comma-separated list of boards to support. When absent, all boards are supported.")
+    parser.add_argument("--configs", metavar="CONFIGS", help="Comma-separated list of configurations to support. When absent, all configurations are supported.")
+    parser.add_argument("--skip-docs", action="store_true", help="Docs will not be built")
+    parser.add_argument("--skip-tar", action="store_true", help="SDK and source tarballs will not be built")
+
     args = parser.parse_args()
+
+    if args.boards is not None:
+        supported_board_names = frozenset(board.name for board in SUPPORTED_BOARDS)
+        selected_board_names = frozenset(args.boards.split(","))
+        for board_name in selected_board_names:
+            if board_name not in supported_board_names:
+                raise Exception(f"Trying to build a board: {board} that does not exist in supported list.")
+        selected_boards = [board for board in SUPPORTED_BOARDS if board.name in selected_board_names]
+    else:
+        selected_boards = SUPPORTED_BOARDS
+
+    if args.configs is not None:
+        supported_config_names = frozenset(config.name for config in SUPPORTED_CONFIGS)
+        selected_config_names = frozenset(args.configs.split(","))
+        for config_name in selected_config_names:
+            if config_name not in supported_config_names:
+                raise Exception(f"Trying to build a configuration: {config} that does not exist in supported list.")
+        selected_configs = [config for config in SUPPORTED_CONFIGS if config.name in selected_config_names]
+    else:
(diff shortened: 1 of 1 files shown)


Ivan-Velickovic:
> These options are useful in cases where the SDK is being built by a build system that doesn't have any use for artifacts other than those necessary for building a particular system.

Would this be solved once the SDK is released?

nspin:
These options would still be useful for cases where projects build Microkit from source. One example that comes to mind is working off of HEAD rather than the latest release. Another example is hacking on Microkit itself.

Ivan-Velickovic:
Yep I agree I just meant if your specific use-case would still need these options.

Ivan-Velickovic:
I assume the use-case is for the Rust Microkit demos which I think should just be able to use a released version available for download but just want to check.

nspin:
Yes, those demos will use the binary release.

Ivan-Velickovic:
Can you change —only-board to something like —boards and make it take a comma separated list of boards that are built, same thing for the config?

This has been pretty useful for me in the past and is something I have in my fork.

nspin:
I changed it to take multiple `--board` and `--config` arguments. How is that?

nspin:
The PR as is allows one to specify `--board` and `--config` multiple times to select multiple boards or configs. For example:

```
microkit --board imx8mm_evk --board qemu_virt_aarch64 --config release --config debug
```

Would you prefer `--boards BOARD1,BOARD2` or `--board BOARD1 --board BOARD2`?

I'll add the existence check.

build_sdk.py
```diff
@@ -483,18 +501,19 @@ def main() -> None:
...
-
-    # Build the source tar
-    process = popen("git ls-files")
-    filenames = [Path(fn.strip()) for fn in process.readlines()]
-    process.close()
-    source_prefix = Path(f"{NAME}-source-{VERSION}")
-    with tar_open(source_tar_file, "w:gz") as tar:
-        for filename in filenames:
-            tar.add(filename, arcname=source_prefix / filename, filter=tar_filter)
+    if not args.skip_source_tarball:
```
Ivan-Velickovic: This skips both the SDK tar and the source tar. Can you call this option just `--skip-tar` instead?

Ivan-Velickovic:
> Would you prefer --boards BOARD1,BOARD2 or --board BOARD1 --board BOARD2?

Sorry that's what I was trying to say in my previous comment. Doing `--boards BOARD1,BOARD2` saves me some typing when targeting more than one board, that's all.

build_sdk.py
nspin: Done

nspin:
Okay, I've made both of those changes.

build_sdk.py
```diff
@@ -411,7 +411,33 @@ def main() -> None:
     parser = ArgumentParser()
     parser.add_argument("--sel4", type=Path, required=True)
     parser.add_argument("--tool-target-triple", default=get_tool_target_triple(), help="Compile the Microkit tool for this target triple")
+    parser.add_argument("--boards", metavar="BOARDS", help="Comma-separated list of boards to support. When absent, all boards are supported.")
+    parser.add_argument("--configs", metavar="CONFIGS", help="Comma-separated list of configurations to support. When absent, all configurations are supported.")
+    parser.add_argument("--skip-docs", action="store_true", help="Docs will not be built")
+    parser.add_argument("--skip-tar", action="store_true", help="Source tarball will not be built")
```
Ivan-Velickovic:
```suggestion
    parser.add_argument("--skip-tar", action="store_true", help="SDK and source tarballs will not be built")
```

Ivan-Velickovic: Sorry, one more minor thing and then it's good to merge.

Ivan-Velickovic:
Cool, thanks!

PR #132, opened 2024-06-07
tool: fix off-by-one error

nspin:
Also fixes the relevant conditional expression.


Ivan-Velickovic:
Thanks, could you leave it as the for loop though?

nspin:
https://github.com/seL4/microkit/pull/133 fixed some cases, but not all. Consider:

```
let mut x = DisjointMemoryRegion::new();
x.insert_region(1, 2);
x.insert_region(0, 1);
```

Ivan-Velickovic:
Cool, thanks again for the fix.

PR #137, opened 2024-06-14
chore: apply clippy lints on Rust code

wucke13:
There is a bunch of code flagged by clippy.

- abb7902fca34ffbfc470e204047bbd6ad56d57a4 is a manual fix for a specific problem flagged by clippy
- 731c204d056729dd11f68f8a85327e6553e98037 adds an allow for something `cargo clippy --fix` would remove automatically
- 2115578eaf331ed124460dfa030f7d6f3c3ac62f automatically fixes most of the issues, it contains precisely the result after `cargo clippy --fix`
- fe8ee09cc082f1b8b06794927f09f024173bbb96 fixes the remaining lints by hand. Some of the issues flagged by clippy are non-issues IMHO, hence I added `#[allow(...)]` statements
- 4e464071017623d23be1a044a326f0783508cbb1 adds  a CI check to point out lints violations in an automated fashion

### DIFF
--- a/tool/microkit/src/lib.rs
+++ b/tool/microkit/src/lib.rs
@@ -123,21 +123,12 @@ impl MemoryRegion {
     }
 }
 
+#[derive(Default)]
 pub struct DisjointMemoryRegion {
     pub regions: Vec<MemoryRegion>,
 }
 
-impl Default for DisjointMemoryRegion {
-    fn default() -> Self {
-        Self::new()
-    }
-}
-
 impl DisjointMemoryRegion {
-    pub fn new() -> DisjointMemoryRegion {
-        DisjointMemoryRegion { regions: Vec::new() }
-    }
-
     fn check(&self) {
         // Ensure that regions are sorted and non-overlapping
         let mut last_end: Option<u64> = None;
--- a/tool/microkit/src/lib.rs
+++ b/tool/microkit/src/lib.rs
@@ -127,6 +127,12 @@ pub struct DisjointMemoryRegion {
     pub regions: Vec<MemoryRegion>,
 }
 
+impl Default for DisjointMemoryRegion {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
 impl DisjointMemoryRegion {
     pub fn new() -> DisjointMemoryRegion {
         DisjointMemoryRegion { regions: Vec::new() }
@@ -218,7 +224,7 @@ impl DisjointMemoryRegion {
         match region_to_remove {
             Some(region) => {
                 self.remove_region(region.base, region.base + size);
-                return region.base;
+                region.base
             },
             None => panic!("Unable to allocate {} bytes", size)
         }
@@ -287,7 +293,7 @@ impl ObjectAllocator {
             untyped.push(UntypedAllocator::new(*ut, 0, vec![]));
         }
 
-        ObjectAllocator { allocation_idx: 0, untyped: untyped }
+        ObjectAllocator { allocation_idx: 0, untyped }
     }
 
     pub fn alloc(&mut self, size: u64) -> KernelAllocation {
--- a/tool/microkit/src/loader.rs
+++ b/tool/microkit/src/loader.rs
@@ -274,6 +274,7 @@ impl<'a> Loader<'a> {
 
         let mut boot_lvl1_lower: [u8; PAGE_TABLE_SIZE] = [0; PAGE_TABLE_SIZE];
         for i in 0..512 {
+            #[allow(clippy::identity_op)] // keep the (0 << 2) for clarity
             let pt_entry: u64 =
                 ((i as u64) << AARCH64_1GB_BLOCK_BITS) |
                 (1 << 10) | // access flag
--- a/tool/microkit/src/loader.rs
+++ b/tool/microkit/src/loader.rs
@@ -288,14 +288,14 @@ impl<'a> Loader<'a> {
         let mut boot_lvl0_upper: [u8; PAGE_TABLE_SIZE] = [0; PAGE_TABLE_SIZE];
         {
             let pt_entry = (boot_lvl1_upper_addr | 3).to_le_bytes();
-            let idx = Aarch64::lvl0_index(first_vaddr) as usize;
+            let idx = Aarch64::lvl0_index(first_vaddr);
             boot_lvl0_upper[8 * idx..8 * (idx + 1)].copy_from_slice(&pt_entry);
         }
 
         let mut boot_lvl1_upper: [u8; PAGE_TABLE_SIZE] = [0; PAGE_TABLE_SIZE];
         {
             let pt_entry = (boot_lvl2_upper_addr | 3).to_le_bytes();
-            let idx = Aarch64::lvl1_index(first_vaddr) as usize;
+            let idx = Aarch64::lvl1_index(first_vaddr);
             boot_lvl1_upper[8 * idx..8 * (idx + 1)].copy_from_slice(&pt_entry);
         }
 
--- a/tool/microkit/src/loader.rs
+++ b/tool/microkit/src/loader.rs
@@ -243,20 +243,20 @@ impl<'a> Loader<'a> {
         let mut loader_buf = BufWriter::new(loader_file);
 
         // First write out all the image data
-        loader_buf.write(self.image.as_slice()).expect("Failed to write image data to loader");
+        loader_buf.write_all(self.image.as_slice()).expect("Failed to write image data to loader");
 
         // Then we write out the loader metadata (known as the 'header')
         let header_bytes = unsafe { struct_to_bytes(&self.header) };
-        loader_buf.write(header_bytes).expect("Failed to write header data to loader");
+        loader_buf.write_all(header_bytes).expect("Failed to write header data to loader");
         // For each region, we need to write out the region metadata as well
         for region in &self.region_metadata {
             let region_metadata_bytes = unsafe { struct_to_bytes(region) };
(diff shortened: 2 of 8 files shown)


wucke13:
Worth pointing out: https://github.com/wucke13/microkit/blob/dev/improve-rust/tool/microkit/src/util.rs#L129-L143 is still unsafe (and likely has actual soundness holes for little vs big endian machines, as irrelevant as these computers may seem). There is multiple ways to address this; one would be https://lib.rs/crates/deku . I would be happy to contribute the code to get rid of the remaining two `unsafe` functions, but I don't want to make the effort if upstream is not interested in this change. Please let me know what you think, would you be OK to get deku (or any other thing you suggest) in to get rid of the `unsafe` bits?

nspin:
The [zerocopy](https://docs.rs/zerocopy/latest/zerocopy/) crate would be another good way to solve this problem. It provides [zerocopy::byteorder](https://docs.rs/zerocopy/latest/zerocopy/byteorder/index.html) for the endianness part.

Ivan-Velickovic:
> (and likely has actual soundness holes for little vs big endian machines, as irrelevant as these computers may seem)

Yes, that's why there is this check https://github.com/seL4/microkit/blob/eef6faff0783ba4c9fa4af94e0c68216836567ef/tool/microkit/src/elf.rs#L181

> multiple ways to address this; one would be https://lib.rs/crates/deku

Based on that page, we would be replacing 15 lines of code with 57,000 lines of code. We would also be adding at least 4 new dependencies. I looked at the source code for deku, and it is also using unsafe. I don't think there's a way to get this behaviour without using unsafe in Rust.

It seems like we're not removing any unsafe, just making it look like we don't have any unsafe by moving it to a dependency.

Ivan-Velickovic:
There are probably more places to assert that we're working with little-endian machines, but I would much prefer improving the existing code than adding a massive dependency.

wucke13:
> I don't think there's a way to get this behaviour without using unsafe in Rust (I'm not that familiar with Rust so I could be wrong here).

There is, for sure. The question is what runtime cost we might accept for it (IMHO neglectable) and how one wants it done. @nspin suggested the use of zerocopy, which might provide fewer degrees of freedom in the binary format but might be smaller in code size. Do I correctly read your answer as an "no unsafe solution is desired, but not if it involves big dependencies and not if the dependencies use unsafe"?

In any case I still think having a dependency with vetted unsafe is most often preferable than home grown unsafe blocks.

Ivan-Velickovic:
> In any case I still think having a dependency with vetted unsafe is most often preferable than home grown unsafe blocks.

In my opinion it depends, in general I trust dependencies less than our own code. We control everything in this repository, we do not control our dependencies.

> Do I correctly read your answer as an "no unsafe solution is desired, but not if it involves big dependencies and not if the dependencies use unsafe"?

Sure. Not all unsafe is the same so I guess it depends.

I will admit that I didn't look to closely at these unsafe bits in the tool so the first thing would be to understand under which cases this leads to unsafe behaviour and whether those cases are acceptable.

.github/workflows/rust.yaml
Ivan-Velickovic, line 1: Can you put this in `.github/workflows/pr.yaml`? Thanks.

tool/microkit/src/lib.rs
```diff
@@ -123,15 +123,12 @@ impl MemoryRegion {
     }
 }
 
+#[derive(Default)]
```
Ivan-Velickovic, line 126: Default doesn't really make sense for DisjointMemoryRegion. Can you keep this as `new`?

tool/microkit/src/loader.rs
```diff
@@ -274,6 +274,7 @@ impl<'a> Loader<'a> {
 
         let mut boot_lvl1_lower: [u8; PAGE_TABLE_SIZE] = [0; PAGE_TABLE_SIZE];
         for i in 0..512 {
+            #[allow(clippy::identity_op)] // keep the (0 << 2) for clariry
```
Ivan-Velickovic:
```suggestion
            #[allow(clippy::identity_op)] // keep the (0 << 2) for clarity
```

tool/microkit/src/main.rs
```diff
@@ -4,6 +4,9 @@
 // SPDX-License-Identifier: BSD-2-Clause
 //
 
+// we want our asserts, even if the compiler figures out they hold true already during compile-time
+#![allow(clippy::assertions_on_constants)]
```
Ivan-Velickovic, line 8: I would be happy to move the asserts on constants outside of runtime functions but I could not find a way to do compile time asserts in Rust. Do you know how? Everything I see online tells me to download a crate....

tool/microkit/src/sel4.rs
```diff
@@ -410,6 +369,7 @@ impl Aarch64Regs {
         ]
     }
 
+    #[allow(clippy::len_without_is_empty)] // this will never be *empty*, bad lint
     /// Just returns the count of registers
     pub const fn len(&self) -> u64 {
```
Ivan-Velickovic: This should probably just be `count` instead of `len`. Do you mind changing it? Then we can get rid of this clippy ignore.

.github/workflows/rust.yaml
```diff
@@ -0,0 +1,20 @@
...
+env:
+  RUSTFLAGS: "-Dwarnings"
+
+jobs:
+  clippy_check:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: Run Clippy
+        run: cargo clippy --all-targets --all-features
```
Ivan-Velickovic:
```suggestion
        run: cd tool/microkit && cargo clippy --all-targets --all-features
```

tool/microkit/src/lib.rs
wucke13: I can do so. I do not have a credible source, but I got the impression that having `::new()` that does nothing but initialize the struct with default values for its members is an anti-pattern. I would only use `new()` if it takes arguments and does some sort of non-default initialization. Or to set non-public members to a non-default value. If it does not take arguments but does non-default initialization, and has some side-effects, `::new()` is also fine. But if `::new()` really does not do anything, no side-effects, no arguments, no non-default initialization, where is the benefit in not saving a couple of lines and just using `[derive(Default)]`? That has also the benefit that it unlocks some syntax constructs like the `MyStruct { some_member: 4, some_other_attr: 1, .. Default::default()}` to initialize most members with default values but a few.

TL;DR, I don't agree with your statement, and from a technical perspective think its less idiomatic. I can not provide strong evidence for my point. I do not want to argue for long, so if you insist I will comply without further arguments :smile: .

tool/microkit/src/loader.rs
wucke13: Good catch!

tool/microkit/src/main.rs
wucke13: According to https://github.com/rust-lang/rust-clippy/issues/8159 there is `const _: () = assert!(usize::BITS >= 32);` to denote that the assert is const. Grasping from the comments it did however not work for complext expression, i. e. 

```rust
const _: () = {
    assert!(X1P_20 <= 1.0e-6); // warns :disappointed:
};
```
We can introduce that, but I'm not too fond of the additional syntax. Adding a const item that is a zero sized type only to assign the non-existant, zero-sized return value of an assert seems weird to me.

tool/microkit/src/lib.rs
Ivan-Velickovic: Ok sure. I really don't care that much so let's just leave it with `Default`.

tool/microkit/src/main.rs
Ivan-Velickovic: Hmm okay let's just keep the clippy ignore then

tool/microkit/src/sel4.rs
wucke13: Will do. General note for future code: It should be possible to make a `Register` trait with an associated constant `REGISTER_COUNT`, where each implementer of that trait (i. e. `Aarch64Regs`) has to provide a value for the `REGISTER_COUNT` constant.

wucke13: ~~And I'll add a test case to verify the constant is correct.~~ I'll simply put the const into the return type signature, that way this is a compile-time enforced property that `LEN` is correct.

.github/workflows/rust.yaml
wucke13: Done :smile:

tool/microkit/src/sel4.rs
Ivan-Velickovic: This is a runtime function because it is sub-optimal to return the full count of the registers. This function is used when setting up the TCB write/read register kernel invocations so you know how many registers to read/write. In some cases you are only writing one or two registers so you do not want this function to return the number of registers but that was done now for simplicity.

I probably should have left a note/TODO, but that's why it's not a const.

wucke13:
@Ivan-Velickovic I think I addressed most of your nitpicks. Shall we keep this history, or do you want me to squash the bits that belong together?

.github/workflows/rust.yaml
Ivan-Velickovic: Ah sorry, `pr.yml` not `pr.yaml`. It's just because that file already exists which is why I thought it'd be better to put this step in there than create a new file.

Ivan-Velickovic:
I've just pushed a couple changes to get CI passing.

wucke13:
I might have force pushed over your changes inadvertently

Ivan-Velickovic:
> I might have force pushed over your changes inadvertently

No worries I'll add back the changes and then merge.

Thanks for making the fixes.

PR #144, opened 2024-06-24
libmicrokit: fix declaration for `fault()`

nspin:
Change `microkit_channel ch` parameter to `microkit_pd pd`.


Ivan-Velickovic:
Fixed this while I am upstreaming virtualisation support but might as well fix it now.

PR #145, opened 2024-06-24
tool: fix bug in index calculation

nspin:
`pd_endpoint_objs` had one entry per PD with an endpoint, but was being indexed as if it had one entry per PD.


Ivan-Velickovic:
Thanks!

PR #147, opened 2024-06-26
docs: type set manual in seL4 foundation style

lsf37:
The text should come out fairly close to the one before, but chapter markings are a bit different, and of course font etc.

There might be some line breaks that don't work out nicely any more because it is a different font. We should look through that and tweak where necessary.


Ivan-Velickovic:
The trailing whitespace check is failing because of the PDF so going to just ignore that.

Ivan-Velickovic:
Hmm....

```
 Error producing PDF.
! Undefined control sequence.
<argument> \hspace *{-25mm}\ifb 
                                @nner\makebox [0pt][l]{\rule {\paperwidth }{...
l.91   }

Traceback (most recent call last):
  File "/home/runner/work/microkit/microkit/build_sdk.py", line 532, in <module>
    main()
  File "/home/runner/work/microkit/microkit/build_sdk.py", line 490, in main
    build_doc(root_dir)
  File "/home/runner/work/microkit/microkit/build_sdk.py", line 360, in build_doc
    assert r == 0
AssertionError
```

Ivan-Velickovic:
I can’t reproduce the CI error and it’s not a very helpful error message (I am also not that familiar with LaTeX).

lsf37:
Weird, I can't reproduce that one either. I'm still seeing a font failure in the CI, though (roboto). Might try to fix that at least.

lsf37:
Ok, now I see the same error (`texlive-fonts-extra` is the right package for roboto)

lsf37:
Thanks for the ubuntu-22 idea. Things are passing now and can be merged from my side.

Not sure if it's worth investigating what went wrong on ubuntu-20.

Ivan-Velickovic (approved):
Given that 22.04 has been out for a while it's probably better that the CI is using that anyways.

If people have issues with building the SDK for 20.04 they can open an issue and we can properly investigate.

PR #149, opened 2024-06-26
tool: fixes for ELF parsing errors

Ivan-Velickovic:
The current code would build up a map of ELF symbols when
first parsing the ELF for quick lookup later on when we
patch/read certain symbols. In this process if we encounter
a symbol that is already in the map, we error and exit.

This patch changes the error to only happen when a symbol we
care about (because we are either writing to it or reading
from it) occurs multiple times.

This lets us parse ELFs even when they have multiple symbols
with the same name and only error if it actually matters.

This issue was brought up by @nspin.

### DIFF
--- a/tool/microkit/src/elf.rs
+++ b/tool/microkit/src/elf.rs
@@ -275,11 +275,11 @@ impl ElfFile {
         Ok(ElfFile { word_size, entry, segments, symbols })
     }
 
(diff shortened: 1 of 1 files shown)


nspin (approved):

nspin:
I tested it, and it works for my case.

PR #150, opened 2024-06-26
build_sdk.py: add support for skipping tool

nspin:
An extension of https://github.com/seL4/microkit/pull/131

Adds the `--skip-tool` flag.

Cargo takes some special care to work in hermetic build environments such as Nix. One good way to deal with all such cases at once is to provide the option to skip building the tool entirely and allow those more particular build environments to take care of building the tool themselves.


Ivan-Velickovic:
Thanks for explaining the motivation.

PR #152, opened 2024-07-01
Add CHANGES.md and VERSION.md

Ivan-Velickovic:


lsf37 (approved):
Looks good. Post release, if we want to use the scripts for the next release, we should add the `---` line they expect as a marker. But not yet.

PR #159, opened 2024-07-03
tool: add rust-toolchain.toml

nspin:
The benefit of a `rust-toolchain.toml` would be enabling anyone building the Microkit SDK to know that they are building it with the same toolchain that upstream used for development and testing.

### DIFF
--- a/tool/microkit/rust-toolchain.toml
+++ b/tool/microkit/rust-toolchain.toml
@@ -1,0 +1,10 @@
+#
+# Copyright 2024, UNSW
+#
+# SPDX-License-Identifier: BSD-2-Clause
+#
+
+[toolchain]
+channel = "1.79.0"
+components = [ "rustfmt" ]
+targets = [ "x86_64-unknown-linux-musl", "x86_64-apple-darwin", "aarch64-apple-darwin" ]


Ivan-Velickovic:
If you build the SDK with this file does cargo download all three or just the one that is targeted?

Ivan-Velickovic:
Looking at the CI logs it downloads all three even though only one is needed. Is there any way around this?

nspin:
I don't think so. For reference, here are the sizes of each target's stdlib on my machine:

```
du -sh ~/.rustup/toolchains/1.79.0-x86_64-unknown-linux-gnu/lib/rustlib/{x8
6_64-unknown-linux-musl,x86_64-apple-darwin,aarch64-apple-darwin}
196M    x86_64-unknown-linux-musl
115M    x86_64-apple-darwin
110M    aarch64-apple-darwin
```

Ivan-Velickovic:
I'd prefer not to make people download any more than they need.

I'll try see if there's a work around.

Ivan-Velickovic:
I can't find anything. Maybe I'll open an issue at https://github.com/rust-lang/rustup/ and see what happens.

It doesn't seem like an edge-case though, I feel like it's reasonable for us to have a different default toolchain depending on the host.

The only thing I can think of is that we have three `rust-toolchain.toml` and have one per target, but that's just a hack.

Ivan-Velickovic:
Ah, what about something like `export RUSTUP_TOOLCHAIN=1.79.0-x86_64-unknown-linux-musl`? And `build_sdk.py` just exports the right thing depending on the host?

nspin:
That would work. Another option would be to have a toolchain file with no targets and use -Zbuild-std

Ivan-Velickovic:
Hmm. I get this when doing that export:
```
Error loading shared library libgcc_s.so.1: No such file or directory (needed by /home/ivanv/.rustup/toolchains/1.79.0-x86_64-unknown-linux-musl/bin/cargo)
Error relocating /home/ivanv/.rustup/toolchains/1.79.0-x86_64-unknown-linux-musl/bin/cargo: _Unwind_Resume: symbol not found
Error relocating /home/ivanv/.rustup/toolchains/1.79.0-x86_64-unknown-linux-musl/bin/cargo: _Unwind_GetIPInfo: symbol not found
Error relocating /home/ivanv/.rustup/toolchains/1.79.0-x86_64-unknown-linux-musl/bin/cargo: _Unwind_GetDataRelBase: symbol not found
Error relocating /home/ivanv/.rustup/toolchains/1.79.0-x86_64-unknown-linux-musl/bin/cargo: _Unwind_GetRegionStart: symbol not found
Error relocating /home/ivanv/.rustup/toolchains/1.79.0-x86_64-unknown-linux-musl/bin/cargo: _Unwind_SetGR: symbol not found
Error relocating /home/ivanv/.rustup/toolchains/1.79.0-x86_64-unknown-linux-musl/bin/cargo: _Unwind_FindEnclosingFunction: symbol not found
Error relocating /home/ivanv/.rustup/toolchains/1.79.0-x86_64-unknown-linux-musl/bin/cargo: _Unwind_GetTextRelBase: symbol not found
Error relocating /home/ivanv/.rustup/toolchains/1.79.0-x86_64-unknown-linux-musl/bin/cargo: _Unwind_DeleteException: symbol not found
Error relocating /home/ivanv/.rustup/toolchains/1.79.0-x86_64-unknown-linux-musl/bin/cargo: _Unwind_RaiseException: symbol not found
Error relocating /home/ivanv/.rustup/toolchains/1.79.0-x86_64-unknown-linux-musl/bin/cargo: _Unwind_GetIP: symbol not found
Error relocating /home/ivanv/.rustup/toolchains/1.79.0-x86_64-unknown-linux-musl/bin/cargo: _Unwind_Backtrace: symbol not found
Error relocating /home/ivanv/.rustup/toolchains/1.79.0-x86_64-unknown-linux-musl/bin/cargo: _Unwind_GetCFA: symbol not found
Error relocating /home/ivanv/.rustup/toolchains/1.79.0-x86_64-unknown-linux-musl/bin/cargo: _Unwind_GetLanguageSpecificData: symbol not found
Error relocating /home/ivanv/.rustup/toolchains/1.79.0-x86_64-unknown-linux-musl/bin/cargo: _Unwind_SetIP: symbol not found
```

I thought that the musl toolchain would have worked even though my Linux distro is glibc based.

Instead it looks like I need the `x86-64-unknown-linux-gnu` toolchain that has added the `x86-64-unknown-linux-musl` target added.

So if we do `export RUSTUP_TOOLCHAIN=1.79.0` that should work to pin the toolchain. But it means that sometimes people will need to manually run the command `rustup target add 1.79.0-x86-64-unknown-linux-musl` which is a bit annoying but I think that's the compromise that we settle on.

@nspin should we have a `rust-toolchain.toml` that doesn't specify any targets, or use the envar?

Ivan-Velickovic:
@nspin thinking of closing this one since the `Cargo.toml` file now specifies a minimum Rust version. It would be nice to have everyone use the same version with the `rust-toolchain.toml` file but the configuration available seems quite limiting right now.

It doesn't make sense for everyone having to download every single other toolchain. And in this case it's actually misleading because Rust can't compile from macOS to Linux or vice-versa.

What do you think?

Ivan-Velickovic:
I'll close this for the reasons in my comment above, if I missed anything please feel free to re-open. Right now it just seems that `rust-toolchain.toml` is a bit too limited for our use case.

nspin:
I agree. I think that specifying a MSRV basically solves the problem this issue is actually about without introducing any new problems like those related to the toolchain file that we’ve discussed in this thread

PR #202, opened 2024-08-16
Disable 'KernelArmVtimerUpdateVOffset' by default

Ivan-Velickovic:
This changes the behaviour of virtual machines to see the architectural timer change even when it is not running.

I do not understand why this option applies to all vCPUs in the kernel, instead of being behaviour that depends on the configuration of a particular vCPU.

I tried to learn more about this kernel configuration option in https://github.com/seL4/seL4/issues/1306 but no-one responded.


JorgeMVP:
@Ivan-Velickovic if it is false, does it mean we give to the VM the real-time notion? vcnt = pcnt?

do you know any other solution where this is configurable per vCPU?

PR #215, opened 2024-09-03
Add support for ARM SMC calls

Ivan-Velickovic:


Ivan-Velickovic:
It should be noted that this PR doesn't actually make enable any platforms to use this feature by default.

The problem is that most people probably don't want this feature enabled, although if you do want this feature you would have to change the `build_sdk.py` file and manually build the SDK. This isn't ideal and is a problem that will come up more in the future as certain other config options largely depend on the user itself and probably shouldn't be mandated by the SDK.

PR #217, opened 2024-09-04
Allow specifying notify or ppcalls on a per-channel, not per-protection domain basis

### LINKED ISSUE #168: Tutorial: part4 ppcall from VMM to less prioritized Wordle_server
Hi, thank you for the nice work :)
I just did the tutorial and wondered during part4. 
I have copy/pasted your `wordle.system` code into my config and found out that the PPC from `VVM` to` Wordle_Server` still worked even though it violated the priority callee-caller setup. VMM ist at prio 101 and Wordle_server at 25 and it works like a charm.

> A protected procedure call is only possible if the callee has strictly higher priority than the caller. This means that if A has a priority of 1 and B has a priority of 254, A can invoke B's protected procedure, but B cannot invoke A's protected procedure. (Tutorial description and documentation)

```
<!-- Create a VMM protection domain -->
    <protection_domain name="vmm" priority="101">
        <program_image path="vmm.elf" />
        <!-- [unchanged from tutorial code] -->
        <!-- possibly mistake in the template: id in VM missing; vcpu id unnecessary -->
        <virtual_machine name="linux" id="0" priority="100">
            <!-- [unchanged from tutorial code] -->
        </virtual_machine>
        <!-- [unchanged from tutorial code] -->
    </protection_domain>

    <protection_domain name="wordle_server" pp="true" priority="25">
        <program_image path="wordle_server.elf" />
    </protection_domain>

    <!-- vmm -> wordle server -->
    <channel>
        <!-- The VMM code expects the channel ID to be 1. -->
        <end pd="vmm" id="1" />
        <end pd="wordle_server" id="2" />
    </channel>
```

midnightveil:
Also fixes #168, as this allows us enough precision to check priorities.

Can specify `pp="true"` or `notify="false"` (or vice-versa, for the defaults):

```xml
<?xml version="1.0" encoding="UTF-8"?>
<system>
    <protection_domain name="test1">
        <program_image path="test" />
    </protection_domain>
    <protection_domain name="test2">
        <program_image path="test" />
    </protection_domain>

    <channel>
        <end pd="test1" id="0" pp="true" notify="false" />
        <end pd="test2" id="5" />
    </channel>
</system>
```

ToDo:
- [x] Parsing tests
- [x] Behaviour tests (it works on at least one system, but...)
- [x] Probably documentation of changes (where?)
- [x] Enforcing ppcall priority invariant
- [x] Syntax bikeshedding

### DIFF
--- a/tool/microkit/src/main.rs
+++ b/tool/microkit/src/main.rs
@@ -2860,29 +2860,27 @@ fn build_system(
         system_invocation.add_raw_invocation(config, &mut system_invocation_data);
     }
 
-    let mut pd_setvar_values: Vec<Vec<u64>> = vec![vec![]; system.protection_domains.len()];
-    for (i, pd) in system.protection_domains.iter().enumerate() {
-        for setvar in &pd.setvars {
-            assert!(setvar.region_paddr.is_some() || setvar.vaddr.is_some());
-            assert!(!(setvar.region_paddr.is_some() && setvar.vaddr.is_some()));
+    let pd_setvar_values: Vec<Vec<u64>> = system
+        .protection_domains
+        .iter()
+        .map(|pd| {
+            pd.setvars
+                .iter()
+                .map(|setvar| match &setvar.kind {
+                    sdf::SysSetVarKind::Vaddr { address } => *address,
+                    sdf::SysSetVarKind::Paddr { region } => {
+                        let mr = system
+                            .memory_regions
+                            .iter()
+                            .find(|mr| mr.name == *region)
+                            .unwrap_or_else(|| panic!("Cannot find region: {}", region));
 
-            let value;
-            if let Some(region_paddr) = &setvar.region_paddr {
-                let mr = system
-                    .memory_regions
-                    .iter()
-                    .find(|mr| mr.name == *region_paddr)
-                    .unwrap_or_else(|| panic!("Cannot find region: {}", region_paddr));
-                value = mr_pages[mr][0].phys_addr;
-            } else if let Some(vaddr) = setvar.vaddr {
-                value = vaddr;
-            } else {
-                panic!("Internal error: expected setvar to either have region paddr or vaddr");
-            }
-
-            pd_setvar_values[i].push(value);
-        }
-    }
+                        mr_pages[mr][0].phys_addr
+                    }
+                })
+                .collect()
+        })
+        .collect();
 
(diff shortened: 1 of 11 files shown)


tool/microkit/src/sdf.rs
```diff
@@ -115,14 +115,16 @@ pub struct SysIrq {
...
 }
 
-// The use of SysSetVar depends on the context. In some
-// cases it will contain a symbol and a physical or a
-// symbol and vaddr. Never both.
+#[derive(Debug, PartialEq, Eq, Hash)]
+pub enum SysSetVarKind {
+    Vaddr { address: u64 },
+    Paddr { region: String },
+}
```
Ivan-Velickovic, line 122: Nice, I did not know you could do this.

tool/microkit/src/main.rs
```diff
@@ -3130,6 +3130,9 @@ impl<'a> Args<'a> {
         if config.is_none() {
             missing_args.push("--config");
         }
+        if system.is_none() {
+            missing_args.push("system");
+        }
```
Ivan-Velickovic, line 3103: Good catch.

Ivan-Velickovic:
The first two commits look good, I'll review the main commit once the PR is ready for review.

tool/microkit/src/main.rs
midnightveil: I accidentally did that :)

tool/microkit/src/sdf.rs
midnightveil: Rust's sum types are great.

tool/microkit/src/sdf.rs
```diff
@@ -268,8 +275,13 @@ impl SysMap {
...
 
 impl ProtectionDomain {
-    pub fn needs_ep(&self) -> bool {
-        self.pp || self.has_children || self.virtual_machine.is_some()
+    pub fn needs_ep(&self, self_id: usize, channels: &[Channel]) -> bool {
+        self.has_children
+            || self.virtual_machine.is_some()
+            || channels.iter().any(|channel| {
+                (channel.end_a.can_ppcall_other && channel.end_b.pd == self_id)
+                    || (channel.end_b.can_ppcall_other && channel.end_a.pd == self_id)
```
midnightveil: I think this is right....

PR #220, opened 2024-09-20
aarch64/riscv64: make the microkit loader relocatable

alwin-joshy:
This allows a microkit image to be loaded anywhere in memory at which point it will relocate itself to the LOADER_LINK_ADDRESS. These changes are based on the elfloader from seL4/seL4_tools.

Changes include:

* crt0.s was changed to crt0.S and compiled with gcc to pass in LINK_ADDRESS
* memmove and memcpy implementations were added
* loader_data was changed to include a size field. This is the size of loader_data, the list of regions, and the regions themselves. These also need to be relocated to ensure that everything works correctly

Tested on: imx8mm with LOADER_LINK_ADDRESS = 0x41000000 and uboot load address parameter set to 0x42000000


alwin-joshy:
Added support for RISC-V. Tested on star64 by loading image at address 0x70000000 (LOADER_LINK_ADDRESS is 0x60000000)

tool/microkit/src/loader.rs
```diff
@@ -108,6 +108,7 @@ struct LoaderRegion64 {
 
 #[repr(C)]
 struct LoaderHeader64 {
+    size: u64,
     magic: u64,
     flags: u64,
     kernel_entry: u64,
```
Ivan-Velickovic: The magic should always be first.

Ivan-Velickovic (changes requested):
This would be great to get in before the next release. Main thing that would be good to change is to have the loader print if it is relocating itself and to where within memory it is relocating itself. This kind of thing tends to be fragile and so its good to print that info in debug mode.

So, the first thing the loader should do is still jump to main where it decides whether it needs to relocate itself or not, and print appropriately.

PR #254, opened 2024-12-17
tool/microkit: fix SchedContext allocation size

matneutrality:
When creating CNodes, the `size` is the (log2) number of CSlots. When creating SchedContexts the `size` is the (log2) size of the object in bytes.


Ivan-Velickovic (approved):
Good catch, thank you!

Issue #174, opened 2024-07-15
Complain at a microkit level of IRQ ACK is called on a non-irq channel


midnightveil:
Currently it's just a message from the kernel that it was invoked on a null cap which isn't obvious.

wucke13:
This got me doing the tutorial as well, while logical when all facts are known, this did surprise at the beginning.

PR #282, opened 2025-03-04
libmicrokit: more error reporting

Ivan-Velickovic:
This patch adds error reporting for libmicrokit APIs that make seL4 system calls that do not have an error return value and so the user does not know what went wrong. There is an error message from the kernel, but it refers to cap numbers which a Microkit user would not understand.

Closes https://github.com/seL4/microkit/issues/174.


Ivan-Velickovic:
> Maybe only do this in debug mode?

It's probably fine either way but I would prefer to have the same behaviour on release/debug mode which is if it's valid it leads to a system call, if it's not then it doesn't.

PR #298, opened 2025-03-16
github: use central CI workflows

lsf37:
Use GitHub `workflow_call` feature to reduce workflow duplication. Runs the same checks as before.

The addition of `workflow_dispatch` is for being able to start checks manually. 

~~When this is merged, we'll need to update the "required checks"  names in the repo settings.~~ Looks like none of these were set to required before.

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -6,32 +6,12 @@
 
 name: PR
 
-on: [pull_request]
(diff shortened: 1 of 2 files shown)


lsf37:
The link check failure is independent of this change and being being fixed in https://github.com/seL4/ci-actions/pull/392

Ivan-Velickovic (approved):

PR #304, opened 2025-04-09
Enable "KernelAllowSMCCalls" by default on ARM

Ivan-Velickovic:
Previously I did not turn this option on by default because I thought it would make the kernel proofs not apply anymore. In reality this is only true really when a user makes use of the SMC feature. Most users probably will never have to enable it and the ones that that do will no longer have to patch Microkit to enable it.

I've added a warning about using SMC calls in the manual to make users aware of the implications of using this feature.

### DIFF
--- a/build_sdk.py
+++ b/build_sdk.py
@@ -93,6 +93,7 @@ SUPPORTED_BOARDS = (
             "KernelArmExportPCNTUser": True,
             "KernelArmHypervisorSupport": True,
             "KernelArmVtimerUpdateVOffset": False,
+            "KernelAllowSMCCalls": True,
         },
     ),
     BoardInfo(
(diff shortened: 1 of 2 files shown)


lsf37:
To be entirely pedantic, the kernel proofs do currently indeed not apply when this option is on, but I'm in the process of verifying the feature and they should soon.

Ivan-Velickovic:
> To be entirely pedantic, the kernel proofs do currently indeed not apply when this option is on, but I'm in the process of verifying the feature and they should soon.

Okay thanks. Right now the Microkit doesn't really talk about the relationship between the kernel's it supplies and what is verified, I'll do that in the future https://github.com/seL4/microkit/issues/305.

PR #308, opened 2025-05-28
use the initial task untypes for no kernel patch

### LINKED ISSUE #52: Allowing Microkit to not require a kernel patch
There is currently a single patch to the seL4 kernel (https://github.com/seL4/seL4/commit/c8ef493d038b81cc43f82c0190788e4b3bdb4d9d) that Microkit depends on.

**What is the patch for?**

The patch allows the boot-loader, which runs before seL4, to specify an extra region of device untyped at run-time when the boot-loader jumps to seL4.

**Why is the patch necessary?**

The boot-loader that Microkit systems use places all binaries in the expected places in physical memory *before* jumping to seL4. This includes binary for seL4 itself, the initial task (aka the monitor), as well as the binary for each protection domain.

However, when seL4 starts, it clears all physical memory that it defines as 'normal'. This means all memory in RAM (except seL4 itself and the initial task) will be cleared. Without the kernel patch, this would include the PD ELFs, meaning that when the PDs start, all the code for them will be gone.

In order to avoid this, we place all the PD ELFs in a contiguous region of physical memory that we tell seL4 at boot is 'device' memory, meaning that it *won't* clear it and hence the PD ELFs remain in tact after seL4 has booted.

**How will the patch be removed?**

Assuming there are no blocking issues, the new Rust capDL initialiser can be used to embed the binaries for each PD in the initialiser itself, while not copying the binaries around in memory. This allows us to have the memory efficiency desired without a patch to the kernel.

After the kernel patch is removed, a specific branch of seL4 will no longer be required. Note that Microkit may be pinned to a specific version of seL4 still, so not just any commit of the kernel will work.

midnightveil:
Previously, microkit used a kernel patch which added two extra parameters to
the seL4 kernel entry code: `extra_device_addr_p` and `extra_device_size`. This
allowed for microkit to pass the protection domains ELFs and system invocation
data to the initial task without the data being zeroed by seL4 upon an Untyped
Retype invocation. This was done by making it appear as extra device untypes.

That had several downsides. For one, having a kernel patch is not a great
situation to live in. Furthermore, on certain platforms, e.g. RISC-V SMP, it
would be impossible to add these two extra parameters as it would require extra
C-ABI registers which don't exist.

Instead, what this patch does, is pass the extra data as part of the initial
task region. This is fine, as it was *already* contiguous with the initial task
(monitor) ELF file; being present in memory just before it.

What this looks like from the initial task's perspective is that it gets
capabilities to Frames in the seL4_BootInfo data. These frames are for the
initial task's memory. As they are already frame objects, we don't have to do
untype retypes for these objects. As they are already mapped into the initial
task memory, we also don't have to create page tables mappings via the monitor
bootstrap invocations.

Fixes #52 

---

- [ ] Can we verify that the initial task frames are for the correct physical addresses (i.e. that they match what we expected?). I had an off-by-1 bug when developing which took a while to figure out; I used an seL4 patch that made DebugCapIdentify print out the `cap_getAddr` value but that's not viable in general.
- [ ] Does the code structure make sense? I don't know if the way that I wrote the allocation of the frames makes sense.
- [x] Test all the sDDF examples as a regression test

### DIFF
--- a/monitor/src/debug.c
+++ b/monitor/src/debug.c
@@ -95,6 +95,25 @@ void dump_bootinfo(seL4_BootInfo *bi)
     puthex64(bi->extraBIPages.end - 1);
     puts("\n");
 
+#if 0
+    for (i = bi->userImageFrames.start; i < bi->userImageFrames.end; i++) {
+        // seL4_DebugCapIdentify(i);
+        // puts("userImageFramesList[");
+        // puthex32(i);
+        // puts("]        = slot: ");
+        // puthex32(bi->userImageFrames.start + i);
+        // puts(", paddr: ");
+        // puthex64(bi->untypedList[i].paddr);
+        // puts(" - ");
+        // puthex64(bi->untypedList[i].paddr + (1UL << bi->untypedList[i].sizeBits));
+        // puts(" (");
+        // puts(bi->untypedList[i].isDevice ? "device" : "normal");
+        // puts(") bits: ");
+        // puthex32(bi->untypedList[i].sizeBits);
+        // puts("\n");
+    }
+#endif
+
 #if 1
     for (i = 0; i < bi->untyped.end - bi->untyped.start; i++) {
         puts("untypedList[");
--- a/tool/microkit/src/loader.rs
+++ b/tool/microkit/src/loader.rs
@@ -116,8 +116,6 @@ struct LoaderHeader64 {
     ui_p_reg_end: u64,
     pv_offset: u64,
     v_entry: u64,
-    extra_device_addr_p: u64,
-    extra_device_size: u64,
     num_regions: u64,
 }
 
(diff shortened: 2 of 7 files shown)


monitor/src/debug.c
```diff
@@ -95,6 +95,25 @@ void dump_bootinfo(seL4_BootInfo *bi)
...
+        // puthex64(bi->untypedList[i].paddr);
+        // puts(" - ");
+        // puthex64(bi->untypedList[i].paddr + (1UL << bi->untypedList[i].sizeBits));
+        // puts(" (");
+        // puts(bi->untypedList[i].isDevice ? "device" : "normal");
+        // puts(") bits: ");
+        // puthex32(bi->untypedList[i].sizeBits);
+        // puts("\n");
+    }
+#endif
```
Ivan-Velickovic, line 115: Do you want to make this uncommented?

tool/microkit/src/loader.rs
```diff
@@ -193,26 +193,26 @@ impl<'a> Loader<'a> {
...
+        let monitor_last_vaddr = round_up(segment.virt_addr + segment.mem_size(), kb(4));
 
-        let inittask_first_paddr = match initial_task_phys_base {
-            Some(paddr) => paddr,
-            None => segment.phys_addr,
-        };
-        let inittask_p_v_offset = inittask_first_vaddr - inittask_first_paddr;
+        // TODO: Why are we even rederiving it here, tbh?
+        assert_eq!(monitor_virt_subregion.base, monitor_first_vaddr);
+        assert_eq!(monitor_virt_subregion.end, monitor_last_vaddr);
```
Ivan-Velickovic, line 206: Retrieving `monitor_first_vaddr` do you mean?

monitor/src/debug.c
midnightveil: No, it doesn't work because there is no userImageFrames list. I *don't think* we get told about what physical addresses the user image frames are for. This is the TODO list item № 1.

tool/microkit/src/loader.rs
midnightveil: Yeah, since `monitor_virt_subregion.base` was derived from the monitor ELF first vaddr *anyway*, I don't know why we load it again here.

(also, not a typo: re-deriving)

PR #316, opened 2025-07-17
Patch page tables of child PD's into their parent

Kswin01:
This PR adds support for patching a copy of a child PD's page tables into the parent.

We add an extra attribute to the protection domain xml tag, if set to `true`, the child page tables will be added into the parent.

We create a vector for all potential child page tables, with the length of the total amount of PDs in the system. This vector is a vector of 64 references to potential PGD's of potential child PDs. If the above tag is set to true, we will go through the current elf segments, memory regions and the ipc buffer associated with each of the child PDs, and create a copy of this paging structure and add to the above vector. At this point, the underlying frame caps haven't been allocated. We will add a blank elf segment, `.table_data` to the parent elf at this point, of the correct size that we have calculated from creating these paging structures.

Later on when these frame caps have been allocated,  its a simple matter of filling in the already create paging structures. To make this simpler, we have added some helper structs and functions in `lib.rs`. *I'm not sure if this is the best place for this code to live*. 

We then populate the elf segment we created before with these complete paging structures. Additionally we patch a struct variable into the parent containing the base address of this new segment, as well as an array of 64 pointers into this region that are the start of the PGD of each child. We haven't added this struct to the microkit header, so it is therefore up to the user to define this structure. Please let me know if we should add this to `microkit.h` instead. The entries in this table are pointers to the next level tables, or the frame cap. We support both large and small pages. This has been tested on aarch64, but should also work on RISCV64 provided that the address are Sv48 (48bit addressing). The caps to the underlying frames are also minted into the parent's cspace. 

Additionally, we extend the current elf library in the microkit tool to be able to create and populate new segments to the in-memory copy of the supplied elf's.

Please feel free to provide any feedback!


tool/microkit/src/elf.rs
```diff
@@ -363,4 +373,47 @@ impl ElfFile {
...
+
+    pub fn create_segment(&mut self, segment_name: &str, size: u64) -> ElfSegment {
+        let mut last_addr = 0;
+        for segment in self.segments.iter() {
+            if segment.virt_addr > last_addr {
+                last_addr = segment.virt_addr + segment.data.len() as u64;
+            }
+        }
+
+        last_addr = last_addr + (0x10000 - (last_addr % 0x10000));
```
midnightveil: Why `0x10000`? Is this supposed to be for a 2MiB large page? If so it should probably be an arch-specific constant.

tool/microkit/src/elf.rs
```diff
@@ -363,4 +373,47 @@ impl ElfFile {
...
+                if name == segment_name {
+                    assert!(data.len() <= segment.data.len());
+                    segment.data = data.to_vec();
+                }
+            }
+        }
+    }
+
+    pub fn get_segment(&mut self, segment_name: &str) -> Option<&ElfSegment> {
+        for segment in &self.segments {
```
midnightveil: Why not just `self.segments.find(|segment| segment.name == Some(segment_name))`?

tool/microkit/src/elf.rs
```diff
@@ -363,4 +373,47 @@ impl ElfFile {
...
+            phys_addr: last_addr,
+            virt_addr: last_addr,
+            loadable: true,
+            attrs: ElfSegmentAttributes::Read as u32,
+        }
+    }
+
+    pub fn populate_segment(&mut self, segment_name: &str, data: &[u8]) {
+        // Find segment
+        for segment in self.segments.iter_mut() {
```
midnightveil, line 400: This can just use get_segment if it returned a `&mut ElfSegment`

tool/microkit/src/lib.rs
```diff
@@ -24,6 +26,235 @@ pub const MAX_VMS: usize = 63;
...
+            puds: vec![None; 512],
+        }
+    }
+
+    pub fn recurse(&mut self, mut curr_offset: u64, buffer: &mut Vec<u8>) -> u64 {
+        let mut offset_table: [u64; 512] = [u64::MAX; 512];
+        for (i, entry) in offset_table.iter_mut().enumerate() {
+            if let Some(pud) = &mut self.puds[i] {
+                curr_offset = pud.recurse(curr_offset, buffer);
+                *entry = curr_offset - (512 * 8);
```
midnightveil: These constants should definitely be parametrised over the architecture

tool/microkit/src/lib.rs
```diff
@@ -24,6 +26,235 @@ pub const MAX_VMS: usize = 63;
...
+            }
+        }
+
+        for value in &mut offset_table {
+            buffer.append(&mut value.to_le_bytes().to_vec());
+        }
+        curr_offset + (512 * 8)
```
midnightveil: ditto

tool/microkit/src/lib.rs
```diff
@@ -24,6 +26,235 @@ pub const MAX_VMS: usize = 63;
...
+    }
+
+    pub fn add_page_at_vaddr(&mut self, vaddr: u64, frame: u64, size: PageSize) {
+        let pgd_index = ((vaddr & (0x1ff << 39)) >> 39) as usize;
```
midnightveil: ditto

tool/microkit/src/lib.rs
```diff
@@ -24,6 +26,235 @@ pub const MAX_VMS: usize = 63;
...
+
+#[derive(Clone, Debug, PartialEq, Eq, Hash)]
+pub struct PUD {
+    dirs: Vec<Option<DIR>>,
+}
+
+impl PUD {
+    fn new() -> Self {
+        PUD {
+            dirs: vec![None; 512],
```
midnightveil: ditto

tool/microkit/src/lib.rs
```diff
@@ -24,6 +26,235 @@ pub const MAX_VMS: usize = 63;
...
+        }
+    }
+
+    fn recurse(&mut self, mut curr_offset: u64, buffer: &mut Vec<u8>) -> u64 {
+        let mut offset_table: [u64; 512] = [u64::MAX; 512];
+        for (i, entry) in offset_table.iter_mut().enumerate() {
+            if let Some(dir) = &mut self.dirs[i] {
+                curr_offset = dir.recurse(curr_offset, buffer);
+                *entry = curr_offset - (512 * 8);
```
midnightveil: ditto

tool/microkit/src/main.rs
```diff
@@ -759,14 +763,81 @@ fn build_system(
...
+                }
+                // Find all associated memory regions and their size.
+                for child_map in &child_pd.maps {
+                    // Find the memory region associated with this map
+                    for mr in &system.memory_regions {
+                        if mr.name == child_map.mr {
+                            child_pgd.add_page_at_vaddr_range(
+                                child_map.vaddr,
+                                mr.size as i64,
+                                1,
```
midnightveil: what is this 1?

tool/microkit/src/main.rs
```diff
@@ -1971,6 +2042,101 @@ fn build_system(
...
+
+            for i in child_pds {
+                offset =
+                    all_child_page_tables[pd_idx].as_mut().unwrap()[i].recurse(offset, &mut table_data);
+                table_metadata.pgd[i] = offset - (512 * 8);
+            }
+
+            // patch the data in
+            let elf = &mut pd_elf_files[pd_idx];
+            elf.populate_segment(".table_data", &table_data);
```
midnightveil, line 2200: We had this discussion in person but hard NACK on actually editing the ELF files directly to do this, as opposed to doing this via adding a new segment into the memory regions the tool adds directly.

tool/microkit/src/elf.rs
Kswin01: This was supposed to just be 0x1000 to align to a small page. Will change this to a constant

tool/microkit/src/main.rs
Kswin01: The 1 is a dummy value in place of the actual frame. This is because we needed to compute the size of these tables before the page caps are allocated. I'll add a comment for this

PR #324, opened 2025-08-27
Added builds for extra rpi4B models

Kacper-M-Michalik:
This PR adds Microkit builds for 2gb, 4gb and 8gb versions of the rpi4B.

All rpi4B models have same mini-UART MMIO layout, so we use the same code in loader.c.
Builds successfully on Ubuntu using nix.


Ivan-Velickovic:
Thanks!

PR #327, opened 2025-09-02
loader: force strict alignment

midnightveil:
It appears that in QEMU SCTLR_EL2.A (Alignment Check Enable) is off by default, and certain code patterns will cause unaligned accesses in e.g. puthex32() which uses a stack-allocated char[].

Forcing -mstrict-align means we should always generate code that respects the alignment requirements. We could have instead set the A bit in EL2, but this is just easier.

Also, we don't appear to have an EL2 fault handler in the loader, so these faults just silently crash the loader.

### DIFF
--- a/loader/Makefile
+++ b/loader/Makefile
@@ -40,7 +40,7 @@ $(error PRINTING must be specified)
 endif
 
 ifeq ($(ARCH),aarch64)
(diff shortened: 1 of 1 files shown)


Ivan-Velickovic:
> Also, we don't appear to have an EL2 fault handler in the loader, so these faults just silently crash the loader.

Hmm we should fix that then.

PR #337, opened 2025-10-30
capDL and x86_64 support in Microkit

dreamliner787-9:
This PR re-implemented the Microkit to be capDL based and added support for the x86_64 architecture (x86 VMs are supported but experimental).

seL4 remains pinned on the `microkit` branch to reduce the number of changes in this PR. We can switch to the `master` branch in the near future as we no longer rely on the extra device untypeds logic.

Breaking changes:
- Highest PD priority is now 253 rather than 254 to accomodate the capDL initialiser.

Otherwise, the capDL changes are unintrusive to end-users of Microkit.

There are a couple of rough edges to fix, but making a PR so we can test the CI.

### DIFF
--- a/tool/microkit/src/sdf.rs
+++ b/tool/microkit/src/sdf.rs
@@ -1593,16 +1593,16 @@ pub fn parse(filename: &str, xml: &str, config: &Config) -> Result<SystemDescrip
         }
     }
 
-    let mut vms = vec![];
+    let mut vms: Vec<&String> = vec![];
     for pd in &pds {
         if let Some(vm) = &pd.virtual_machine {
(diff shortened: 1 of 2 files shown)


Ivan-Velickovic (approved):
These changes have been reviewed/tested internally. There are various things to still fix/improve before we make a release. We are currently aiming for a release including these changes by the end of November. An issue with the list of things we are working on fixing/improving before the next release will be posted soon.

With the integration of capDL, I believe the Microkit RFC can move from 'active' to 'implemented'.

Other than the minor breaking change listed in the PR description, users should not notice any functional differences. The transition to using capDL for system initialisation is purely an internal change.

PR #338, opened 2025-10-30
tool: add support for `setvar_id`

nspin:
The `setvar_id` attribute for the `<end>` and `<irq>` elements is like the `setvar_vaddr` and `setvar_size` attributes of the `<map>` element. It injects the channel ID of the `<end>` or `<irq>` into the specified variable.

### DIFF
--- a/tool/microkit/src/sdf.rs
+++ b/tool/microkit/src/sdf.rs
@@ -239,6 +239,8 @@ pub struct ProtectionDomain {
     /// Index into the total list of protection domains if a parent
     /// protection domain exists
     pub parent: Option<usize>,
+    /// Value of the setvar_id attribute, if a parent protection domain exists
+    pub setvar_id: Option<String>,
     /// Location in the parsed SDF file
     text_pos: Option<roxmltree::TextPos>,
 }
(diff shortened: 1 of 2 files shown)


tool/microkit/src/sdf.rs
```diff
@@ -897,6 +916,28 @@ impl ChannelEnd {
             })?;
 
         if let Some(pd_idx) = pds.iter().position(|pd| pd.name == end_pd) {
+            let pd = &mut pds[pd_idx];
+
+            if let Some(setvar_id) = node.attribute("setvar_id") {
+                // Check that the symbol does not already exist
+                for setvar in &pd.setvars {
```
midnightveil: Doing it this way (threading mutable pds down into the parsing of the IRQ and then channel elements) has a few downsides:

- needing to track this in several places
- needing to make this mutable at all
- intermixing parsing and validation

I think it would personally make more sense if the instantiation of the setvars for these elements was done *after* parsing of all the elements. The benefit of this are a few:

- One place where all this logic is (rather than duplicated in both IRQs and Channels).
- You can print more useful error information, including a list of *anything* that conflicted, instead of just the first conflict.
- (Not really that important) would run faster as you don't need to do a for loop everytime you add something

We already have code in the sdf.rs file that does this sort of validation. See e.g. L1285, just at the end of the PR preview I can see, which says "we can valid any global properties e.g. duplicate IDs". This is that kind of property.

nspin: I've updated the PR taking this into account. For the elements that work like the ones with existing `setvar_{vaddr,size}` attributes, I've implemented their `setvar_id` attribute handling to match what was already there. Though I did factor out a helper function for doing the checks.

For elements whose parser functions are passed a reference to a PD, I added a `setvar_id` field to their parsed forms, and added them to the appropriate `setvars` list after parsing.

What do you think of this change?

Ivan-Velickovic (approved):
Cool, thanks.

PR #342, opened 2025-11-05
tool: deal with breaking Rust change

Ivan-Velickovic:
Unfortunately the capDL initialiser has to make use of unstable Rust features which are able to have breaking changes at any time.

The custom Rust target JSON files had a breaking change in commit 0e28b4201ac8f12cfafe5383f442c7b00f8bdcc6 of the Rust compiler.

Instead of doing:
"target-pointer-width": "64"

we now have to do:
"target-pointer-width": 64

This breaking change means that people using 1.90.0 or earlier of the Rust compiler cannot compile the Microkit SDK from source and so we have to pin the minimum Rust version of the to 1.91.0 which is not particularly ideal.

### DIFF
--- a/.github/workflows/sdk.yaml
+++ b/.github/workflows/sdk.yaml
@@ -33,6 +33,7 @@ jobs:
             path: seL4
       - name: Install SDK dependencies
         run: |
+          rustup install 1.91.0
           rustup target add x86_64-unknown-linux-musl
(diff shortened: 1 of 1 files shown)


Ivan-Velickovic:
Actually let me see if I can get rid of the initialiser not have a custom target in the first place, this would remove all this complexity in the first place.

PR #343, opened 2025-11-05
Remove Rust custom targets for initialiser

Ivan-Velickovic:
The motivation for this is that the Rust custom targets are an unstable feature which means it is subject to change at any point in the forseeable future. This just happened with Rust 1.91.0 which changed the format of target-pointer-width which lead to the Microkit SDK failing to build if you had a too new Rust compiler but then if we fix it for that anyone that doesn't have the latest verison of Rust then we break the SDK build for them.

I can't see anything special about these targets so the embedded no OS targets that already exist for Rust should be appropriate.


midnightveil:
Why was it not the best approach?

Ivan-Velickovic:
> If you can do it without an unstable feature with no downsides why not?

x86-64 targets require some changes, `x86_64-unknown-none` does not work directly. There are also some minor other things like passing the page size explicitly that rust-sel4 does.

To replicate the targets without JSON I think we'd need to use unstable features anyways?

This made me think that it is not worth it.

PR #345, opened 2025-11-05
Pin Rust toolchain

Ivan-Velickovic:
Before 9d6b4e09663d4192b00d1c219d28d2b299659d55, we used Rust for the Microkit tool. Since that commit, we now rely on Rust for a runtime component as well, the capDL initialiser as part of the rust-sel4 project.

This use of Rust is quite different to the tool, in that it relies on various unstable features of Rust in order to make use of a custom Rust language runtime. At the time of writing, this makes it unrealistic to use only stabilised features of Rust.

These unstable features are subject to change anytime, which means that allowing users to build the SDK from source with any toolchain has lead to various issues and will likely lead to more as more versions of Rust are released. Not pinning a toolchain was realistic before, because the Microkit tool only made use of stable Rust features.

The Rust eco-system makes use of the rust-toolchain.toml to handle this situation as it allows for pinning the default toolchain and automatically donwload it when trying to compile.

The main issue with this approach is it involves more downloading, especially as all the targets you want to support must be in the 'targets' field from what I understand. This is especially annoying as it means that you have to specify all the targets you *might* want to compile for rather than the ones you actually care about. For example, the CI compiles from a Linux host for x86-64/aarch64 Linux and from a macOS host for x86-64/aarch64 macOS, but you cannot express this kind of situation in the rust-toolchain.toml. This means everyone has to download 4 toolchains because there's 4 targets that we can possibly compile for.

This situation mainly exists due to complexity involved for using the Rust toolchain and the fact that the components and the targets cannot just be one binary you donwload unlike other compilers/toolchains.

### DIFF
--- a/.github/workflows/sdk.yaml
+++ b/.github/workflows/sdk.yaml
@@ -33,10 +33,6 @@ jobs:
             path: seL4
       - name: Install SDK dependencies
         run: |
-          rustup target add x86_64-unknown-linux-musl
-          rustup component add rust-src --toolchain stable-x86_64-unknown-linux-gnu
-          rustup target add aarch64-unknown-linux-musl
-          rustup component add rust-src --toolchain stable-aarch64-unknown-linux-musl
           sudo apt update
           sudo apt install software-properties-common
           sudo add-apt-repository ppa:deadsnakes/ppa
--- a/build_sdk.py
+++ b/build_sdk.py
@@ -760,12 +760,16 @@ def build_initialiser(
     component_build_dir = build_dir / board.name / config.name / component_name
     component_build_dir.mkdir(exist_ok=True, parents=True)
 
+    toolchain_setup = "cd tool/microkit && rustup toolchain install && cd ../../"
+
     if custom_rust_sel4_dir is None:
         capdl_init_elf = component_build_dir / "bin" / "sel4-capdl-initializer.elf"
         cmd = f"""
+            {toolchain_setup} && \
             RUSTC_BOOTSTRAP=1 \
-            RUST_TARGET_PATH={rust_target_path} SEL4_PREFIX={sel4_src_dir.absolute()} \
-            cargo install {cargo_cross_options} \
+            RUST_TARGET_PATH={rust_target_path} \
+            SEL4_PREFIX={sel4_src_dir.absolute()} \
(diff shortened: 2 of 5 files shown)


dreamliner787-9:
> Despite `x86_64-unknown-linux-gnu` and `rust-src` being in the `rust-toolchain.toml`, I now get this error:
> 
> ```
> 
> error: failed to compile `sel4-capdl-initializer v0.1.0 (https://github.com/au-ts/rust-seL4?branch=capdl_dev#118c9cd6)`, intermediate artifacts can be found at `/home/runner/work/microkit/microkit/build/initialiser`.
> To reuse those artifacts with a future compilation, set the environment variable `CARGO_TARGET_DIR` to that path.
> 
> Caused by:
>   "/home/runner/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/Cargo.lock" does not exist, unable to build with the standard library, try:
>           rustup component add rust-src --toolchain stable-x86_64-unknown-linux-gnu
> Traceback (most recent call last):
> Building sel4: sel4_dir=PosixPath('seL4') root_dir=PosixPath('release/microkit-sdk-2.0.1-dev.86+a690a5e-linux-x86-64') build_dir=PosixPath('build/tqma8xqp1gb/release/sel4') board=BoardInfo(name='tqma8xqp1gb', arch=<KernelArch.AARCH64: 1>, gcc_cpu='cortex-a35', loader_link_address=2150105088, kernel_options={'KernelPlatform': 'tqma8xqp1gb', 'KernelArmExportPCNTUser': True, 'KernelArmHypervisorSupport': True, 'KernelArmVtimerUpdateVOffset': False, 'KernelAllowSMCCalls': True, 'KernelIsMCS': True, 'KernelRootCNodeSizeBits': '17'}) config=ConfigInfo(name='release', debug=False, kernel_options={}, kernel_options_arch={})
> ```

I think you got that error because the step of building the initialiser is independant of `tool/microkit/rust-toolchain.toml`. Try running `rustup component add rust-src --toolchain stable-x86_64-unknown-linux-gnu` in `build_initialiser()` of build_sdk.py. I don't have push permission to this branch so I can't test this theory.

PR #346, opened 2025-11-06
build_sdk: make cargo use deps from lockfile 

midnightveil:
Otherwise the versions of dependencies can vary.

### DIFF
--- a/build_sdk.py
+++ b/build_sdk.py
@@ -510,7 +510,7 @@ def test_tool() -> None:
 
 def build_tool(tool_target: Path, target_triple: str) -> None:
     r = system(
-        f"cd tool/microkit && cargo build --release --target {target_triple}"
+        f"cd tool/microkit && cargo build --release --locked --target {target_triple}"
     )
     assert r == 0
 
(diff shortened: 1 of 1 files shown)


Ivan-Velickovic:
Should this apply to the tool itself? When we do `cargo build` on the Microkit tool?

nspin:
> If you have any ideas around how we could not have duplication with the version of rust-sel4 we pin to without using something like a submodule that would be appreciated (we have to pin the version in both build_sdk.py and tool/microkit/Cargo.toml which isn't ideal).

Just a note to add here, one reason why it is so important that the same rust-sel4 version is used for both the tool and the initializer is because, while the JSON CapDL spec format is well-defined and stable, the zero-copy binary serialization format that the tool and initializer agree on depends on the version of the [rkyv](https://rkyv.org/) library used by both. So, it is very important that the same version of rust-sel4 is used for the tool and the initializer.

I bring this up because of the existence of this argument:

https://github.com/seL4/microkit/blob/fa7f1ee6fe9b2eaa91274a64d3cb014aa13dfdac/build_sdk.py#L798

Which might be a footgun.

nspin:
If you’re talking about leveraging [workspace dependencies](https://doc.rust-lang.org/cargo/reference/workspaces.html#the-dependencies-table) to ensure that both the tool package and the local dummy CapDL initializer package use the same git rev of the rust-sel4 repo in the stable scheme I described, then yes, that would be the right way to do it.

We’d still need to make that minor change to the structure of the upstream CapDL initializer package (splitting it into a [lib] and a dummy [bin]) and adding a dummy initializer [bin] to the tool workspace.

I was a bit unclear about the stable scheme above, the local dummy CapDL initializer [bin] would be in a separate package (e.g. called capdl_initialiser) in the same workspace as the Microkit tool package.

I can make a PR on Monday (North America Monday, that is).

PR #347, opened 2025-11-07
cleanups for the loader

midnightveil:
This involves moving code from `#ifdef ARCH_` to instead arch directories.

Having one single file for the whole loader was getting very unwieldy.

The other feature here is that we distinguish between the name of the sdk folder ('sdk name') and the name of the board ('board name'), this makes the raspberry pi 4 boards much nicer, and will be useful for SMP and multikernel support as we don't need to add more defines for every variant of the same board.

### DIFF
--- a/loader/src/aarch64/init.c
+++ b/loader/src/aarch64/init.c
@@ -9,15 +9,17 @@
 #include "../arch.h"
 #include "el.h"
 
-#if defined(BOARD_zcu102) || defined(BOARD_ultra96v2)
+#include <kernel/gen_config.h>
+
+#if defined(CONFIG_PLAT_ZYNQMP_ZCU102) || defined(CONFIG_PLAT_ZYNQMP_ULTRA96V2)
 #define GICD_BASE 0x00F9010000UL
 #define GICC_BASE 0x00F9020000UL
-#elif defined(BOARD_qemu_virt_aarch64)
+#elif defined(CONFIG_PLAT_QEMU_ARM_VIRT)
 #define GICD_BASE 0x8000000UL
 #define GICC_BASE 0x8010000UL
 #endif
 
-#if defined(BOARD_zcu102) || defined(BOARD_ultra96v2) || defined(BOARD_qemu_virt_aarch64)
+#if defined(CONFIG_PLAT_ZYNQMP_ZCU102) || defined(CONFIG_PLAT_ZYNQMP_ULTRA96V2) || defined(CONFIG_PLAT_QEMU_ARM_VIRT)
 static void configure_gicv2(void)
 {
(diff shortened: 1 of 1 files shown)


Ivan-Velickovic:
That would also let us remove extra configuration like `FLAG_SEL4_HYP`.

Ivan-Velickovic:
> Having one single file for the whole loader was getting very unwieldy.

Yes especially with the different architecture support, thanks for cleaning this up I've been meaning to do it for ages.

loader/src/aarch64/init.c
```diff
@@ -0,0 +1,70 @@
...
+ * Copyright 2025, UNSW
+ *
+ * SPDX-License-Identifier: BSD-2-Clause
+ */
+
+#include "../uart.h"
+#include "../arch.h"
+#include "el.h"
+
+#if defined(BOARD_zcu102) || defined(BOARD_ultra96v2)
```
Ivan-Velickovic: Are the `BOARD_` uses in this file intentional?

loader/Makefile
```diff
@@ -51,29 +51,38 @@ else ifeq ($(ARCH),riscv64)
 	ARCH_DIR := riscv
 endif
 
-CFLAGS := -std=gnu11 -g -O3 -nostdlib -ffreestanding $(CFLAGS_ARCH) -DBOARD_$(BOARD) -DPRINTING=$(PRINTING) -Wall -Werror -Wno-unused-function
+CFLAGS := -std=gnu11 -g -O3 -nostdlib -ffreestanding -MP -MD $(CFLAGS_ARCH) -DPRINTING=$(PRINTING) -Wall -Werror -I$(SEL4_SDK)/include
```
Ivan-Velickovic: I think we can remove `-DPRINTING` now?

midnightveil: In preference to `CONFIG_PRINTING`, I'm guessing?

loader/src/loader.h
```diff
@@ -0,0 +1,37 @@
...
+ * SPDX-License-Identifier: BSD-2-Clause
+ */
+
+#include <stdint.h>
+#include <stddef.h>
+
+#define REGION_TYPE_DATA 1
+#define REGION_TYPE_ZERO 2
+
+#define FLAG_SEL4_HYP (1UL << 0)
```
Ivan-Velickovic: Should we keep this or rely on `CONFIG_ARM_HYPERVISOR_SUPPORT` instead?

midnightveil: ~~I'm too lazy but we probably should (we'd need to decide to change the struct that the tool patches in as well).~~ Fixed

loader/Makefile
Ivan-Velickovic: Yep

PR #348, opened 2025-11-10
libmicrokit: add x86_64 vCPU invocations

dreamliner787-9:
Add x86_64 vCPU invocations for x86_64 VM support implemented in https://github.com/seL4/microkit/pull/337.

As of this PR, x86_64 VM support in Microkit is still experimental.

### DIFF
--- a/docs/manual.md
+++ b/docs/manual.md
@@ -853,21 +853,23 @@ The following fields can be written to:
 
 ## `seL4_Word microkit_vcpu_x86_read_msr(microkit_child vcpu, seL4_Word field)`
 
-Read a 64-bit Model Specific Register of a given virtual CPU with ID `vcpu` specified by,
+Read a 64-bit Model Specific Register of a given virtual CPU with ID `vcpu` specified by
 `field` from hardware using the `rdmsr` instruction. The following registers can be written to:
-`IA32_LSTAR_MSR`,
-`IA32_STAR_MSR`,
-`IA32_CSTAR_MSR` and
-`IA32_FMASK_MSR`.
+
+* `IA32_LSTAR_MSR`
+* `IA32_STAR_MSR`
+* `IA32_CSTAR_MSR`
+* `IA32_FMASK_MSR`
 
 ## `void microkit_vcpu_x86_write_msr(microkit_child vcpu, seL4_Word field, seL4_Word value)`
 
-Write a 64-bit Model Specific Register of a given virtual CPU with ID `vcpu` specified by,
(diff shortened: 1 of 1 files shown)


libmicrokit/include/microkit.h
```diff
@@ -394,6 +394,91 @@ static inline seL4_Uint32 microkit_x86_ioport_read_32(microkit_ioport ioport_id,
 }
 #endif
 
+#if defined(CONFIG_ARCH_X86_64) && defined(CONFIG_VTX)
+static inline seL4_Word microkit_vcpu_x86_read_vmcs(microkit_child vcpu, seL4_Word field)
+{
+    seL4_X86_VCPU_ReadVMCS_t ret;
+    ret = seL4_X86_VCPU_ReadVMCS(BASE_VCPU_CAP + vcpu, field);
+    if (ret.error != seL4_NoError) {
+        microkit_dbg_puts("microkit_x86_read_vmcs: error reading data\n");
```
midnightveil, line 403: Does it really make sense to always handle these errors by crashing?

libmicrokit/include/microkit.h
```diff
@@ -394,6 +394,92 @@ static inline seL4_Uint32 microkit_x86_ioport_read_32(microkit_ioport ioport_id,
 }
 #endif
 
+#if defined(CONFIG_ARCH_X86_64) && defined(CONFIG_VTX)
+static inline seL4_Word microkit_vcpu_x86_read_vmcs(microkit_child vcpu, seL4_Word field)
```
midnightveil, line 398: Should probably be added to the manual.

dreamliner787-9: Whoops this slipped my mind will add now.

libmicrokit/include/microkit.h
dreamliner787-9: That's what the ARM stuff does as well so I just followed in the same style.

midnightveil: Sure. Still don't like it, and I don't think we should keep doing it.

midnightveil: :(

Ivan-Velickovic: Forgot to comment before merging. I am merging this as it is because that is the expectation for all the libmicrokit calls right now but if you have a better idea I am open to it @midnightveil.

Crashing makes sense for certain APIs in my opinion because trying to access some resource you don't have access to means something has gone horribly wrong since you would usually access all system resources during initialisation or shortly after initialisation.

Maybe crashing is too harsh on non-debug builds, I do not know.

PR #353, opened 2025-11-11
Add SMP support for ARM/RISC-V

midnightveil:
This PR adds support for SMP builds of the kernel on AArch64.

We have a `psci_target_cpus[]` array, currently hardcoded, for the IDs of the cores to boot. This should match the values specified in the device tree /cpu@0/ node's `<reg>` field. Note that due to quirks of various platforms (e.g. OdroidC4), these do not (but can) match the `MPIDR` values that hardware reports, so we maintain a different table of their MPIDR values.

There a few TODOs scattered throughout this code.

### DIFF
--- a/.github/workflows/sdk.yaml
+++ b/.github/workflows/sdk.yaml
@@ -16,7 +16,7 @@ on:
     branches: [ "main" ]
 
 env:
-  SEL4_VERSION: 0f497ab3a0a25500b7dc70f3a2c2ad34b8453c24
+  SEL4_VERSION: 11c5d5052780a5156c345355567612b35d685a78
 
 # To reduce the load we cancel any older runs of this workflow for the current
 # PR. For deployment to the main branch, the workflow will run on each push,
--- a/DEVELOPER.md
+++ b/DEVELOPER.md
@@ -120,7 +120,7 @@ It should be noted that while Microkit uses mainline seL4, it is fairly tied to 
 specific version used.
 
 For this version of Microkit please use the following commit:
-`0f497ab3a0a25500b7dc70f3a2c2ad34b8453c24`.
+`11c5d5052780a5156c345355567612b35d685a78`.
 
 ## Building the SDK
 
--- a/tool/microkit/src/sel4.rs
+++ b/tool/microkit/src/sel4.rs
@@ -434,7 +434,13 @@ impl ObjectType {
     pub fn fixed_size_bits(self, config: &Config) -> Option<u64> {
         match self {
             ObjectType::Tcb => match config.arch {
-                Arch::Aarch64 => Some(11),
+                Arch::Aarch64 => {
+                    if config.hypervisor && config.benchmark && config.num_cores > 0 {
+                        Some(12)
+                    } else {
+                        Some(11)
+                    }
+                }
                 Arch::Riscv64 => match config.fpu {
                     true => Some(11),
                     false => Some(10),


loader/src/aarch64/crt0.S
```diff
@@ -10,16 +10,6 @@
 .global _start;
 .type _start, %function;
 _start:
-
-     mrs    x0, mpidr_el1
-     and    x0, x0,#0xFF        // Check processor id
```
midnightveil, line 15: I'll add a quick note on why this was removed: this is wrong, as we are not guaranteed that the CPU we are booting on will have the Aff0 value of the MPIDR_EL1 to be 0. Nor do we expect that this part of the code should run on multiple cores. This dates back the initial public release of seL4CP.

build_sdk.py
```diff
@@ -156,6 +156,16 @@ class ConfigInfo:
...
         } | DEFAULT_KERNEL_OPTIONS_AARCH64,
     ),
+    BoardInfo(
+        name="maaxboard_4_cores",
+        arch=KernelArch.AARCH64,
+        gcc_cpu="cortex-a53",
+        loader_link_address=0x50000000,
+        kernel_options={
+            "KernelPlatform": "maaxboard",
+            "KernelMaxNumNodes": 4,
```
Indanz: I don't think adding new boards is the way to define SMP configurations of existing boards.

Okay for a stop-gap solution, but long-term there must be a better way. How do you deal with this with e.g. debug versus release builds, or other config options like that?

loader/src/aarch64/cpus.c
```diff
@@ -0,0 +1,250 @@
...
+#elif defined(CONFIG_PLAT_QEMU_ARM_VIRT)
+/* QEMU is special and can have arbitrary numbers of cores */
+// TODO.
+size_t psci_target_cpus[4] = {0x00, 0x01, 0x02, 0x03};
+#else
+#error "psci_target_cpus not defined for this board"
+#endif
+
+_Static_assert(sizeof(psci_target_cpus) / sizeof(psci_target_cpus[0]) >= NUM_ACTIVE_CPUS,
+               "active CPUs cannot be more than available CPUs");
```
Indanz:
```suggestion
_Static_assert(NUM_ACTIVE_CPUS <= sizeof(psci_target_cpus) / sizeof(psci_target_cpus[0]),
               "active CPUs cannot be more than available CPUs");
```
and ARRAY_SIZE if you have it.

loader/src/aarch64/cpus.c
```diff
@@ -0,0 +1,250 @@
...
+ *
+ *      For PSCI functions that use only 32-bit parameters, the arguments are
+ *      passed in R0 to R3 (AArch32) or W0 to W3 (AArch64), with return values
+ *      in R0 or W0. For versions using 64-bit parameters, the arguments are
+ *      passed in X0 to X3, with return values in X0. In line with the SMC
+ *      Calling Conventions, the immediate value used with an SMC (or HVC)
+ *      instruction must be 0.
+ *
+ * Hence we need both the SMC64 and SMC32 conventions implemented, with support
+ * for up to 3 arguments.
```
Indanz: You are allowed to do the calls following the 32-bit calling convention from 64-bit mode, I believe. Whether that will work in practice is another matter. Maybe some implementations will assume the caller's mode is equal to the API used, which won't work for CPU_ON.

loader/src/aarch64/cpus.c
```diff
@@ -0,0 +1,250 @@
...
+    register uint64_t x2 asm("x2") = arg1;
+    register uint64_t x3 asm("x3") = arg2;
+    asm volatile(
+        "smc #0\n"
+        : "=r"(x0)
+        : "r"(x0), "r"(x1), "r"(x2), "r"(x3)
+        : "x4", "x5", "x6", "x7", "x8", "x9", "x10", "x11", "x12", "x13", "x14", "x15", "x16", "x17",
+        "memory"
+    );
+
```
Indanz: Why all these empty lines everywhere?

loader/src/aarch64/cpus.c
```diff
@@ -0,0 +1,250 @@
...
+ * specifically §2.7 "SMC64/HVC64 argument passing".
+ *
+ * We only support up to 4 arguments, but the actual convention supports up
+ * to 17, and clobbers X4--X17. The convention also supports up to 17 returns,
+ * but we again only support 1.
+ *
+ **/
+uint64_t arm_smc64_call(uint32_t function_id, uint64_t arg0, uint64_t arg1, uint64_t arg2)
+{
+    // Per Table 2-1, it should be 1 for SMC64.
```
Indanz: Better to say directly that bit 30 determines the calling convention. Or drop the comment.

loader/src/aarch64/cpus.c
```diff
@@ -0,0 +1,250 @@
...
+ **/
+uint32_t arm_smc32_call(uint32_t function_id, uint32_t arg0, uint32_t arg1, uint32_t arg2)
+{
+    // Per Table 2-1, it should be 0 for SMC32.
+    if ((function_id & BIT(30)) == BIT(30)) {
+        puts("LDR|ERROR: SMC64 function used in SMC32 call\n");
+        return PSCI_RETURN_INVALID_PARAMETERS;
+    }
+    // TODO: This only supports AArch64 mode for the moment, in AArch32 mode
+    //       we need to use the x... registers.
```
Indanz: Why add this function at all then?

loader/src/aarch64/cpus.c
```diff
@@ -0,0 +1,250 @@
...
+        : "r"(w0), "r"(w1), "r"(w2), "r"(w3)
+        : "w4", "w5", "w6", "w7", "w8", "w9", "w10", "w11", "w12", "w13", "w14", "w15", "w16", "w17",
+        "memory"
+    );
+
+    return w0;
+}
+
+/** defined in util64.S */
+extern char arm_secondary_cpu_entry_asm[1];
```
Indanz:
```suggestion
void arm_secondary_cpu_entry_asm(void* sp);
```
and use `&arm_secondary_cpu_entry_asm` elsewhere. (May need to rename the function to `_arm_secondary_cpu_entry_asm` in ASM for this to work).

However, I thought CPU_ON gave an additional parameter to the called function. Good to document it here, even if you're not using it.

loader/src/aarch64/cpus.c
```diff
@@ -0,0 +1,250 @@
...
+fail:
+    for (;;) {}
+}
+
+int plat_start_cpu(int logical_cpu)
+{
+    LDR_PRINT("INFO", 0, "Starting CPU ");
+    puts((const char[]) {
+        '0' + logical_cpu, '\0'
+    });
```
Indanz:
```suggestion
    putchar('0' + logical_cpu);
```
Or combine it with the next puts into:
```
    char id = "0\n";
    id[0] += logical_cpu;
    puts(id);
```

loader/src/aarch64/cpus.c
```diff
@@ -0,0 +1,250 @@
...
+    uint64_t sp = stack_top;
+    /* write the logical id to the stack, and a dummy (since we don't use it) */
+    asm volatile(
+        /* use the pre-index variant of STP */
+        "stp %[v0], %[v1], [%[sp], #-16]!"
+        : [sp] "+r"(sp)
+        : [v0] "r"(logical_cpu),
+        [v1] "r"(/* dummy */ 0)
+        : "memory"
+    );
```
Indanz:
```suggestion
    /* on aarch64 stack grows down */
    uint64_t* stack = _stack[logical_cpu + 1]; // And make _stack an array of uint64_t to avoid casting
    uint64_t sp = (uint64_t)stack;
   
    /* write the logical id to the stack, and a dummy (since we don't use it) */
    stack[-1] = logical_cpu;
    stack[-2] = 0;
```

loader/src/aarch64/init.c
```diff
@@ -64,9 +64,34 @@ static void configure_gicv2(void)
...
+     * note the issues where it forces us to flush any shared addresses all the
+     * way to cache as we might have mixed non-cached/cached access.
+     */
+    puts("LDR|INFO: disabling MMU (if it was enabled)\n");
+    enum el el = current_el();
+
+    if (el == EL1) {
+        el1_mmu_disable();
+    } else if (el == EL2) {
+        el2_mmu_disable();
```
Indanz, line 95: Hm, not sure I like this. Did you ran into issues when you didn't do this?

Isn't it better to setup the loader's own page tables before potentially overwriting U-Boot's page tables instead? Then you have a known state and can have caching enabled when you do all the memory copies and other relatively slow operations.

loader/src/aarch64/util64.S
```diff
@@ -391,6 +419,25 @@ BEGIN_FUNC(el2_mmu_enable)
...
+ * We choose to pass the appropriate stack value in x0.
+ * We have put on this stack:
+ *  - the logical core ID
+ **/
+BEGIN_FUNC(arm_secondary_cpu_entry_asm)
+    mov sp, x0
+    /* retrieve values from the stack (x0=logical id, x1=dummy) */
+    ldp x0, x1, [sp], #16
+    /* which we then overwrite */
+    mrs x1, mpidr_el1
```
Indanz: Make up your mind what you want to do, this is confusing. Either pass the info and use it, or don't. But please don't pass it and then overwrite it in assembly, that's sneaky.

loader/src/cutil.h
```diff
@@ -4,12 +4,14 @@
...
 
-#include <stddef.h>
+#pragma once
 
+#include <stddef.h>
 
 #define ALIGN(n)  __attribute__((__aligned__(n)))
 
-#define MASK(x) ((1UL << x) - 1)
+#define BIT(n)  (1UL << n)
```
Indanz:
```suggestion
#define BIT(n)  (1UL << (n))
```

loader/src/cutil.h
```diff
@@ -4,12 +4,14 @@
...
+#define MASK(x) (BIT(x) - 1)
```
Indanz:
```suggestion
#define MASK(x) (BIT(x) - 1)
```
```suggestion
#define MASK(x) (BIT(x) - 1UL)
```

loader/src/loader.c
```diff
@@ -42,7 +41,7 @@ extern char _text;
 extern char _bss_end;
 const struct loader_data *loader_data = (void *) &_bss_end;
 
-char _stack[STACK_SIZE] ALIGN(16);
+char _stack[NUM_ACTIVE_CPUS][STACK_SIZE] ALIGN(16);
```
Indanz:
```suggestion
unint64_t _stack[NUM_ACTIVE_CPUS][STACK_SIZE / sizeof(unint64_t)] ALIGN(16);
```
But whatever you prefer.

loader/src/loader.c
```diff
@@ -106,8 +105,27 @@ static void copy_data(void)
     }
 }
 
-static void start_kernel(void)
+#ifdef CONFIG_PRINTING
+static int cpu_up = 0;
```
Indanz: I'd make this an _Atomic and use simple assignments instead. And if it's a lock, perhaps give it another name?

loader/src/loader.c
```diff
@@ -162,16 +183,29 @@ int main(void)
...
+            puthex32(r);
+            goto fail;
+        }
+
+    #ifdef CONFIG_PRINTING
+        /* wait for boot */
+        while(__atomic_load_n(&cpu_up, __ATOMIC_ACQUIRE) != 1);
+        /* allow the next CPU to boot */
+        __atomic_store_n(&cpu_up, 0, __ATOMIC_RELEASE);
+    #endif
```
Indanz: I don't understand this bit.

loader/src/loader.h
```diff
@@ -32,3 +36,9 @@ struct loader_data {
 };
 
 extern const struct loader_data *loader_data;
+
+#define STACK_SIZE 4096
```
Indanz: The assembly hard-codes `0xff0`, maybe it should use the define instead?

loader/src/aarch64/init.c
midnightveil: Yes, without this it doesn't work on real hardware. Presumably because then core 0 is running with the mmu on and we need to do cache flushing for other cores to see data writes. That is however, probably a better solution. (The PSCI implementation does not seem to flush the cached of this sort for you)

loader/src/aarch64/util64.S
midnightveil: It's because aarch64 requires a double word aligned stack. I have to pass the dummy value. there's nothing I can do about that.

loader/src/aarch64/cpus.c
midnightveil: This then misses decrementing the stack.

Also, I initially did this, but I didn't like the extra decrements you have to do separately to writing to the stack. Using the inline assembly gives the precide behaviour we want without needing to think about it.

loader/src/aarch64/cpus.c
midnightveil: What we really should do is add a very simple putdecimal() function.

putchar() doesn't work because currently it's not exposed from the uart implementation.

loader/src/aarch64/cpus.c
midnightveil: I think I added it so that I can call the psci version function but then forgot. Oops.

build_sdk.py
midnightveil: Yep, it's horrible.

Debug and release builds are subfolders in the SDK; so it is /sdk/boards/maaxboard/debug/bin/sel4.elf for example.

One proposal I think was to turn on SMP by default. I personally don't like this.

Moving the folder layout around isn't great because C's lack of a build system means that you have to hard code the include paths in your project; moving these around would then break most projects.

loader/src/aarch64/cpus.c
midnightveil: The additional parameter is the "context id". In which I'm directly passing the stack and data on the stack, as per your recommendation previously. There's no actual $sp argument.

loader/src/aarch64/util64.S
Indanz: The number of arguments passed doesn't affect the stack alignment, the passed stack pointer must be aligned, but that's independent.

loader/src/aarch64/cpus.c
Indanz: Ah, yes, forgot about that. And totally missed that in the assembly. I think passing `stack - 2` would be clearer and more explicit, but if you don't like the negative indexing, you can also first set stack to the final value and then use positive indexing for the arguments.

build_sdk.py
Indanz: I don't see the problem. What is the problem? You can add extra include paths via `-I` if you want to change a set of includes, but I don't see how that's relevant to a python build script.

As far as build systems go, seL4 uses CMake already and you have your python scripts and Rust tool already, how would another language based build system solve any of your problems here?

midnightveil: Restructuring the SDK folder structure is the public "API" of the SDK. Moving the folders around to also add an SMP subfolder would likely break all existing consumers of the SDK.

To be clear, I'm not saying it's a good solution, just the least bad one we could think of

loader/src/aarch64/cpus.c
Indanz: I thought CPU_ON also passed another argument, but looking at the spec there's just the context id and nothing else.

loader/src/aarch64/util64.S
midnightveil: OK, but we need to read off the stack?

Unless you're saying I should choose to explicit disregard the aarch64 stack calling convention?

build_sdk.py
Indanz: So you don't have a solution for multiple config options, except for debug. You better think of something, as this is a common problem. What about:

/sdk/boards/maaxboard/smp-debug/
/sdk/boards/maaxboard/smp-release/
/sdk/boards/maaxboard/smp-otheroption-debug/

etc.

midnightveil: That's almost identical to the maaxboard_smp solution IMO.

It might make a bit more sense to make them subdirectories of the boards, but now you have like several config variants and it's the same problem as new board names?

(Actually saying that now I think it would be a better solution than the board name)

p.s.: the current solution where we hardcode add new boards to the boards list is obviously terrible and could be easily improved to be generic; it's just the easiest way to do it when I don't know the perfect answer.

loader/src/loader.h
midnightveil: Oh joy. I just moved it from the loader.c to loader.h, it was like this before.

loader/src/loader.c
midnightveil: What about it?

It's purely just so that prints aren't interleaved, and it's spin-waiting for the other core to store 1 to it, then we set back to 0 for the next iteration of the loop.

loader/src/aarch64/cpus.c
midnightveil: I like the return on a separate line from this larger block of code. It makes it a bit easier to see what's going on IMO.

loader/src/riscv/cpus.c
```diff
@@ -0,0 +1,101 @@
...
+#elif defined(CONFIG_PLAT_HIFIVE_P550)
+int logical_to_hart_id[4] = {0x0, 0x1, 0x2, 0x3};
+#else
+#error "Unsupported platform TODO"
+#endif
+
+int hart_id_to_logical(int hart_id);
+void riscv_secondary_cpu_entry(int hart_id);
+
+int hart_id_to_logical(int hart_id)
```
Ivan-Velickovic: TODO: remove this and pass the logical CPU ID on the stack.

loader/src/riscv/crt0.S
```diff
@@ -103,7 +104,9 @@ hsm_switch_hart:
    * secondary hart
    */
   mv a0, s0 /* restore a0 to hold hart ID passed by OpenSBI */
-  j spin_hart /* Spin any hart that isn't FIRST_HART_ID since we only support single-core right now. */
+  li a7, SBI_HSM_BASE_EID
+  li a6, 0x1
+  ecall
```
Ivan-Velickovic, line 103: Need to clean this up and explain what's going on.

loader/src/riscv/crt0.S
```diff
@@ -17,6 +17,7 @@
 
 #define STACK_SIZE 4096
 
+// TODO: defined in sbi.h instead
```
Ivan-Velickovic: Fix TODO

build_sdk.py
midnightveil: I've settled on your proposal (and discussed with Ivan in person). These config options are better than various board configs, but we're not entirely happy with them either.

midnightveil:
This is blocked on https://github.com/seL4/seL4/pull/1563

loader/src/aarch64/cpus.c
midnightveil: Specifically: when I say AArch64 mode here, I mean that it can only work if AArch64 calls SMC32.

loader/src/riscv/cpus.c
```diff
@@ -0,0 +1,129 @@
...
+#include <stdint.h>
+
+#include "../cpus.h"
+#include "../cutil.h"
+#include "../loader.h"
+#include "../uart.h"
+#include "sbi.h"
+
+/**
+ * TODO(ivanv): explain how hart IDs are used for RISC-V
```
midnightveil: @Ivan-Velickovic

loader/src/riscv/cpus.c
```diff
@@ -0,0 +1,129 @@
...
+       to have space for the arguments to the entrypoint */
+    uint64_t *sp = (uint64_t *)((uintptr_t)stack_base + STACK_SIZE - 2 * sizeof(uint64_t));
+    /* store the logical cpu on the stack */
+    sp[0] = logical_cpu;
+    /* zero out what was here before */
+    sp[1] = 0;
+
+    uint64_t hart_id = plat_get_hw_id(logical_cpu);
+
+    // struct sbi_ret ret = sbi_call(SBI_EXT_HSM, SBI_HSM_HART_STOP, hart_id, 0, 0, 0, 0, 0);
```
midnightveil: @Ivan-Velickovic

Ivan-Velickovic: Experimental changes, removed.

loader/src/riscv/cpus.c
Ivan-Velickovic: Done.

loader/src/riscv/crt0.S
Ivan-Velickovic: Fixed.

loader/src/aarch64/cpus.c
```diff
@@ -0,0 +1,134 @@
...
+#include <stdint.h>
+
+#include "smc.h"
+#include "../cpus.h"
+#include "../cutil.h"
+#include "../loader.h"
+#include "../uart.h"
+
+uint64_t arm_smc64_call(uint32_t function_id, uint64_t arg0, uint64_t arg1, uint64_t arg2);
+uint32_t arm_smc32_call(uint32_t function_id, uint32_t arg0, uint32_t arg1, uint32_t arg2);
```
midnightveil: oops I should have removed these prototypes as they are now in SMC.h

(should that be an object file as well like SBI? idk)

Ivan-Velickovic: I put SBI in a separate file just because UART uses it as well.

loader/src/aarch64/smc.h
```diff
@@ -0,0 +1,133 @@
...
+#define PSCI_RETURN_INVALID_PARAMETERS -2
+#define PSCI_RETURN_DENIED -3
+#define PSCI_RETURN_ALREADY_ON -4
+#define PSCI_RETURN_ON_PENDING -5
+#define PSCI_RETURN_INTERNAL_FAILURE -6
+#define PSCI_RETURN_NOT_PRESENT -7
+#define PSCI_RETURN_DISABLED -8
+#define PSCI_RETURN_INVALID_ADDRESS -9
+
+static inline const char *psci_return_as_string(uint32_t ret)
```
midnightveil, line 45: I should add a note on why this is correct for 64-bit: 64-bit to 32-bit casts will truncate the upper bits, which for errors will be 0xFFFFFFFF_XXXXXXXX and the X's will be identical between both.

loader/src/aarch64/cpus.c
Ivan-Velickovic: @midnightveil I've removed the declarations.

loader/src/aarch64/util64.S
```diff
@@ -391,6 +419,25 @@ BEGIN_FUNC(el2_mmu_enable)
...
+ * initially be given x0, our context_id as a parameter.
+ * We choose to pass the appropriate stack value in x0.
+ * We have these arguments on the stack:
+ *  - the logical core ID
+ **/
+BEGIN_FUNC(arm_secondary_cpu_entry_asm)
+    mov sp, x0
+    /* grab the core ID off the stack */
+    ldr x0, [sp]
+    /* which we then overwrite */
```
midnightveil: this comment I forgot to remove

loader/src/cpus.h
```diff
@@ -0,0 +1,50 @@
...
+/**
+ * Get the hardware ID from a logical ID from platform specific code
+ **/
+size_t plat_get_hw_id(int logical_cpu);
+
+/**
+ * Start the CPU with the given logical ID.
+ * Returns a non-zero integer on failure.
+ **/
+int plat_start_cpu(int logical_cpu);
```
midnightveil, line 51: nit for me: should rename this, most other int logical_cpu arguments fns use that as the current CPU, whereas this function assumes CPU0 and starts that logical cpu.

loader/src/loader.c
```diff
@@ -58,13 +48,13 @@ static void print_flags(void)
 
 static void print_loader_data(void)
 {
-    puts("LDR|INFO: Flags:\n");
+    puts("LDR|INFO: flags:\n");
     print_flags();
-    puts("LDR|INFO: Kernel:      entry:   ");
+    puts("LDR|INFO: fernel:      entry:   ");
```
midnightveil: kernel not fernel

loader/src/loader.c
```diff
@@ -158,16 +163,29 @@ int main(void)
...
+    puthex32(plat_get_active_cpus());
+    puts("\n");
 
-    puts("LDR|INFO: jumping to kernel\n");
-    start_kernel();
+    for (int cpu = 1; cpu < plat_get_active_cpus(); cpu++) {
+        r = plat_start_cpu(cpu);
+        if (r != 0) {
+            LDR_PRINT("ERROR", 0, "unable to start CPU ");
+            puthex32(cpu);
```
midnightveil: put decimal now

midnightveil:
Previous before force push cleanups: https://github.com/au-ts/microkit/tree/julia/smp-presquash

<details><summary>diff</summary>

```
diff --git a/flake.nix b/flake.nix
index 5a3f926..65a24df 100644
--- a/flake.nix
+++ b/flake.nix
@@ -96,6 +96,7 @@
             ninja
             libxml2
             qemu
+            gnutar
           ];

           # Necessary for Rust bindgen
diff --git a/loader/src/aarch64/cpus.c b/loader/src/aarch64/cpus.c
index dc76a1f..7ba31e2 100644
--- a/loader/src/aarch64/cpus.c
+++ b/loader/src/aarch64/cpus.c
@@ -24,7 +24,7 @@ size_t cpu_mpidrs[NUM_ACTIVE_CPUS];

 void plat_save_hw_id(int logical_cpu, size_t hw_id)
 {
-    cpu_mpidrs[(uint8_t)logical_cpu] = hw_id;
+    cpu_mpidrs[logical_cpu] = hw_id;
 }

 uint64_t plat_get_hw_id(int logical_cpu)
@@ -73,6 +73,9 @@ void arm_secondary_cpu_entry(int logical_cpu, uint64_t mpidr_el1)
     } else if (logical_cpu >= NUM_ACTIVE_CPUS) {
         LDR_PRINT("ERROR", logical_cpu, "secondary CPU should not be >NUM_ACTIVE_CPUS\n");
         goto fail;
+    } else if (logical_cpu < 0) {
+        LDR_PRINT("ERROR", logical_cpu, "secondary CPU should not have negative logical id\n");
+        goto fail;
     }

     plat_save_hw_id(logical_cpu, mpidr_el1);
```

</details>

PR #356, opened 2025-11-13
Add support for Radxa Rock3b board

JDuchniewicz:


midnightveil:
This needs to be redone on the loader changes. It should appear in uart.c/uart.h instead of loader.c.

loader/src/uart.c
```diff
@@ -169,7 +169,19 @@ void putc(uint8_t ch)
     while ((*UART_REG(ULSR) & ULSR_THRE) == 0);
     *UART_REG(UTHR) = ch;
 }
+#elif defined(BOARD_rock3b)
+#define UART_BASE   0xfe660000
+#define UTHR        0x0
+#define ULSR        0x14
+#define ULSR_THRE   (1 << 5)
+
+void uart_init() {}
```
midnightveil: This will break with #357, do you mind adding `void uart_init(void) {}`?

JDuchniewicz: Fixed

loader/src/uart.c
```diff
@@ -169,7 +169,19 @@ void putc(uint8_t ch)
     while ((*UART_REG(ULSR) & ULSR_THRE) == 0);
     *UART_REG(UTHR) = ch;
 }
+#elif defined(BOARD_rock3b)
```
midnightveil: You missed the other change here, there's no more `BOARD_` defines, it uses the seL4 defines.

Which, now I think of it, requires that `rk3568` is in seL4, which I don't think it is yet? I don't see how this would build.

JDuchniewicz: I am struggling to understand what the order of contribution should be - probably seL4 would be the first stepping stone then. But then I think contributions there might take much longer to be merged to we remain in the weird state where the changes in this PR work neither with current code nor with what will be later available in seL4 kernel.

midnightveil: This doesn't work now? The board defines don't exist anymore.

And I'm not sure what you mean, everything depends on seL4 so I'm not sure how else it is supposed to work.

JDuchniewicz: Fixed

midnightveil (approved):

lsf37:
seL4 support is now merged, so this one can be merged whenever things are ready.

build_sdk.py
```diff
@@ -317,6 +317,15 @@ class ConfigInfo:
             "KernelPlatform": "rockpro64",
         } | DEFAULT_KERNEL_OPTIONS_AARCH64,
     ),
+    BoardInfo(
+        name="rock3b",
+        arch=KernelArch.AARCH64,
+        gcc_cpu="cortex-a55",
+        loader_link_address=0x30000000,
```
Ivan-Velickovic, line 324: The loader link address is `0x30000000` but the documentation has `0x02000000`?

It should be fine since we added support for the loader to relocate itself but I think it's probably best for at least the docs instructions to be consistent.

JDuchniewicz: Thanks for that, fixed up the documentation much later than this and did not come back to patching that up

PR #359, opened 2025-11-17
Switch to no-heap capDL initialiser

dreamliner787-9:
This PR update the capDL implementation in Microkit to use the new no-heap capDL initialiser that was merged into rust-seL4 as part of https://github.com/seL4/rust-sel4/pull/297.

System boot-up time and memory consumption has been significantly improved as we are now using the embed frames feature and the initialiser no longer need to deserialise the spec on boot.

rust-seL4 is now pinned to the `main` branch.

### DIFF
--- a/tool/microkit/src/capdl/builder.rs
+++ b/tool/microkit/src/capdl/builder.rs
@@ -96,7 +96,7 @@ const PD_BASE_VCPU_CAP: u64 = PD_BASE_VM_TCB_CAP + 64;
 const PD_BASE_IOPORT_CAP: u64 = PD_BASE_VCPU_CAP + 64;
 
 pub const PD_CAP_SIZE: u32 = 512;
-const PD_CAP_BITS: u64 = PD_CAP_SIZE.ilog2() as u64;
+const PD_CAP_BITS: u8 = PD_CAP_SIZE.ilog2() as u8;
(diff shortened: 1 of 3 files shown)


build_sdk.py
```diff
@@ -768,7 +768,7 @@ def build_initialiser(
             cargo install {cargo_cross_options} \
             --target {cargo_target} \
             --locked \
-            --git https://github.com/au-ts/rust-seL4 --branch capdl_dev sel4-capdl-initializer --rev 0186e53991b48118b39d9a048c526baabdf6f5d4 \
+            --git https://github.com/au-ts/rust-seL4 --branch capdl_dev_no_heap_rebased sel4-capdl-initializer --rev 5cd50c6e94745700ec8438277ed87db532d88feb \
```
Ivan-Velickovic:
```suggestion
            --git https://github.com/au-ts/rust-seL4 sel4-capdl-initializer --rev 5cd50c6e94745700ec8438277ed87db532d88feb \
```

Ivan-Velickovic: I think `--branch` is redundant now that we do `--rev`.

midnightveil: Git still needs to know which branch to look at to be able to find a rev if it's not in main, otherwise it's not reachable usually. (iirc)

dreamliner787-9: I've removed --branch and it seems to work.

PR #367, opened 2025-11-18
(Re-)allow PD priority of 254

Ivan-Velickovic:
This partially reverts 9d6b4e09663d4192b00d1c219d28d2b299659d55 where the maximum possible priority allowed for a PD was decreased from 254 to 253.

This would break many, many existing Microkit systems and even though it is a trivial fix, I do not think it is worth it for users.

Previously, the monitor was the initial task and so it had the maximum seL4 TCB priority by default, 255.

Because the monitor is a fault handler and it has some other responsibilities such as converting PDs to passive PDs during init, it makes sense for it to be always the highest priority thread in the system. Hence, the maximum priority of any PD was 254, one less than the monitor's priority.

Now with the capDL transition, we have another run-time component involved, the capDL initialiser. It is the initial task now and so it gets the default priority of 255. If the monitor also had a priority of 255 it would mean that some debug printing with the initialiser would get interleaved/messed up which looks unprofessional, so we made the monitor have priority 254 and then all PDs got 253 or lower.

But, that only happens at the very end of the capDL initialiser doing its job so it's probably better to just remove that particular print and instead have the monitor be 255 and so Microkit users don't have to deal with any breaking changes. This print was removed in [1].

[1]: https://github.com/seL4/rust-sel4/pull/310/commits/571093a4aac29210ac3d41ba43ddaa16325f6368.

### DIFF
--- a/docs/manual.md
+++ b/docs/manual.md
@@ -168,7 +168,7 @@ The overall computational model for a Microkit system is a set of isolated compo
 
 The PD has a number of scheduling attributes that are configured in the system description:
 
-* priority (0 -- 253)
+* priority (0 -- 254)
 * period (microseconds)
 * budget (microseconds)
 * passive (boolean)
--- a/example/ethernet/ethernet.system
+++ b/example/ethernet/ethernet.system
@@ -44,7 +44,7 @@
 
     <memory_region name="eth_clk" size="0x1_000" phys_addr="0x5b200000" />
 
-    <protection_domain name="gpt" priority="253">
+    <protection_domain name="gpt" priority="254">
         <program_image path="gpt.elf" />
         <map mr="lsio_gpt0" vaddr="0x2_000_000" perms="rw" cached="false" setvar_vaddr="gpt_regs" />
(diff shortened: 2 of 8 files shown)


lsf37:
The initialiser correctness does depend on the initialiser being the highest priority thread in the system. If you can make sure that monitor is enabled/resumed as the last thread and the initialiser doesn't have to do anything afterwards any more, the correctness argument should still work, though.

Ivan-Velickovic:
> The initialiser correctness does depend on the initialiser being the highest priority thread in the system

Does the initialiser depend on being the highest priority thread in the system or just that it will be scheduled enough such that it can finish initialising the system?

lsf37:
In a microkit system that is probably benign in the sense that almost nothing has enough authority to change cap state, but I guess the monitor would be exactly the component that can. I don't think in this case it would actually violate correctness, but the correctness argument would still no longer apply.

Not sure how the overall initialisation phase looks like. If the monitor orchestrates it after being loaded, then increasing its own priority as the first step before it allows other components to run would satisfy everything. If it doesn't orchestrate initialisation then we'd still have a race between the priority increase and the initialisation of other high priority threads in the system.

Issue #249, opened 2024-11-07
add option to generate an ELF file together with the bootable image


wucke13:
The ELF should contain the sections with information on target load addresses so that the ELF file and the entry point (all standard metadata supported by ELF AFAIK) can be directly loaded with a debugger (i.e. GDB's `load`). This is primarily easing work with debuggers and tracing hardware.

The Binary itself can of course be trivially extracted from the respective ELF, but no the other way round.

Ivan-Velickovic:
I'm confused because can't you already load the ELFs into GDB? E.g `add-symbol-file myprogram.elf`.

I'm not sure how a single ELF would work, since it would have multiple ELFs nested inside of it, how does GDB know which one to debug?

Ivan-Velickovic:
Do you mean a single ELF for the whole system or are you looking to debug a particular component such as the loader, initial task or kernel?

wucke13:
> I'm confused because can't you already load the ELFs into GDB

Yes, that works. However, for a larger deployment with dozens if not hundreds of components each bringing their own ELF this might not scale, especially when verifying integration or debugging problems going across.

> I'm not sure how a single ELF would work, since it would have multiple ELFs nested inside of it, how does GDB know which one to debug?

I think one would merge the ELF files into one, which has different sections per protection domain each with code and debug symbols. I'm unsure about the specifics, tagging @moritzmeier for further input.

> Do you mean a single ELF for the whole system

This. The individual ELF files all can be found already scattered across the local builddir and the `MICROKIT_SDK` path, right?

wucke13:
> Sure, although I'm still not quite sure how that would end up working. I don't know how GDB would be able to interpret the ELF as actually being multiple ELFs, each in a separate address space.

Why does it need to interpret as multiple ELFs? In the end, there is multiple code sections, each loaded to different addresses in physical memory, and each code section also is accompanied by individual DWARF. The fact that this could also be represented by multiple, individual ELF files should be irrelevant? I'm aware of a (unfortunately closed source, hence I can not share) example of an OS doing this, they use a small python script to generate a linker script that just links together the different binaries, each as an individual section, and each with the target load address in the metadata.

With that ELF, for example in GDB or in Xilinx' `xsct` you can do a single load, without specifying addresses, and all is set (and the debugger has the source code).

moritz-meier:
[...]

PR #370, opened 2025-11-19
tool: add an option to output ELF image 

dreamliner787-9:
Added an option to the Microkit tool to output the image as an ELF on ARM and RISC-V platforms.

Closes https://github.com/seL4/microkit/issues/249


tool/microkit/src/main.rs
```diff
@@ -55,12 +55,45 @@ fn print_help(available_boards: &[String]) {
...
+
+impl ImageOutputType {
+    fn default(arch: &Arch) -> Self {
+        match arch {
+            Arch::Aarch64 | Arch::Riscv64 => ImageOutputType::Binary,
+            Arch::X86_64 => ImageOutputType::Elf,
+        }
+    }
+
+    fn from(str: &str) -> Option<Self> {
```
midnightveil: don't name this from if you implement the TryFrom semantics;

(this doesn't actually seem to implement the From trait at all, so if you don't want that you might want to name it something else that doesn't conflict for easy of understanding)

tool/microkit/src/main.rs
```diff
@@ -55,12 +55,45 @@ fn print_help(available_boards: &[String]) {
...
     println!("  --search-path [SEARCH_PATH ...]");
 }
 
+enum ImageOutputType {
+    Binary,
+    Elf,
+}
+
+impl ImageOutputType {
+    fn default(arch: &Arch) -> Self {
```
midnightveil: same here, if you're not implementing the Default::default trait don't name it default otherwise rustaceans are going to get confused.

tool/microkit/src/main.rs
```diff
@@ -373,6 +418,27 @@ fn main() -> Result<(), String> {
         _ => panic!("Unsupported kernel config architecture"),
     };
 
+    let image_output_type = if let Some(image_type) = args.output_image_type {
+        match ImageOutputType::from(image_type) {
+            Some(output_image_type) => {
+                if !output_image_type.valid(arch) {
```
midnightveil: "parse don't validate" please?

this links with the try_from comment above... if you return a result you can then print this error

would cleanup half of these if statement

PR #374, opened 2025-11-20
Move to master seL4

Ivan-Velickovic:
Use commit 0f497ab3a0a25500b7dc70f3a2c2ad34b8453c24 instead which
at the time of writing is the latest HEAD.

The 'microkit' branch on the seL4 repository will remain (maybe
we will make it a tag instead) to ensure that previous versions
of the SDK still reliable build.

### DIFF
--- a/DEVELOPER.md
+++ b/DEVELOPER.md
@@ -25,6 +25,7 @@ This section attempts to list the packages or external development tools which a
 * qemu-system-x86_64
 
 To build the documentation you also need
+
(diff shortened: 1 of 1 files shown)


Ivan-Velickovic:
For context, this could have been done since https://github.com/seL4/microkit/pull/337 but I didn't want to change too much at the same time because https://github.com/seL4/microkit/pull/337 was already such a large internal change.

PR #375, opened 2025-11-21
monitor: fix error overwrite scenario

dreamliner787-9:
Fixed a bug in the Microkit Monitor where making a PD passive required two consecutive seL4 system calls, and any error from the first call was never reported.


Ivan-Velickovic:
Just to be clear, there is no error currently and this should never happen but of course we should be checking the results of these invocations.

PR #376, opened 2025-11-21
Copy kernel from SDK to image output directory for x86 platform

dreamliner787-9:
Since Multiboot wants the kernel and initial ram disk (initial task in Microkit's case) as separate images. Having the kernel be in the output directory alongside the initial task output image is more convenient to users.

In addition, we also provide seL4 as a 32-bit ELF since qemu-system-x86_64's -kernel option only accept a 32-bit elf32-i386 kernel image.

### DIFF
--- a/docs/manual.md
+++ b/docs/manual.md
@@ -354,15 +354,15 @@ Additionally, for each supported board configuration the following are provided:
 
 * `libmicrokit.a`
 * `initialiser.elf`
-* `sel4.elf`
 * `monitor.elf`
+* seL4 kernel image
+     * On ARM/RISC-V: `sel4.elf`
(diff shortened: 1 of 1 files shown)


midnightveil:
For "convenience" isn't helpful as justification. Is this because of multiboot wants a kernel and initrd separately? And so having it in the output directory is then useful? OR is it something else?

dreamliner787-9:
> For "convenience" isn't helpful as justification. Is this because of multiboot wants a kernel and initrd separately? And so having it in the output directory is then useful? OR is it something else?

Because Multiboot wants them separately. You are right I will re-word it.

Ivan-Velickovic:
Still more to do for the x86-64 board support package section but it's getting closer.

Issue #313, opened 2025-07-10
Option to spit out `uImage`


Ivan-Velickovic:
For the HiFive P550 and potentially future RISC-V boards the binary image can have some issues during booting and it looks like uImage might be the way to go. See discussion in https://github.com/seL4/seL4/pull/1397 for details.

This should be trivial to do, I assume it's just some uImage header that needs to be added to the binary image.

PR #381, opened 2025-11-24
Output uImage for RISC-V by default instead of binary format

dreamliner787-9:
- Added ability for the Microkit tool to output a uImage with Linux header for RISC-V platforms.
- Output a uImage for RISC-V platforms by default instead of the binary format. So that U-Boot follows the Linux boot protocol and gives us the HART ID in `a0` instead of leaving it undefined.

Closes https://github.com/seL4/microkit/issues/313

### DIFF
--- a/docs/manual.md
+++ b/docs/manual.md
@@ -423,7 +423,7 @@ The format of the system description is described in a subsequent chapter.
 
 Usage:
 
-    microkit [-h] [-o OUTPUT] [-r REPORT] [--capdl-json CAPDL_SPEC] [--image-type {binary,elf}]
+    microkit [-h] [-o OUTPUT] [-r REPORT] [--capdl-json CAPDL_SPEC] [--image-type {binary,elf,uimage}]
               --board [BOARD] --config CONFIG [--search-path [SEARCH_PATH ...]] system
 
 The path to the system description file, board to build the system for, and configuration to build for must be provided.
@@ -444,10 +444,12 @@ It is not intended to be machine readable.
 
 ## Image format
 
-### ARM and RISC-V
+### ARM
 
-On ARM and RISC-V, the image produced by the Microkit tool is typically a raw binary that is intended
-to be loaded and jumped. There is a specified entry point of the image that is documented in the board-specific
+On ARM, the Microkit tool will produce either a raw binary that is intended to be loaded and jumped or
+an ELF. By default, the tool will produce a raw binary unless the format is specified using the `--image-type` argument.
+
+For the raw binary format, there is a specified entry point of the image that is documented in the board-specific
 instructions in this manual, under [Board Support Packages](#bsps). If the image is not loaded at this address
 it should still work by relocating itself upon boot.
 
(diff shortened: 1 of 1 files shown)


docs/manual.md
```diff
@@ -1159,7 +1167,25 @@ Build the firmware payload using platform `fpga/cheshire`.
...
+```sh
+dtc -I dts -O dtb seL4/tools/dts/cheshire.dts > cheshire.dtb
+```
+
+Now you can load the images into U-Boot at the appropriate addresses, for example via TFTP:
+
+    => setenv imgaddr "0xB0000000"
+    => setenv dtbaddr "0xA0000000"
+    => tftpboot ${imgaddr} /path/to/loader.img
+    => tftpboot ${dtbaddr} /path/to/cheshire.dtb
```
midnightveil: cheshire is not a good example, tftpboot doesn't work

docs/manual.md
```diff
@@ -1182,9 +1208,10 @@ Using a GDB prompt via openOCD:
 
         (gdb) load /path/to/opensbi/fw_jump.elf
 
-4.  Load the Microkit image:
+4. Load the Microkit image, which starts after the uImage and Linux headers (128 bytes into the uImage).
+   So we will load the uImage at `0x8fffff80`, such that the Microkit image starts at `0x90000000`:
 
-        (gdb) restore /path/to/loader.img binary 0x90000000
+        (gdb) restore /path/to/loader.img binary 0x8fffff80
```
midnightveil: Please don't for Cheshire, an ELF is more useful, or binary. 

(also this will 100% break all the other users of our CI for cheshire - because again, we don't load over the network, it's special)

tool/microkit/src/uimage.rs
```diff
@@ -0,0 +1,148 @@
...
+
+use crate::struct_to_bytes;
+use crate::{crc32::crc32, sel4::Arch};
+use std::fs::File;
+use std::io::Write;
+
+const UIMAGE_NAME: &str = "seL4 Microkit";
+
+// OS-type code
+const IH_OS_LINUX: u8 = 5;
```
midnightveil, line 19: fun: there's also `IH_OS_ELF`(which mentions "use cases such as seL4")

docs/manual.md
dreamliner787-9: My bad, I misunderstood an offline conversation earlier about this, will fix.

tool/microkit/src/main.rs
```diff
@@ -73,26 +73,37 @@ fn print_help(available_boards: &[String]) {
...
+                    Arch::Aarch64 | Arch::Riscv64 => Ok(ImageOutputType::Binary),
+                    Arch::X86_64 => Err(format!(
+                        "building the output image as binary is unsupported for target architecture '{arch}'"
+                    )),
+                },
+                "elf" => Ok(ImageOutputType::Elf),
+                "uimage" => match arch {
+                    Arch::Riscv64 => Ok(ImageOutputType::Uimage),
+                    Arch::X86_64 | Arch::Aarch64 => Err(format!(
+                        "building the output image as uImage is unsupported for target architecture '{arch}'"
```
Indanz, line 102: I thought RISC-V followed Aarch64's uImage format, so why isn't it supported?

tool/microkit/src/uimage.rs
```diff
@@ -0,0 +1,148 @@
...
+    ih_os: u8,               // Operating System
+    ih_arch: u8,             // CPU architecture
+    ih_type: u8,             // Image Type
+    ih_comp: u8,             // Compression Type
+    ih_name: [u8; IH_NMLEN], // Image Name
+}
+
+// Bits 0:15 minor, bits 16:31 major
+// Linux defines this to be 0.2 currently
+const LINUX_RISCV_HEADER_VERSION: u32 = 2 << 16;
```
Indanz, line 54: Shouldn't that be 2.0?

Ivan-Velickovic: Yes this is wrong right now.

tool/microkit/src/main.rs
Ivan-Velickovic: Haven't had a need for uImage for AArch64 whereas for reliable booting on RISC-V with U-Boot they're necessary.

Since we were trying to get the release out we just left out AArch64 uImage support, although it would be easy to add.

tool/microkit/src/uimage.rs
Ivan-Velickovic: Fixed in 119044f9573674342cedb9694142cce7b832d2ff

PR #385, opened 2025-11-25
Default image type fix

dreamliner787-9:
- Fixed a bug in the tool that doesn't select the correct default image types for certain boards.
- Switched the default image type from binary to ELF for `ariane`, `cheshire` and `serengeti`.

### DIFF
--- a/docs/manual.md
+++ b/docs/manual.md
@@ -1139,14 +1139,16 @@ on the Digilent Genesys2 board. CVA6 is an open-source RISC-V (rv64i) processor.
 Microkit support expects that a compatible RISC-V SBI (e.g OpenSBI) has executed before
 jumping to the beginning of the loader image.
 
-Note that the loader link address is 0x90000000 and this is where the binary must
-be located and where OpenSBI (or U-Boot) should begin execution.
+Microkit outputs an ELF for this platform.
 
 You may compile OpenSBI with the Microkit image as a payload, or alternately install
 OpenSBI (with U-Boot optionally) to the SD card.
 
 If you are booting from U-Boot, use the following command to start the system image:
-    => go 0x90000000
+
+    => bootelf ${image_addr}
+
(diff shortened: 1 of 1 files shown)


docs/manual.md
```diff
@@ -1139,14 +1139,15 @@ on the Digilent Genesys2 board. CVA6 is an open-source RISC-V (rv64i) processor.
...
 
-Note that the loader link address is 0x90000000 and this is where the binary must
-be located and where OpenSBI (or U-Boot) should begin execution.
+Microkit outputs an ELF for this device.
 
 You may compile OpenSBI with the Microkit image as a payload, or alternately install
 OpenSBI (with U-Boot optionally) to the SD card.
 
 If you are booting from U-Boot, use the following command to start the system image:
-    => go 0x90000000
```
midnightveil, line 1149: Cheshire (if it had SMP support), using bootelf would still be wrong when we use uboot. Though I guess this is better than `go` here.

(you do want to use uimage for uboot, but it's not easy to use uboot, or any point without ethernet)

midnightveil: I think we should just remove the uboot part from this manual, really: https://docs.sel4.systems/Hardware/cheshire.html just do what upstream does.

docs/manual.md
```diff
@@ -1164,23 +1165,13 @@ Microkit will produce a raw binary file by default, so when using U-Boot run the
 Support is available for [Cheshire](https://github.com/pulp-platform/cheshire).
 It is an SoC design based on the CVA6 core, implementing a 64-bit RISC-V CPU.
 
-Microkit outputs a raw binary for this device. Several steps are required in order to boot.
+Microkit outputs an ELF for this device. Several steps are required in order to boot.
```
Ivan-Velickovic:
```suggestion
Microkit outputs an ELF for this platform. Several steps are required in order to boot.
```

Just since we're changing this anyways.

PR #399, opened 2025-11-27
tool: fix version in Linux uImage header

Ivan-Velickovic:


Ivan-Velickovic:
I should not that this doesn't seem to make any difference when booting with U-Boot but we should fix it anyways.

PR #400, opened 2025-11-28
tool: fix cpu error message

Ivan-Velickovic:


Indanz:
No criticism, just an observation: If you open a PR and immediately close it by merging it, having a PR adds not much value compared to normal committing and directly pushing to the main git branch.

I suppose Github makes doing it this way easy, because when you make a branch, it gives an option to create a PR and then does the merge for you, so it's only a few clicks on the website, but still. I guess it does run the CI checks and builds, so that's something at least.

Ivan-Velickovic:
I used to directly push to main but my colleagues prefer if I make PRs so I stopped doing that, this way they get email notifications about the changes.

And like you say, with the PR I get to see CI checks before I merge.

The PRs also give a chance for people to comment on the changes we make which is usually useful (e.g the bug you noticed with the Linux image header version yesterday).

PR #403, opened 2025-12-02
tool: emit empty strtab when serialising ELF

dreamliner787-9:
Emit an empty strtab when serialising ELF. To fix this issue when converting the tool outputted ELF back into a binary:
```
% riscv64-unknown-elf-objcopy -O binary loader.elf loader.bin
riscv64-unknown-elf-objcopy: warning: loader.elf has a corrupt string table index
riscv64-unknown-elf-objcopy: unable to modify 'loader.elf' due to errors
```

### DIFF
--- a/tool/microkit/src/elf.rs
+++ b/tool/microkit/src/elf.rs
@@ -108,6 +108,7 @@ const PF_R: u32 = 0x4;
 
 /// ELF section-header type (`sh_type`)
 const SHT_PROGBITS: u32 = 0x1;
+const SHT_STRTAB: u32 = 0x3;
 
 /// ELF section-header flags (`sh_flags`)
 const SHF_WRITE: u64 = 0x1;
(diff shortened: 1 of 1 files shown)


tool/microkit/src/elf.rs
```diff
@@ -617,6 +619,33 @@ impl ElfFile {
...
+            info: 0,
+            addralign: 0,
+            entsize: 0,
+        };
+
+        // Uncomment this if more data needs to be written.
+        // data_off_watermark += 1;
+        // Have to comment out to keep clippy happy, else:
+        // "warning: value assigned to `data_off_watermark` is never read", or if we try to allow with `#[allow(unused_assignments)]`:
+        // "error[E0658]: attributes on expressions are experimental"
```
midnightveil: https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=7239ecb018912b29a4ba0631ee050f94

midnightveil: or just remove these commented lines entirely?

Ivan-Velickovic:
I was able to get the objcopy working with the segment having zero size so maybe we should just remove the fake data part.

Ivan-Velickovic:
> I was able to get the objcopy working with the segment having zero size so maybe we should just remove the fake data part.

Nevermind, looks like GNU binutils is fine with zero size, but not LLVM...

PR #404, opened 2025-12-11
Add support for FPU flag for PDs

JDuchniewicz:
In conjunction with https://github.com/seL4/rust-sel4/pull/316 allows for using the `fpu` flag to specify that a PD can be using the FPU.

### DIFF
--- a/docs/manual.md
+++ b/docs/manual.md
@@ -723,9 +723,7 @@ This will set the program counter of the vCPU to `entry_point`.
 
 Stop the execution of the VM's virtual CPU with ID `vcpu`.
 
-## `void microkit_vcpu_arm_inject_irq(microkit_child vcpu, seL4_Uint16 irq,
-seL4_Uint8 priority, seL4_Uint8 group,
-seL4_Uint8 index)`
+## `void microkit_vcpu_arm_inject_irq(microkit_child vcpu, seL4_Uint16 irq, seL4_Uint8 priority, seL4_Uint8 group, seL4_Uint8 index)`
 
 Inject a virtual ARM interrupt for a virtual CPU `vcpu` with IRQ number `irq`.
 The `priority` (0-31) determines what ARM GIC (generic interrupt controller)


Indanz:
There is a good reason why I made the FPU enabled by default.

If you do disable it by default, you better give a big warning that people should compile the code with -mgeneral-regs-only, otherwise they'll get crashes.

JDuchniewicz:
Yes. Now it's enabled by default.

tool/microkit/src/capdl/builder.rs
```diff
@@ -911,6 +912,7 @@ pub fn build_capdl_spec(
                             affinity: Word(vcpu.cpu.0.into()),
                             prio: virtual_machine.priority,
                             max_prio: virtual_machine.priority,
+                            fpu: true,
```
midnightveil: we should probably have a VM fpu attribute here too

tool/microkit/src/sdf.rs
```diff
@@ -607,6 +610,13 @@ impl ProtectionDomain {
             ));
         }
 
+        // FPU is enabled by default
+        let fpu = if let Some(xml_fpu) = node.attribute("fpu") {
```
midnightveil, line 620: https://github.com/seL4/microkit/blob/95044b3962f2b3c14f18fdfb1b8b7eda9f944d49/tool/microkit/src/sdf.rs#L390

use str_to_bool

also add tests for XML parsing please

tool/microkit/src/capdl/builder.rs
```diff
@@ -322,6 +322,7 @@ impl CapDLSpecContainer {
             prio: 0,
             max_prio: 0,
             resume: false,
+            fpu: true,
```
midnightveil: should come from pd.fpu, no?

tool/microkit/src/capdl/builder.rs
Indanz: Very small chance anyone would want to disable the FPU for a VM though.

Also, this wouldn't work for x86, where the flag is used for the host task and the FPU is always enabled for the guest itself.

tool/microkit/src/capdl/builder.rs
dreamliner787-9: As discussed offline, `fpu: true` is fine here. As this function's purpose is to only create what can be inferred from just the ELF, while leaving everything else with reasonable defaults. Refining the TCB will happen at a later step in the caller at https://github.com/seL4/microkit/blob/23f1f6decbafa35f78a52fc080cccbed0e84bce4/tool/microkit/src/capdl/builder.rs#L972

tool/microkit/src/capdl/builder.rs
JDuchniewicz: As I understand it it will be refined at a later step during the system setup as mentioned below:

> As discussed offline, `fpu: true` is fine here. As this function's purpose is to only create what can be inferred from just the ELF, while leaving everything else with reasonable defaults. Refining the TCB will happen at a later step in the caller at
> 
> https://github.com/seL4/microkit/blob/23f1f6decbafa35f78a52fc080cccbed0e84bce4/tool/microkit/src/capdl/builder.rs#L972

docs/manual.md
```diff
@@ -988,6 +988,7 @@ It supports the following attributes:
   Must be be between 4KiB and 16MiB and be 4K page-aligned. Defaults to 8KiB.
 * `cpu`: (optional) set the physical CPU core this PD will run on. Defaults to zero.
 * `smc`: (optional, only on ARM) Allow the PD to give an SMC call for the kernel to perform.. Defaults to false.
+* `fpu`: (optional) disable the FPU support for this PD. Defaults to true.
```
Ivan-Velickovic:
```suggestion
* `fpu`: (optional) whether this PD can access the FPU. Defaults to true.
```

JDuchniewicz:
https://github.com/seL4/rust-sel4/pull/316 is merged, we can proceed with this one

Willmish:
I believe https://github.com/seL4/microkit/blob/c83d507e1a0a47cb63130d594c525f567044db3f/Cargo.toml#L16 needs to be bumped to a more recent commit including https://github.com/seL4/rust-sel4/commit/2e06ec18da1cd0d119a76af586bed4b235f28979

JDuchniewicz:
@Willmish @Ivan-Velickovic I think the new domain-related functionality has to also be implemented/adjusted, as my PR was merged after that and cannot really base on it in isolation.

docs/manual.md
```diff
@@ -1008,7 +1008,7 @@ It supports the following attributes:
   Must be be between 4KiB and 16MiB and be 4K page-aligned. Defaults to 8KiB.
 * `cpu`: (optional) set the physical CPU core this PD will run on. Defaults to zero.
 * `smc`: (optional, only on ARM) Allow the PD to give an SMC call for the kernel to perform.. Defaults to false.
-* `fpu`: (optional) whether this PD can access the FPU. Defaults to true.
+* `fpu_disabled`: (optional) whether this PD can access the FPU. Defaults to false.
```
Ivan-Velickovic: Just because seL4 has it a particular way doesn't mean the Microkit level has to specify it the same way. Since Microkit gets to choose its own defaults it can just leave it as defaulting to true.

Ivan-Velickovic: I will revert the commit that changes it to `fpu_disabled`, having just `fpu` is much easier to read.

JDuchniewicz: Okay, I thought we are aiming for consistency - I think it saves developers head-scratching if we keep everything consistent. But I am not opposed.

JDuchniewicz: As discussed offline - it makes more sense to stick to the `fpu` convention instead. Closing.

PR #406, opened 2026-01-05
Enable mapping extra TCB and SC caps into PD's

Kswin01:
> [!NOTE]
> See the modifications to manual.md for the behaviour, or alternatively, the old version of the PR description which is roughly similar but differs in the details:

---

This PR adds functionality to map arbitrary TCB and SC caps of other PD's into another PD. This can allow a user to share a SC cap of one PD to another without having to explicitly make the destination PD the parent (and therefore have to also handle faults of the child).

Adding a cap to a PD requires adding a `<cap/>` node to a PD: 
```xml
    <protection_domain name="parent" priority="55">
        <program_image path="parent.elf" />
        <cap type="tcb" pd="child" dest_cspace_slot="5" />
        <cap type="sc" pd="parent" dest_cspace_slot="6" />
    </protection_domain>
```

You must specify the type of cap (currently only implemented for TCBs and SCs but can be extended in the future). Additionally, the name of the pd that is the source of that cap. The destination cspace slot is an index into the `BASE_USER_CAPS` region of the cspace.

This PR is also reflected in: https://github.com/au-ts/microkit_sdf_gen/tree/cap_sharing

### DIFF
--- a/docs/manual.md
+++ b/docs/manual.md
@@ -984,6 +984,13 @@ virtual CPU with ID `vcpu`.
 Write the registers of a given virtual CPU with ID `vcpu`. The `regs` argument is the pointer to
 the struct of registers `seL4_VCPUContext` that are written from.
 
+## `seL4_CPtr microkit_cspace_slot_to_cptr(seL4_Word slot)` {#libmicrokit_cspace_slot_to_cptr}
+
+Converts the slot identifier of the `<cspace>`'s capability element into an
+`seL4_CPtr` value to be used in `libsel4` calls by the PD.
+
+If the slot exceeds the valid range of inputs (`0 <= slot < MICROKIT_MAX_USER_CAPS`), it returns the value `seL4_CapNull`.
+
 # System Description File {#sysdesc}
 
 This section describes the format of the System Description File (SDF).
@@ -1025,6 +1032,7 @@ Additionally, it supports the following child elements:
 * `protection_domain`: (zero or more) Describes a child protection domain.
 * `virtual_machine`: (zero or one) Describes a child virtual machine.
 * `ioport`: (zero or more) Describes an I/O port, x86-64 only.
+* `cspace`: (zero or one) Describes ["extra" capabilities](#sdf-cspace) in the microkit-provided CSpace.
 
 The `program_image` element has the following attributes:
 
@@ -1124,6 +1132,29 @@ It supports the following attributes:
 * `prefill_path`: (optional) Path to a file containing data that the memory region will be filled with at initialisation.
 
 The `memory_region` element does not support any child elements.
+
+## `cspace` {#sdf-cspace}
+
+The `cspace` element represents the *Capability Space* of each protection domain.
+This is an advanced feature designed for users of Microkit who want more complex runtime behaviour than the system description file allows.
+For more details on how CSpaces and capabilities work, please see the ['Capability Spaces' section of the seL4 reference manual](https://sel4.systems/Info/Docs/seL4-manual-latest.pdf#chapter.3).
+
+It supports no attributes, but supports the following elements as children:
+
+* `cap_tcb`: A capability to a protection domain's Thread Control Block (TCB).
+* `cap_sc`: A capability to a protection domain's Scheduling Context (SC).
+  If the protection domain is passive, this is a capability to the notification's scheduling context.
+* `cap_vspace`: A capability to a protection domain's VSpace.
+* `cap_cspace`: A capability to a protection domain's CSpace.
+
+All of the elements support the `slot` attribute, which is is an opaque identifier used to address the capability at runtime.
+To convert the `slot` to an `seL4_CPtr`, use the [`seL4_CPtr microkit_cspace_slot_to_cptr(seL4_Word slot)`](#libmicrokit_cspace_slot_to_cptr) function.
+
+See the 'cap_sharing' example packaged in your SDK or [on GitHub](https://github.com/seL4/microkit/tree/main/example/cap_sharing).
+
+All capability elements (currently) all support the `pd` attribute, the name of the protection domain that the capability is from.
+For instance, `<cap_tcb slot="1" pd="alpha">` will place the TCB of PD 'alpha' in the CSpace of the current PD.
+
 
 ### Page sizes by architecture
 
--- a/example/cap_sharing/secondary.c
+++ b/example/cap_sharing/secondary.c
@@ -1,0 +1,18 @@
+/*
+ * Copyright 2026, UNSW
+ *
+ * SPDX-License-Identifier: BSD-2-Clause
+ */
+#include <stdint.h>
+#include <microkit.h>
+
+void init(void)
+{
+    microkit_dbg_puts("|secondary| hello, world\n");
+}
+
+void notified(microkit_channel ch)
+{
+    microkit_dbg_puts("|secondary| notified\n");
+    microkit_notify(ch);
+}
--- a/tool/microkit/src/capdl/builder.rs
+++ b/tool/microkit/src/capdl/builder.rs
@@ -24,8 +24,8 @@ use crate::{
     },
     elf::ElfFile,
     sdf::{
-        CpuCore, SysMap, SysMapPerms, SystemDescription, BUDGET_DEFAULT, MONITOR_PD_NAME,
-        MONITOR_PRIORITY,
+        CapMapType, CpuCore, SysMap, SysMapPerms, SystemDescription, BUDGET_DEFAULT,
+        MONITOR_PD_NAME, MONITOR_PRIORITY,
     },
     sel4::{Arch, Config, PageSize},
     util::{ranges_overlap, round_down, round_up},
@@ -95,8 +95,14 @@ const PD_BASE_PD_TCB_CAP: u64 = PD_BASE_IRQ_CAP + 64;
 const PD_BASE_VM_TCB_CAP: u64 = PD_BASE_PD_TCB_CAP + 64;
 const PD_BASE_VCPU_CAP: u64 = PD_BASE_VM_TCB_CAP + 64;
 const PD_BASE_IOPORT_CAP: u64 = PD_BASE_VCPU_CAP + 64;
+// The following region can be used for whatever the user wants to map into their
+// cspace. We restrict them to use this region so that they don't accidently
+// overwrite other parts of the cspace. The cspace slot that the users provide
+// for mapping in extra caps such as TCBs and SCs will be as an offset to this
+// index. We are bounding this to 128 slots for now.
+const PD_BASE_USER_CAPS: u64 = PD_BASE_IOPORT_CAP + 64;
 
-pub const PD_CAP_SIZE: u32 = 512;
+pub const PD_CAP_SIZE: u32 = 1024;
 const PD_CAP_BITS: u8 = PD_CAP_SIZE.ilog2() as u8;
 const PD_SCHEDCONTEXT_EXTRA_SIZE: u64 = 256;
 const PD_SCHEDCONTEXT_EXTRA_SIZE_BITS: u64 = PD_SCHEDCONTEXT_EXTRA_SIZE.ilog2() as u64;
@@ -116,6 +122,9 @@ struct PDShadowCspace {
     cspace: ObjectId,
     notification: ObjectId,
     endpoint: Option<ObjectId>,
+    sched_context: ObjectId,
+    tcb: ObjectId,
+    vspace: ObjectId,
 }
 
 pub struct CapDLSpecContainer {
@@ -1064,6 +1073,9 @@ pub fn build_capdl_spec(
                 cspace: pd_cnode_obj_id,
                 endpoint: pd_ep_obj_id,
                 notification: pd_ntfn_obj_id,
+                sched_context: pd_sc_obj_id,
+                vspace: pd_vspace_obj_id,
+                tcb: pd_tcb_obj_id,
             },
         );
     }
@@ -1136,7 +1148,36 @@ pub fn build_capdl_spec(
     }
 
     // *********************************
-    // Step 5. Sort the root objects
+    // Step 5. Handle extra cap mappings
+    // *********************************
+
+    for (pd_dest_idx, pd) in system.protection_domains.iter().enumerate() {
+        let pd_dest_cspace_id = pd_shadow_cspaces[&pd_dest_idx].cspace;
+
+        for cap_map in pd.cap_maps.iter() {
+            // TODO: Once we add more CapMap options, they might not all have
+            // the pd_name. But for now, they do.
+            let pd_src_shadow_cspace = &pd_shadow_cspaces[&cap_map.pd.unwrap()];
+
+            let cap_map_obj = match cap_map.cap_type {
+                CapMapType::Tcb => capdl_util_make_tcb_cap(pd_src_shadow_cspace.tcb),
+                CapMapType::Sc => capdl_util_make_sc_cap(pd_src_shadow_cspace.sched_context),
+                CapMapType::VSpace => capdl_util_make_sc_cap(pd_src_shadow_cspace.vspace),
+                CapMapType::CSpace => capdl_util_make_sc_cap(pd_src_shadow_cspace.cspace),
+            };
+
+            // Map this into the destination pd's cspace and the specified slot.
+            capdl_util_insert_cap_into_cspace(
+                &mut spec_container,
+                pd_dest_cspace_id,
+                (PD_BASE_USER_CAPS + cap_map.slot) as u32,
+                cap_map_obj,
+            );
+        }
+    }
+
+    // *********************************
+    // Step 6. Sort the root objects
     // *********************************
     // The CapDL initialiser expects objects with paddr to come first, then sorted by size so that the
     // allocation algorithm at run-time can run more efficiently.
--- a/tool/microkit/src/capdl/builder.rs
+++ b/tool/microkit/src/capdl/builder.rs
@@ -112,6 +112,12 @@ pub struct ExpectedAllocation {
     pub paddr: u64,
 }
 
+struct PDShadowCspace {
+    cspace: ObjectId,
+    notification: ObjectId,
+    endpoint: Option<ObjectId>,
+}
+
 pub struct CapDLSpecContainer {
     pub spec: Spec<FrameFill>,
     /// Track allocations as we build the system for later use by the report.
@@ -573,10 +579,9 @@ pub fn build_capdl_spec(
         None
     };
 
-    // Keep tabs on each PD's CSpace, Notification and Endpoint objects so we can create channels between them at a later step.
-    let mut pd_id_to_cspace_id: HashMap<usize, ObjectId> = HashMap::new();
-    let mut pd_id_to_ntfn_id: HashMap<usize, ObjectId> = HashMap::new();
-    let mut pd_id_to_ep_id: HashMap<usize, ObjectId> = HashMap::new();
+    // This object keeps track of object IDs for various 'important' / nameable kernel objects for
+    // each PD so that we can make various references to them at later steps.
+    let mut pd_shadow_cspaces: HashMap<usize, PDShadowCspace> = HashMap::new();
 
     // Keep track of the global count of vCPU objects so we can bind them to the monitor for setting TCB name in debug config.
     // Only used on ARM and RISC-V as on x86-64 VMs share the same TCB as PD's which will have their TCB name set separately.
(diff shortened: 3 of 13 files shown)


tool/microkit/src/capdl/builder.rs
```diff
@@ -1079,7 +1116,33 @@ pub fn build_capdl_spec(
     }
 
     // *********************************
-    // Step 5. Sort the root objects
+    // Step 5. Handle extra cap mappings
+    // *********************************
+
+    for (pd_dest_idx, pd) in system.protection_domains.iter().enumerate() {
+        let pd_dest_cspace_id = *pd_id_to_cspace_id.get(&pd_dest_idx).unwrap();
```
midnightveil: if you're going to always panic on .get just do a pd_to_cspace_id[pd_dest_idx] form

tool/microkit/src/capdl/builder.rs
```diff
@@ -1079,7 +1116,33 @@ pub fn build_capdl_spec(
...
+
+        for cap_map in pd.cap_maps.iter() {
+            let pd_src_idx = pd_name_to_id.get(&cap_map.pd_name).ok_or(format!(
+                "PD: '{}', does not exist when trying to map extra TCB cap into PD: '{}'",
+                cap_map.pd_name, pd.name
+            ))?;
+
+            let pd_obj = pd_shadow_cspace.get(pd_src_idx).unwrap()[cap_map.cap_type as usize]
```
midnightveil: ditto

tool/microkit/src/sdf.rs
```diff
@@ -275,6 +276,22 @@ pub struct ProtectionDomain {
     text_pos: Option<roxmltree::TextPos>,
 }
 
+#[derive(Debug, PartialEq, Eq, Clone, Copy)]
+pub enum CapMapType {
+    Tcb = 0,
+    Sc,
+    Vspace,
+    Cnode,
+    __Len,
```
midnightveil: no, this isn't how rust's enums work, they're not just mini constants like C
 please don't, you won't be able to match on this type nicely without discarding __Len

tool/microkit/src/sdf.rs
```diff
@@ -1214,6 +1236,45 @@ impl VirtualMachine {
...
+        let cap_type = match checked_lookup(xml_sdf, node, "type")? {
+            "tcb" => CapMapType::Tcb,
+            "sc" => CapMapType::Sc,
+            "vspace" => CapMapType::Vspace,
+            "cnode" => CapMapType::Cnode,
+            _ => {
+                return Err(value_error(
+                    xml_sdf,
+                    node,
+                    "type must be 'tcb' or 'sc' ".to_string(),
```
midnightveil: seems wrong?

tool/microkit/src/sdf.rs
```diff
@@ -1214,6 +1236,45 @@ impl VirtualMachine {
...
+                ))
+            }
+        };
+
+        let pd_name = checked_lookup(xml_sdf, node, "pd")?.to_string();
+        let dest_cspace_slot =
+            sdf_parse_number(checked_lookup(xml_sdf, node, "dest_cspace_slot")?, node)?;
+
+        if dest_cspace_slot >= 128 {
```
midnightveil: make this a constant somewhere. also add a comment saying to update the corresponding C if you change this

tool/microkit/src/sdf.rs
```diff
@@ -1214,6 +1236,45 @@ impl VirtualMachine {
...
+            return Err(value_error(
+                xml_sdf,
+                node,
+                "There are only 128 destination cspace slots available. Max slot allowed is 63"
```
midnightveil: ??

tool/microkit/src/sdf.rs
```diff
@@ -1890,6 +1950,33 @@ pub fn parse(filename: &str, xml: &str, config: &Config) -> Result<SystemDescrip
         }
     }
 
+    // Ensure that there are no overlapping extra cap maps in the user caps region
+    // and we are not mapping in the same cap from the same source more than once
+    for pd in &pds {
```
midnightveil, line 2218: stylistic nit: either consistently user .iter() or the &vec form. I prefer the former but you use the latter below.

midnightveil (dismissed):

tool/microkit/src/sdf.rs
Kswin01: I've added in a constant for number of types instead

tool/microkit/src/sdf.rs
Kswin01: Updated the below loop for consistency. The rest of this code section uses the &vec form so I'll follow that

tool/microkit/src/capdl/builder.rs
```diff
@@ -555,6 +568,8 @@ pub fn build_capdl_spec(
     for (pd_global_idx, pd) in system.protection_domains.iter().enumerate() {
         let elf_obj = &elfs[pd_global_idx];
 
+        pd_name_to_id.insert(pd.name.clone(), pd_global_idx);
```
Indanz: Why clone? Can't you just use a read-only ref?

tool/microkit/src/capdl/builder.rs
```diff
@@ -564,20 +579,26 @@ pub fn build_capdl_spec(
...
 
+        let pd_tcb_obj = capdl_util_make_tcb_cap(pd_tcb_obj_id);
+        let pd_vspace_obj = capdl_util_make_page_table_cap(pd_vspace_obj_id);
+
+        pd_shadow_cspace
+            .entry(pd_global_idx)
+            .or_insert_with(|| vec![None; CAP_MAP_TYPES])[CapMapType::Tcb as usize] =
+            Some(pd_tcb_obj.clone());
+        pd_shadow_cspace.get_mut(&pd_global_idx).unwrap()[CapMapType::Vspace as usize] =
+            Some(pd_vspace_obj.clone());
```
Indanz: I don't know Rust, but surely there is a less ugly way than this of doing something extremely simple.

Why as usize and why the clone? You just added CapMapType, so why not make sure it's the right type to begin with? Why the Some? If not necessary, it would simplify lookup later I hope.

tool/microkit/src/capdl/builder.rs
```diff
@@ -564,20 +579,26 @@ pub fn build_capdl_spec(
...
+
         // In the benchmark configuration, we allow PDs to access their own TCB.
         // This is necessary for accessing kernel's benchmark API.
         if kernel_config.benchmark {
-            caps_to_insert_to_pd_cspace.push(capdl_util_make_cte(
-                PD_TCB_CAP_IDX as u32,
-                capdl_util_make_tcb_cap(pd_tcb_obj_id),
-            ));
+            caps_to_insert_to_pd_cspace
+                .push(capdl_util_make_cte(PD_TCB_CAP_IDX as u32, pd_tcb_obj));
```
Indanz: PD_TCB_CAP_IDX is an u64, why the cast to u32?

Can't you use consistent u32 or u64 for all CSpace types and functions?

tool/microkit/src/capdl/builder.rs
```diff
@@ -540,11 +547,17 @@ pub fn build_capdl_spec(
...
+    // Mapping between pd name and id for faster lookups
+    let mut pd_name_to_id: HashMap<String, usize> = HashMap::new();
+
     // Keep tabs on each PD's CSpace, Notification and Endpoint objects so we can create channels between them at a later step.
     let mut pd_id_to_cspace_id: HashMap<usize, ObjectId> = HashMap::new();
     let mut pd_id_to_ntfn_id: HashMap<usize, ObjectId> = HashMap::new();
     let mut pd_id_to_ep_id: HashMap<usize, ObjectId> = HashMap::new();
 
+    // Keep tabs on caps such as TCB and SC so that we can create additional mappings for the cap into other PD's cspaces.
+    let mut pd_shadow_cspace: HashMap<usize, Vec<Option<Cap>>> = HashMap::new();
```
Indanz: Why not a hash of cap arrays, or (nested) cap HashMap indexed by "tcb" and "sc" instead of the overly complicated vector construction? Both seem simpler than than what you have now.

tool/microkit/src/capdl/builder.rs
```diff
@@ -681,6 +706,14 @@ pub fn build_capdl_spec(
                 capdl_util_make_tcb_cap(pd_tcb_obj_id),
             );
 
+            // Allow the parent PD to access the child's SC:
+            capdl_util_insert_cap_into_cspace(
+                &mut spec_container,
+                *parent_cspace_obj_id,
+                (PD_BASE_PD_SC_CAP + pd.id.unwrap()) as u32,
+                capdl_util_make_sc_cap(pd_sc_obj_id),
+            );
```
Indanz: Surely unwrap isn't necessary on pd.id? It's not for e.g. pd.name. I know other places do it, but why? Same for the as u32, that seems avoidable.

tool/microkit/src/capdl/builder.rs
```diff
@@ -1079,7 +1114,33 @@ pub fn build_capdl_spec(
...
+
+            let pd_obj = pd_shadow_cspace[pd_src_idx][cap_map.cap_type as usize]
+                .as_ref()
+                .unwrap();
+            // Map this into the destination pd's cspace and the specified slot.
+            capdl_util_insert_cap_into_cspace(
+                &mut spec_container,
+                pd_dest_cspace_id,
+                (PD_BASE_USER_CAPS + cap_map.dest_cspace_slot) as u32,
+                pd_obj.clone(),
```
Indanz: Why the clone? If you want a copy, why as_ref above? (And wouldn't that be the default??)  I'm confused.

tool/microkit/src/sdf.rs
```diff
@@ -1890,6 +1951,33 @@ pub fn parse(filename: &str, xml: &str, config: &Config) -> Result<SystemDescrip
         }
     }
 
+    // Ensure that there are no overlapping extra cap maps in the user caps region
+    // and we are not mapping in the same cap from the same source more than once
```
Indanz, line 2217: Why wouldn't it be allowed to map the same cap twice from the same source? That may be useful for testing purposes, if nothing else.

And frankly, I wouldn't bother checking for overlapping indices either, but if I did, I would try to do it elsewhere if possible.

Indanz (commented):
As you can see from the comments, I don't know Rust, so take everything I say with a big grain of salt. But I'm pretty sure it doesn't need to be quite this ugly.

tool/microkit/src/capdl/builder.rs
Kswin01: The array index needs to be usize in rust. I'm using Some() because i define the vector type as `Option<Cap>`, and initialise the values to "None". I could get rid of this if using a hashmap as you have suggested. The clone is necessary as the `capdl_util_make_cte()` function takes ownership of the Cap, and rust will throw a "use of moved variable" error, requiring a clone

tool/microkit/src/capdl/builder.rs
Kswin01: the rust capdl initializer was refactored to use u32 while the capdl microkit tool was under development. This PR kept consistent with what is currently in the code base, but will look at a separate PR to change from u64 to u32 for cspace types

tool/microkit/src/capdl/builder.rs
Kswin01: pd.id is defined as `Option<u64>` as it's only set if that pd is a child, that requires an unwrap. As mentioned in another comment, the u32 and u64 difference in cspace types and functions was due to a refactor of the rust capdl initializer during the development of the microkit tool.

tool/microkit/src/capdl/builder.rs
Kswin01: This was from following the compiler suggestions from rust, first suggesting to use `as_ref` to borrow, then errored when using pd_obj due to the move of ownership when calling  `capdl_util_insert_cap_into_cspace` and suggesting to clone. I can just add the clone() and get rid of the as_ref().

tool/microkit/src/sdf.rs
Kswin01: I didn't think of a use-case that would require the same cap from the same source to be mapped more than once, if that's a use-case that people may actually require then I'll remove that check.

Overlapping indicies could very well be an error on the user's side, I think it's prudent to check for this, and we have to wait to parse all of the cap maps for a PD before checking for these

tool/microkit/src/capdl/builder.rs
Indanz: But why have typed caps instead of just numbers to begin with? You don't care what cap type it are, you just want to copy them around.

tool/microkit/src/sdf.rs
Indanz: That you can't think of a use case is not a reason to forbid something. That just causes extra code bloat, complexity and artificial limitations.

You can check for it during the parsing of a PD's cap map: Just error out if an index is already used.

midnightveil:
Remaining work on this to get this merged:

- [x] Test example
- [x] Documentation
- [x] Bikeshed the appearance (thought: instead of `<cap>`, doing a `<extra_caps>` element with child `(<endpoint>)` or `(<sc>)`; this would allow more fine grain attribute on the different cap types, instead of smooshing everything onto one `<cap>` element. (edit: `<cnode` element?)

Desired additions in later PRs.
- [ ] remove PD hierarchy. To do this we want to add a feature to automatically migrate one .system file to another; there should be no need for this to be manual user.
- [ ] Make it a fully functional replacement for the PD hierarchy. i.e. allow the fault endpoints to be set up arbitrarily. (it might make sense for this *not* to be the `<extra_caps>`, IDK
- [ ] Make the "user caps" its own cnode that is dynamically sized, there's no reason to limit this (see also: the untypeds PR).
- [ ] Auto-allocation of slots? (edit: maybe not?)
- [ ] perms

midnightveil:
We've bikeshed the appearance internally, and decided that it will look like:

```xml
<protection_domain name="alpha">
   <cspace> <!-- the root cnode (currently size 512) -->
       <!-- implicitly, not mentioned in the xml: the microkit caps -->
       
       <cap_tcb slot="1" pd="beta" />
       <cap_vspace slot="2" pd="beta" />
   </cspace>
</protection_domain>

```

The plan is soon to change the layout of this exactly; I'm going to make the layout of where the user caps be instead an implementation details and you should use a helper function.

<details><summary>Survey of cap types</summary>

We have:
1. Null cap (irrelevant)
2. Untyped cap. There are no static untypeds to refer to, so this cap type does not make sense. When we do the dynamic-remaining untypeds these are dynamic, and we give them a cnode to go into. So we can ignore this. badge is meaningless
3. Endpoint: not allowed, use channels for that
4. Notification: badge useful here, though can default to the original
4. reply cap: not given
5. cnode cap: this will probably want some kind of "name" field to address the cnode, or maybe "pd" to address the cspace. badge contains the 'guard' value which we do want to control.
6. thread/tcb: 1-1 with the pd name. badge means nothing.
7. irqcontrol: badge means nothing. there is only one of these, so pd is meaningless
8. irqhandler: badge has no meaning. 
9. arm smc: badge convey the smc ids that are allowed to use
10. x86 iospace: badge is like PCI id or something, IDK
11. riscv opensbi: badge conveys sbi ids that are allowed to use
12. schedcontext: only assocaited with pd, nothing 
13. frame caps: no badge, associated with ?? something not a pd
14. page table: no badge, associated with ?? not a pd
15. vspace: no badge, associated with a pd
16. asidcontrol/asidpool: no badge, not associated with anything

</details>

Indanz:
> 5\. reply cap: not given

Actually, for timeout fault handlers it makes sense to reply on behalf of the task for recovery purposes, especially if the server is reset.

midnightveil:
> Actually, for timeout fault handlers it makes sense to reply on behalf of the task for recovery purposes, especially if the server is reset.

well, regardless, the badge means nothing. and this would be associated with a PD, at least.

docs/manual.md
```diff
@@ -1125,6 +1126,29 @@ It supports the following attributes:
...
+* `cap_tcb`: A capability to a protection domain's Thread Control Block (TCB).
+* `cap_sc`: A capability to a protection domain's Scheduling Context (SC).
+  If the protection domain is passive, this is a capability to the notification's scheduling context.
+* `cap_vspace`: A capability to a protection domain's VSpace.
+* `cap_cspace`: A capability to a protection domain's CSpace.
+
+All of the elements support the `slot` attribute, which is is an opaque identifier used to address the capability at runtime.
+To convert the `slot` to an `seL4_CPtr`, use the `seL4_CPtr microkit_cspace_slot_to_cptr(seL4_Word slot)` function.
+
+See the 'cap_sharing' example packaged in your SDK or [on GitHub](https://github.com/seL4/microkit/tree/main/example/cap_sharing).
```
dreamliner787-9, line 1153: > See the 'cap_sharing' example packaged in your SDK

It might be worth specifying that the example in the source and is not part of the release package itself

example/cap_sharing/secondary.c
```diff
@@ -0,0 +1,18 @@
...
+ * Copyright 2026, UNSW
+ *
+ * SPDX-License-Identifier: BSD-2-Clause
+ */
+#include <stdint.h>
+#include <microkit.h>
+
+void init(void)
+{
+    microkit_dbg_puts("|secondary| hello, world\n");
```
dreamliner787-9, line 11: I reckon use the `microkit_name` global.

docs/manual.md
```diff
@@ -1125,6 +1126,29 @@ It supports the following attributes:
...
+It supports no attributes, but supports the following elements as children:
+
+* `cap_tcb`: A capability to a protection domain's Thread Control Block (TCB).
+* `cap_sc`: A capability to a protection domain's Scheduling Context (SC).
+  If the protection domain is passive, this is a capability to the notification's scheduling context.
+* `cap_vspace`: A capability to a protection domain's VSpace.
+* `cap_cspace`: A capability to a protection domain's CSpace.
+
+All of the elements support the `slot` attribute, which is is an opaque identifier used to address the capability at runtime.
+To convert the `slot` to an `seL4_CPtr`, use the `seL4_CPtr microkit_cspace_slot_to_cptr(seL4_Word slot)` function.
```
dreamliner787-9: You should also specify the range of valid slots, since it is bound by `MICROKIT_MAX_USER_CAPS`.

tool/microkit/tests/sdf/pd_cap_mappings_overlapping.system
```diff
@@ -0,0 +1,29 @@
...
+
+    <protection_domain name="pd_b">
+        <program_image path="test" />
+
+        <cspace>
+            <cap_tcb    slot="1" pd="pd_a" />
+            <cap_sc     slot="2" pd="pd_a" />
+            <cap_vspace slot="3" pd="pd_a" />
+            <cap_cspace slot="4" pd="pd_a" />
+            <!-- duplicate slot -->
```
dreamliner787-9, line 25: You should add a test for:
```xml
        <cspace>
            <cap_tcb    slot="1" pd="pd_a" />
            <cap_sc     slot="2" pd="pd_a" />
            <cap_vspace slot="3" pd="pd_a" />
            <cap_cspace slot="4" pd="pd_a" />
            <!-- duplicate type and pd -->
            <cap_cspace slot="5" pd="pd_a" />
        </cspace>
```

tool/microkit/src/capdl/builder.rs
```diff
@@ -573,10 +579,19 @@ pub fn build_capdl_spec(
         None
     };
 
-    // Keep tabs on each PD's CSpace, Notification and Endpoint objects so we can create channels between them at a later step.
-    let mut pd_id_to_cspace_id: HashMap<usize, ObjectId> = HashMap::new();
-    let mut pd_id_to_ntfn_id: HashMap<usize, ObjectId> = HashMap::new();
-    let mut pd_id_to_ep_id: HashMap<usize, ObjectId> = HashMap::new();
+    #[derive(Debug)]
+    struct PDShadowCspace {
```
dreamliner787-9: nit: I feel like we should put all type definitions outside of the function, even though I see the utility of being able to immediately tell what is going on. It also avoid creating a precedent of putting type definitions inside functions in other places.

docs/manual.md
midnightveil: But it is in the packaged SDK?

example/cap_sharing/secondary.c
midnightveil: It's kinda annoying though, since we don't have printf :(

tool/microkit/tests/sdf/pd_cap_mappings_overlapping.system
midnightveil: There's nothing wrong with duplicate types. That is perfectly allowed.

tool/microkit/src/capdl/builder.rs
midnightveil: Right, yes I will fix this, I just did it here temporarily.

docs/manual.md
midnightveil: Yes, I guess I should. The plan was to make it dynamic but that's harder for now.

example/cap_sharing/secondary.c
dreamliner787-9: Oh right I forgot, pls ignore this then.

tool/microkit/tests/sdf/pd_cap_mappings_overlapping.system
dreamliner787-9: It feels weird to me when you can refer to one thing in 2 slots

docs/manual.md
dreamliner787-9: Oh the examples are in the distribution, for some reason I thought they are not, ignore me then

tool/microkit/tests/sdf/pd_cap_mappings_overlapping.system
midnightveil: Please see previous discussion: https://github.com/seL4/microkit/pull/406#discussion_r2731973081

It makes the implementation more complex and it doesn't give you any actual benefits, and there are definitely situations where having duplicates could be useful.

midnightveil:
Commit history before I clean it up: https://github.com/au-ts/microkit/tree/archive/cap_sharing_history.

midnightveil:
The link-check is expected to be failing until after this PR is merged.

PR #415, opened 2026-02-13
use LibSel4UseThreadLocals to disable TLS ipcbuf

midnightveil:
See the motivation in the commit
    libsel4: make thread-local ipc buffer optional
in seL4 which adds this option.

Depends on https://github.com/seL4/seL4/pull/1593.


build_sdk.py
```diff
@@ -45,6 +45,13 @@
 DEFAULT_KERNEL_OPTIONS = {
     "KernelIsMCS": True,
     "KernelRootCNodeSizeBits": "17",
+    # Thread local storage is painful and annoying to configure.
+    # We'd really rather NOT use thread local storage (especially
+    # consider we never have more than one thread in a Vspace)
```
midnightveil: this should be considering, oops.

Issue #362, opened 2025-11-17
proto-rfc: the microkit API shouldn't exist (mostly)


midnightveil:
Or, at the very least, we should stop adding more functions to its API. [^1]

To be clear, this is my opinion, and for discussion.

## the issue

Currently, the microkit API includes functions like:

```c
static inline void microkit_notify(microkit_channel ch)
{
    seL4_Signal(BASE_OUTPUT_NOTIFICATION_CAP + ch);
}

static inline seL4_Word microkit_msginfo_get_count(microkit_msginfo msginfo)
{
    return seL4_MessageInfo_get_length(msginfo);
}

static inline void microkit_mr_set(seL4_Uint8 mr, seL4_Word value)
{
    seL4_SetMR(mr, value);
}

static inline seL4_Word microkit_mr_get(seL4_Uint8 mr)
{
    return seL4_GetMR(mr);
}
```

In my opinion, why? This is, in my opinion, a prime example of a leaky abstraction.

There are some good parts:

- You don't need to deal with seL4 capabilities, one of the most confusing parts.
- The API surface area (was) massively reduced.

However, it abstracts something so... simple it seems to have no point. The only thing it does is error handling (by... panicking) and translation of the "channel ID" to the actual cap ID.

Where this falls apart a bit more, is using APIs not exposed by microkit: for instance, anything that wants to be a `fault()` handler tends to want to call `seL4_TCB_ReadRegisters` (and write), but this isn't implemented: instead it gets called directly: https://github.com/au-ts/libvmm/blob/8f0665dea882130bb5f25311dcd47a1500d80ffe/src/arch/aarch64/fault.c#L264-L280. The same thing with other fault information.  Our sDDF "os abstraction" for notifications so that sDDF can work on other seL4 systems that aren't microkit needs to add another layer of abstraction, e.g.: https://github.com/au-ts/sddf/blob/81424f047b09363a1f9f95d293a51167f6403859/include/microkit/os/sddf.h#L31-L34.

In the end, the microkit API either needs to extend to the entire seL4 API, (at a loss of use for non-microkit seL4 libraries) or users need to poke at the internal cap layout. Hence, a "leaky abstraction".

[...]

PR #431, opened 2026-03-09
libmicrokit: add API for resuming x86 VCPU

dreamliner787-9:
Depends on https://github.com/seL4/seL4/pull/1534, I've worked around this for now by including `vmenter.h` in `libmicrokit/include/microkit.h`.

Previously, the Microkit event handler had no mechanism on x86 to resume a vCPU while simultaneously waiting for an incoming notification. This PR adds `microkit_vcpu_x86_on()` and `microkit_vcpu_x86_off()` to libmicrokit, providing a cleaner way to integrate x86 vCPU resumption with the Microkit event handler.

Due to a kernel limitation in `void NORETURN slowpath(syscall_t syscall)` of `src/arch/x86/c_traps.c`, only the VMM's TCB bound Notification is checked for a pending Notification. An endpoint object can't passed to `seL4_VMEnter()` so PPCs won't work while the VCPU is running. Technically, PPCs still work if the VCPU isn't resumed, but I feel like we should not expose this footgun to users.

### DIFF
--- a/docs/manual.md
+++ b/docs/manual.md
@@ -637,7 +637,10 @@ as to what fault occurred.
 The `reply_msginfo` argument is given by libmicrokit and can be used to reply to the fault.
 
 The returned `seL4_Bool` is whether or not to reply to the fault with the message `reply_msginfo`.
-Returning `seL4_True` will reply to the fault. Returning `seL4_False` will not reply to the fault.
+Returning `seL4_True` will reply to the fault and resume the child PD or vCPU at the fault restart program counter.
+For more details, please see the "Faults" section of the [seL4 Manual](https://sel4.systems/Info/Docs/seL4-manual-latest.pdf).
+
+Returning `seL4_False` will not reply to the fault and causes the child PD or vCPU to no longer run.
 
 You can use `microkit_msginfo_get_label` on `msginfo` to deduce what kind of fault happened
 (for example, whether it was a user exception or a virtual memory fault).
@@ -648,6 +651,43 @@ to handle it.
 To find the full list of possible faults that could occur and details regarding to replying to a particular
 kind of fault, please see the 'Faults' section of the
 [seL4 reference manual](https://sel4.systems/Info/Docs/seL4-manual-latest.pdf).
+
+### x86 VCPU fault
+Please see the 'VMX BASIC EXIT REASONS' section of the
+[Intel® 64 and IA-32 Architectures Software Developer’s Manual Combined Volumes: 1, 2A, 2B, 2C, 2D, 3A, 3B, 3C, 3D, and 4]
+(https://cdrdv2.intel.com/v1/dl/getContent/671200) for a list of possible VM Exit reasons.
+
+These message registers contain data relating to the VM Exit:
+- `SEL4_VMENTER_CALL_EIP_MR`: Instruction Pointer,
+- `SEL4_VMENTER_CALL_CONTROL_PPC_MR`: Primary Processor Based VM Execution Controls,
+- `SEL4_VMENTER_CALL_INTERRUPT_INFO_MR`: VM Entry Interruption-Information,
+- `SEL4_VMENTER_FAULT_REASON_MR`: VM Exit reason,
+- `SEL4_VMENTER_FAULT_QUALIFICATION_MR`: VM Exit qualification,
+- `SEL4_VMENTER_FAULT_INSTRUCTION_LEN_MR`: Length of instruction that caused the VM Exit,
+- `SEL4_VMENTER_FAULT_GUEST_PHYSICAL_MR`: Guest Physical Address of the VM Exit,
+- `SEL4_VMENTER_FAULT_RFLAGS_MR`: Guest FLAGS register,
+- `SEL4_VMENTER_FAULT_GUEST_INT_MR`: Guest interruptability,
+- `SEL4_VMENTER_FAULT_CR3_MR`: Guest CR3.
+
+Some of these message registers may not contain valid data depending on the VM Exit reason,
+please consult the Intel SDM for more details.
+
+These message registers contain the guest general purpose registers at the time of VM Exit:
+- `SEL4_VMENTER_FAULT_EAX`
+- `SEL4_VMENTER_FAULT_EBX`
+- `SEL4_VMENTER_FAULT_ECX`
+- `SEL4_VMENTER_FAULT_EDX`
+- `SEL4_VMENTER_FAULT_ESI`
+- `SEL4_VMENTER_FAULT_EDI`
+- `SEL4_VMENTER_FAULT_EBP`
+- `SEL4_VMENTER_FAULT_R8`
+- `SEL4_VMENTER_FAULT_R9`
+- `SEL4_VMENTER_FAULT_R10`
+- `SEL4_VMENTER_FAULT_R11`
+- `SEL4_VMENTER_FAULT_R12`
+- `SEL4_VMENTER_FAULT_R13`
+- `SEL4_VMENTER_FAULT_R14`
+- `SEL4_VMENTER_FAULT_R15`
 
 ## `microkit_msginfo microkit_ppcall(microkit_channel ch, microkit_msginfo msginfo)`
 
@@ -826,6 +866,22 @@ virtual CPU with ID `vcpu`.
 Write the registers of a given virtual CPU with ID `vcpu`. The `regs` argument is the pointer to
 the struct of registers `seL4_VCPUContext` that are written from.
 
+## `void microkit_vcpu_x86_on(void)`
+
+Start running the vCPU bound to this PD.
+
+The vCPU runs after every return from a Microkit entrypoint; it does not run concurrently with the native thread.
+
+The kernel performs minimal setup of the vCPU's architectural state, configuring only the hardware-mandated
+fixed bits and setting VMCS fields required to safely context-switch between VMX root and non-root operations.
+The caller is responsible for initialising all other architectural state such as the instruction pointer using
+`microkit_vcpu_x86_write_vmcs()`, in accordance with the
+[Intel SDM](https://cdrdv2.intel.com/v1/dl/getContent/671200).
+
+## `void microkit_vcpu_x86_off(void)`
+
+Stop the PD from switching to guest execution mode when a Microkit entrypoint returns.
+
 ## `seL4_CPtr microkit_cspace_root_slot_to_cptr(seL4_Word slot)` {#libmicrokit_cspace_root_slot_to_cptr}
 
 Converts the slot identifier of the `<cspace>`'s capability element into an
@@ -938,6 +994,8 @@ The `protection_domain` element has the same attributes as any other protection 
 * `id`: The ID of the child for the parent to refer to.
 * `setvar_id`: (optional) Specifies a symbol in the parent program image. This symbol will be rewritten with the ID of the child.
 
+On x86-64, a PD with a VCPU cannot have child PDs.
+
 The `virtual_machine` element has the following attributes:
 
 * `name`: A unique name for the virtual machine
@@ -948,6 +1006,7 @@ The `virtual_machine` element has the following attributes:
 Additionally, it supports the following child elements:
 
 * `vcpu`: (one or more) Describes the virtual CPU that will be tied to the virtual machine.
+    * On x86-64, there is a limit of one VCPU per PD.
 * `map`: (zero or more) Describes mapping of memory regions into the virtual machine.
 
 The `vcpu` element has the following attributes:
@@ -1027,6 +1086,7 @@ The `end` element has the following attributes:
 * `id`: Channel identifier in the context of the named protection domain. Must be at least 0 and less than 63.
 * `pp`: (optional) Indicates that the protection domain for this end can perform a protected procedure call to the other end; defaults to false.
         Protected procedure calls can only be to PDs of strictly higher priority.
+        On x86-64, PDs with virtual machines cannot receive protected procedure calls.
 * `notify`: (optional) Indicates that the protection domain for this end can send a notification to the other end; defaults to true.
 * `setvar_id`: (optional) Specifies a symbol in the program image. This symbol will be rewritten with the channel identifier.
 
--- a/libmicrokit/include/microkit.h
+++ b/libmicrokit/include/microkit.h
@@ -10,6 +10,9 @@
 #pragma once
 
 #include <sel4/sel4.h>
+#ifdef CONFIG_VTX
+#include <sel4/arch/vmenter.h>
+#endif /* CONFIG_VTX */
 
 typedef unsigned int microkit_channel;
 typedef unsigned int microkit_child;
@@ -47,6 +50,15 @@ extern char microkit_name[MICROKIT_PD_NAME_LENGTH];
 extern seL4_Bool microkit_have_signal;
 extern seL4_CPtr microkit_signal_cap;
 extern seL4_MessageInfo_t microkit_signal_msg;
+#if defined(CONFIG_VTX)
+struct microkit_x86_vcpu_state {
+    seL4_Bool is_on;
+    seL4_Word rip;
+    seL4_Word prim_proc_ctl;
+    seL4_Word irq_info;
+};
+extern struct microkit_x86_vcpu_state microkit_x86_vcpu_state;
+#endif
 
 /* Symbols for error checking libmicrokit API calls. Patched by the Microkit tool
  * to set bits corresponding to valid channels for this PD. */
@@ -386,28 +398,86 @@ static inline seL4_Uint32 microkit_x86_ioport_read_32(microkit_ioport ioport_id,
 
     return ret.result;
 }
-#endif
 
-#if defined(CONFIG_ARCH_X86_64) && defined(CONFIG_VTX)
+#if defined(CONFIG_VTX)
+/* Architecturally defined identifiers for a x86 VCPU's VMCS fields,
+ * see seL4 source: `include/arch/x86/arch/object/vcpu.h`
+ * or Intel® 64 and IA-32 Architectures Software Developer’s Manual
+ * Combined Volumes: 1, 2A, 2B, 2C, 2D, 3A, 3B, 3C, 3D, and 4
+ * Order Number: 325462-084US June 2024
+ * "Table B-8. Encodings for 32-Bit Control Fields (0100_00xx_xxxx_xxx0B)" and
+ * "Table B-14. Encodings for Natural-Width Guest-State Fields (0110_10xx_xxxx_xxx0B) (Contd.)".
+ * */
+#define MICROKIT_VMX_GUEST_RIP 0x0000681E
+#define MICROKIT_VMX_CONTROL_PRIMARY_PROCESSOR_CONTROLS 0x00004002
+#define MICROKIT_VMX_CONTROL_ENTRY_INTERRUPTION_INFO 0x00004016
+
 static inline seL4_Word microkit_vcpu_x86_read_vmcs(microkit_child vcpu, seL4_Word field)
 {
-    seL4_X86_VCPU_ReadVMCS_t ret;
-    ret = seL4_X86_VCPU_ReadVMCS(BASE_VCPU_CAP + vcpu, field);
-    if (ret.error != seL4_NoError) {
-        microkit_dbg_puts("microkit_x86_read_vmcs: error reading data\n");
-        microkit_internal_crash(ret.error);
+    /* This function assumes that a PD would only have access to 1 VCPU object.
+     * For these fields:
+     * - Guest RIP,
+     * - Primary Processor Based VM Execution Controls, and
+     * - VM Entry Interruption-Information
+     * they will be read from local variables that were written
+     * to by `microkit_vcpu_x86_write_vmcs()`. The kernel will service
+     * reading from other fields.
+     */
+
+    seL4_Word value;
+    switch (field) {
+    case MICROKIT_VMX_GUEST_RIP:
+        value = microkit_x86_vcpu_state.rip;
+        break;
+    case MICROKIT_VMX_CONTROL_PRIMARY_PROCESSOR_CONTROLS:
+        value = microkit_x86_vcpu_state.prim_proc_ctl;
+        break;
+    case MICROKIT_VMX_CONTROL_ENTRY_INTERRUPTION_INFO:
+        value = microkit_x86_vcpu_state.irq_info;
+        break;
+    default: {
+        seL4_X86_VCPU_ReadVMCS_t ret = seL4_X86_VCPU_ReadVMCS(BASE_VCPU_CAP + vcpu, field);
+        if (ret.error != seL4_NoError) {
+            microkit_dbg_puts("microkit_x86_read_vmcs: error reading data\n");
+            microkit_internal_crash(ret.error);
+        }
+        value = ret.value;
+        break;
+    }
     }
 
-    return ret.value;
+    return value;
 }
 
 static inline void microkit_vcpu_x86_write_vmcs(microkit_child vcpu, seL4_Word field, seL4_Word value)
 {
-    seL4_X86_VCPU_WriteVMCS_t ret;
-    ret = seL4_X86_VCPU_WriteVMCS(BASE_VCPU_CAP + vcpu, field, value);
-    if (ret.error != seL4_NoError) {
-        microkit_dbg_puts("microkit_x86_write_vmcs: error writing data\n");
-        microkit_internal_crash(ret.error);
+    /* Assumes that a PD would only have access to 1 VCPU object.
+     * These fields will be written to local variables rather than the actual VMCS:
+     * - RIP,
+     * - Primary Processor Based VM Execution Controls, and
+     * - VM Entry Interruption-Information.
+     * Because they will be passed to the kernel on `seL4_VMEnter()`,
+     * then the kernel will write them to the VMCS for us.
+     * Writing other VMCS fields will go to the real VMCS immediately via a syscall.
+     */
+    switch (field) {
+    case MICROKIT_VMX_GUEST_RIP:
+        microkit_x86_vcpu_state.rip = value;
+        break;
+    case MICROKIT_VMX_CONTROL_PRIMARY_PROCESSOR_CONTROLS:
+        microkit_x86_vcpu_state.prim_proc_ctl = value;
+        break;
+    case MICROKIT_VMX_CONTROL_ENTRY_INTERRUPTION_INFO:
+        microkit_x86_vcpu_state.irq_info = value;
+        break;
+    default: {
+        seL4_X86_VCPU_WriteVMCS_t ret = seL4_X86_VCPU_WriteVMCS(BASE_VCPU_CAP + vcpu, field, value);
+        if (ret.error != seL4_NoError) {
+            microkit_dbg_puts("microkit_x86_write_vmcs: error writing data\n");
+            microkit_internal_crash(ret.error);
+        }
+        break;
+    }
     }
 }
 
@@ -472,7 +542,20 @@ static inline void microkit_vcpu_x86_write_regs(microkit_child vcpu, seL4_VCPUCo
     }
 }
 
-#endif
+static inline void microkit_vcpu_x86_on(void)
+{
+    /* On x86, a TCB can only have one bound VCPU at any given time.
+     * So we don't take a `microkit_child vcpu` here. */
+    microkit_x86_vcpu_state.is_on = seL4_True;
+}
+
+static inline void microkit_vcpu_x86_off(void)
+{
+    microkit_x86_vcpu_state.is_on = seL4_False;
+}
+
+#endif /* CONFIG_VTX */
+#endif /* CONFIG_ARCH_X86_64 */
 
 static inline void microkit_deferred_notify(microkit_channel ch)
 {
--- a/libmicrokit/include/microkit.h
+++ b/libmicrokit/include/microkit.h
@@ -190,13 +190,8 @@ static inline void microkit_vcpu_restart(microkit_child vcpu, seL4_Word entry_po
 {
     seL4_Error err;
     seL4_UserContext ctxt = {0};
-#if defined(CONFIG_ARCH_AARCH64)
     ctxt.pc = entry_point;
-#elif defined(CONFIG_ARCH_X86_64)
-    ctxt.rip = entry_point;
-#else
-#error "unknown architecture for 'microkit_vcpu_restart'"
-#endif
+
     err = seL4_TCB_WriteRegisters(
               BASE_VM_TCB_CAP + vcpu,
               seL4_True,
@@ -220,9 +215,7 @@ static inline void microkit_vcpu_stop(microkit_child vcpu)
         microkit_internal_crash(err);
     }
 }
-#endif
 
-#if defined(CONFIG_ARM_HYPERVISOR_SUPPORT)
 static inline void microkit_vcpu_arm_inject_irq(microkit_child vcpu, seL4_Uint16 irq, seL4_Uint8 priority,
                                                 seL4_Uint8 group, seL4_Uint8 index)
 {
@@ -265,7 +258,7 @@ static inline void microkit_vcpu_arm_write_reg(microkit_child vcpu, seL4_Word re
         microkit_internal_crash(err);
     }
 }
-#endif
+#endif /* CONFIG_ARM_HYPERVISOR_SUPPORT */
 
 #if defined(CONFIG_ALLOW_SMC_CALLS)
 static inline void microkit_arm_smc_call(seL4_ARM_SMCContext *args, seL4_ARM_SMCContext *response)
@@ -277,7 +270,7 @@ static inline void microkit_arm_smc_call(seL4_ARM_SMCContext *args, seL4_ARM_SMC
         microkit_internal_crash(err);
     }
 }
-#endif
+#endif /* CONFIG_ALLOW_SMC_CALLS */
 
 #if defined(CONFIG_ARCH_X86_64)
 static inline void microkit_x86_ioport_write_8(microkit_ioport ioport_id, seL4_Word port_addr, seL4_Word data)
--- a/libmicrokit/src/main.c
+++ b/libmicrokit/src/main.c
@@ -17,6 +17,9 @@
 #define PD_MASK 0xff
 #define CHANNEL_MASK 0x3f
 
+#define BADGE_FAULT_BIT 62
+#define BADGE_ENDPOINT_BIT 63
+
 /* All globals are prefixed with microkit_* to avoid clashes with user defined globals. */
 
 bool microkit_passive;
@@ -26,6 +29,10 @@ char microkit_name[MICROKIT_PD_NAME_LENGTH];
 seL4_Bool microkit_have_signal = seL4_False;
 seL4_CPtr microkit_signal_cap;
 seL4_MessageInfo_t microkit_signal_msg;
+
+#if defined(CONFIG_VTX)
+struct microkit_x86_vcpu_state microkit_x86_vcpu_state;
+#endif /* CONFIG_VTX */
 
 seL4_Word microkit_irqs;
 seL4_Word microkit_notifications;
@@ -71,10 +78,49 @@ static void deferred_flush(void)
     }
 }
 
+#if defined(CONFIG_VTX)
+static seL4_MessageInfo_t x86_vcpu_resume(seL4_Word *badge)
+{
+    /* There is no seL4 invocation which combines a VMEnter and a non-blocking send.
+     * Thus, we must perform any deferred signals from `microkit_deferred_notify()` /
+     * `microkit_deferred_irq_ack()` before invoking VMEnter. */
+    deferred_flush();
+
+    seL4_Word is_fault, fault_reason;
+    struct microkit_x86_vcpu_state *s = &microkit_x86_vcpu_state;
+
+    x64_sys_send_recv(seL4_SysVMEnter, 0, badge, 0, &is_fault,
+                      &s->rip, &s->prim_proc_ctl, &s->irq_info, &fault_reason, 0);
+    /* We want to follow the documented kernel behaviour where these 4 values are
+     * always populated in the message registers. However, due to `setMR()`'s behaviour
+     * in include/object/tcb.h of the kernel, these 4 values will only be set in
+     * CPU registers. So we need to copy them to the IPC buffer to conform to the
+     * documented kernel behaviour.
+     *
+     * The other reason why we want to manage these values is so that the user does not
+     * have to worry about saving/updating them when servicing a notification. */
+    microkit_mr_set(SEL4_VMENTER_CALL_EIP_MR, s->rip);
+    microkit_mr_set(SEL4_VMENTER_CALL_CONTROL_PPC_MR, s->prim_proc_ctl);
+    microkit_mr_set(SEL4_VMENTER_CALL_INTERRUPT_INFO_MR, s->irq_info);
+    /* The 4th value (SEL4_VMENTER_FAULT_REASON_MR) is only valid when is_fault is true. */
+
+    /* Create a dummy msgInfo so that we can call `fault()`, as on x86 a VMExit is
+     * not an IPC as on other architectures. */
+    if (is_fault) {
+        microkit_mr_set(SEL4_VMENTER_FAULT_REASON_MR, fault_reason);
+        *badge |= 1ull << BADGE_FAULT_BIT;
+        return seL4_MessageInfo_new(0, 0, 0, SEL4_VMENTER_NUM_FAULT_MSGS);
+    } else {
+        /* VCPU got interrupted due to a notification, no msgInfo. */
+        return seL4_MessageInfo_new(0, 0, 0, 0);
+    }
+}
+#endif
+
 static void handler_loop(void)
 {
     bool have_reply = false;
-    seL4_MessageInfo_t reply_tag;
+    seL4_MessageInfo_t reply_tag = seL4_MessageInfo_new(0, 0, 0, 0);
 
     /**
      * Because of https://github.com/seL4/seL4/issues/1536
@@ -97,7 +143,16 @@ static void handler_loop(void)
         seL4_Word badge;
         seL4_MessageInfo_t tag;
 
+#if defined(CONFIG_VTX)
+        if (microkit_x86_vcpu_state.is_on) {
+            /* We should never have a reply message from the `protected()` endpoint,
+            * as on x86 a PD with a bound vCPU cannot receive PPCs.*/
+            // assert(!have_reply);
+            tag = x86_vcpu_resume(&badge);
+        } else if (have_reply) {
+#else
         if (have_reply) {
+#endif
             deferred_flush();
             tag = seL4_ReplyRecv(INPUT_CAP, reply_tag, &badge, REPLY_CAP);
         } else if (microkit_have_signal) {
@@ -107,13 +162,21 @@ static void handler_loop(void)
             tag = seL4_Recv(INPUT_CAP, &badge, REPLY_CAP);
         }
 
-        uint64_t is_endpoint = badge >> 63;
-        uint64_t is_fault = (badge >> 62) & 1;
+        uint64_t is_endpoint = badge >> BADGE_ENDPOINT_BIT;
+        uint64_t is_fault = (badge >> BADGE_FAULT_BIT) & 1;
 
         have_reply = false;
 
         if (is_fault) {
             seL4_Bool reply_to_fault = fault(badge & PD_MASK, tag, &reply_tag);
+#if defined(CONFIG_VTX)
+            /* If fault() returns false then we shouldn't resume the VCPU. */
+            if (!reply_to_fault) {
+                microkit_x86_vcpu_state.is_on = seL4_False;
+            }
+            /* There won't be anything to reply to for a VCPU fault. */
+            reply_to_fault = seL4_False;
+#endif
             if (reply_to_fault) {
                 have_reply = true;
             }
--- a/tool/microkit/src/sdf.rs
+++ b/tool/microkit/src/sdf.rs
@@ -2071,6 +2071,18 @@ pub fn parse(
             }
             vms.push(&vm.name);
         }
+
+        if config.arch == Arch::X86_64 && pd.virtual_machine.is_some() && pd.has_children {
+            // When seL4_VMEnter() is called, the kernel only checks the VMM's bound
+            // notification for pending signals. Because the endpoint object isn't passed
+            // or checked for pending messages, Child PDs won't work while the VCPU is on.
+            // Technically, Child PDs could still work when the VCPU is off, but we shouldn't
+            // expose this footgun to users.
+            return Err(format!(
+                    "Error: It is not possible for PD '{}' with a bound vCPU to have children on x86_64: {}",
+                    pd.name(),
+                    loc_string(&xml_sdf, pd.text_pos.unwrap())));
+        }
     }
 
     // Ensure no duplicate IRQs
@@ -2148,6 +2160,21 @@ pub fn parse(
                 "Error: PPCs must be to protection domains of strictly higher priorities; \
                         channel with PPC exists from pd {} (priority: {}) to pd {} (priority: {})",
                 pd_b.name, pd_b.priority, pd_a.name, pd_a.priority
+            ));
+        }
+
+        if config.arch == Arch::X86_64
+            && ((ch.end_a.pp && pd_b.virtual_machine.is_some())
+                || (ch.end_b.pp && pd_a.virtual_machine.is_some()))
+        {
+            // Same cause as child PD above
+            return Err(format!(
+                "Error: It is not possible to PPC to PD '{}' with a bound vCPU from PD '{}' on x86_64 @ {}:{}:{}",
+                    if ch.end_a.pp {&pd_b.name} else {&pd_a.name},
+                    if ch.end_a.pp {&pd_a.name} else {&pd_b.name},
+                    filename.display(),
+                    pd_a.text_pos.unwrap().row,
+                    pd_a.text_pos.unwrap().col
             ));
         }
 
--- a/tool/microkit/tests/sdf/vm_x86_ppc_invalid_1.system
+++ b/tool/microkit/tests/sdf/vm_x86_ppc_invalid_1.system
@@ -1,0 +1,39 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+ Copyright 2026, UNSW
+ SPDX-License-Identifier: BSD-2-Clause
+-->
+<system>
+    <memory_region name="hpet_regs" size="0x1000" phys_addr="0xfed00000" />
+    <memory_region name="guest_ram" size="0x10_000_000" page_size="0x200_000" />
+
+    <protection_domain name="VMM" priority="100" stack_size="0x2000">
+        <program_image path="vmm_x86_64.elf" />
+        <map mr="guest_ram" vaddr="0x80000000" perms="rw" setvar_vaddr="guest_ram_vaddr" setvar_size="guest_ram_size" />
+
+        <virtual_machine name="linux">
+            <vcpu id="0" />
+            <map mr="guest_ram" vaddr="0x0" perms="rwx" cached="true" />
+        </virtual_machine>
+    </protection_domain>
+
+    <protection_domain name="timer_driver" priority="254" passive="true">
+        <program_image path="timer_driver.elf" />
+        <map mr="hpet_regs" vaddr="0x50000000" perms="rw" cached="false" />
+        <irq pin="2" vector="107" id="0" ioapic="0" trigger="edge" />
+    </protection_domain>
+
+    <protection_domain name="some_service" priority="1" >
+        <program_image path="some_service.elf" />
+    </protection_domain>
+
+    <channel>
+        <end pd="timer_driver" id="1" />
+        <end pd="VMM" id="11" pp="true" />
+    </channel>
+
+    <channel>
+        <end pd="some_service" id="1" pp="true" />
+        <end pd="VMM" id="12" />
+    </channel>
+</system>
--- a/.linkcheck-ignore.yml
+++ b/.linkcheck-ignore.yml
@@ -1,2 +1,3 @@
 urls:
   - "https://developer.arm.com/downloads/*"
+  - "https://cdrdv2.intel.com/*"
--- a/tool/microkit/tests/sdf/vm_x86_ppc_invalid_2.system
+++ b/tool/microkit/tests/sdf/vm_x86_ppc_invalid_2.system
@@ -1,0 +1,39 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+ Copyright 2026, UNSW
+ SPDX-License-Identifier: BSD-2-Clause
+-->
+<system>
+    <memory_region name="hpet_regs" size="0x1000" phys_addr="0xfed00000" />
+    <memory_region name="guest_ram" size="0x10_000_000" page_size="0x200_000" />
+
+    <protection_domain name="VMM" priority="100" stack_size="0x2000">
+        <program_image path="vmm_x86_64.elf" />
+        <map mr="guest_ram" vaddr="0x80000000" perms="rw" setvar_vaddr="guest_ram_vaddr" setvar_size="guest_ram_size" />
+
+        <virtual_machine name="linux">
+            <vcpu id="0" />
+            <map mr="guest_ram" vaddr="0x0" perms="rwx" cached="true" />
+        </virtual_machine>
+    </protection_domain>
+
+    <protection_domain name="timer_driver" priority="254" passive="true">
+        <program_image path="timer_driver.elf" />
+        <map mr="hpet_regs" vaddr="0x50000000" perms="rw" cached="false" />
+        <irq pin="2" vector="107" id="0" ioapic="0" trigger="edge" />
+    </protection_domain>
+
+    <protection_domain name="some_service" priority="1" >
+        <program_image path="some_service.elf" />
+    </protection_domain>
+
+    <channel>
+        <end pd="timer_driver" id="1" />
+        <end pd="VMM" id="11" pp="true" />
+    </channel>
+
+    <channel>
+        <end pd="VMM" id="12" />
+        <end pd="some_service" id="1" pp="true" />
+    </channel>
+</system>
--- a/tool/microkit/tests/sdf/vm_x86_ppc_valid.system
+++ b/tool/microkit/tests/sdf/vm_x86_ppc_valid.system
@@ -1,0 +1,30 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+ Copyright 2026, UNSW
+ SPDX-License-Identifier: BSD-2-Clause
+-->
+<system>
+    <memory_region name="hpet_regs" size="0x1000" phys_addr="0xfed00000" />
+    <memory_region name="guest_ram" size="0x10_000_000" page_size="0x200_000" />
+
+    <protection_domain name="VMM" priority="0" stack_size="0x2000">
+        <program_image path="vmm_x86_64.elf" />
+        <map mr="guest_ram" vaddr="0x80000000" perms="rw" setvar_vaddr="guest_ram_vaddr" setvar_size="guest_ram_size" />
+
+        <virtual_machine name="linux">
+            <vcpu id="0" />
+            <map mr="guest_ram" vaddr="0x0" perms="rwx" cached="true" />
+        </virtual_machine>
+    </protection_domain>
+
+    <protection_domain name="timer_driver" priority="254" passive="true">
+        <program_image path="timer_driver.elf" />
+        <map mr="hpet_regs" vaddr="0x50000000" perms="rw" cached="false" />
+        <irq pin="2" vector="107" id="0" ioapic="0" trigger="edge" />
+    </protection_domain>
+
+    <channel>
+        <end pd="timer_driver" id="1" />
+        <end pd="VMM" id="11" notify="false" pp="true" />
+    </channel>
+</system>
--- a/tool/microkit/tests/sdf/vm_x86_vm_with_child_invalid.system
+++ b/tool/microkit/tests/sdf/vm_x86_vm_with_child_invalid.system
@@ -1,0 +1,20 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+ Copyright 2026, UNSW
+ SPDX-License-Identifier: BSD-2-Clause
+-->
+<system>
+    <protection_domain name="VMM" priority="100" stack_size="0x2000">
+        <program_image path="vmm_x86_64.elf" />
+        <map mr="guest_ram" vaddr="0x80000000" perms="rw" setvar_vaddr="guest_ram_vaddr" setvar_size="guest_ram_size" />
+
+        <virtual_machine name="linux">
+            <vcpu id="0" />
+            <map mr="guest_ram" vaddr="0x0" perms="rwx" cached="true" />
+        </virtual_machine>
+
+        <protection_domain id="1" name="children_service" priority="1" >
+            <program_image path="children_service.elf" />
+        </protection_domain>
+    </protection_domain>
+</system>
--- a/tool/microkit/tests/test.rs
+++ b/tool/microkit/tests/test.rs
@@ -761,6 +761,38 @@ mod virtual_machine {
             "Error: invalid memory region name 'mr1' on 'map' @",
         )
     }
+
+    #[test]
+    fn test_x86_ppc_invalid_1() {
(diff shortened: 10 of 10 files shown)


midnightveil:
For the record: I'm not entirely convinced this is the best way the API could look/behave, it feels quite ugly.

dreamliner787-9:
> For the record: I'm not entirely convinced this is the best way the API could look/behave, it feels quite ugly.

Yes I agree, as discussed internally, there isn't a clean way to integrate the way virtualisation works on x86 seL4 with the Microkit event loop. Unless we allow the user to override the `main()` entrypoint and let the user call into the library in some way...

Indanz:
> > For the record: I'm not entirely convinced this is the best way the API could look/behave, it feels quite ugly.
> 
> Yes I agree, as discussed internally, there isn't a clean way to integrate the way virtualisation works on x86 seL4 with the Microkit event loop. 

Why can't the Microkit event loop be x86 virtualisation aware? Or have a x86 version of the event loop that calls VMEnter instead of recv. You can add a special function that enters this x86 virtualisation aware event loop and never returns. Then the user only needs to do one special call during init once. If the tooling is aware of all this, it could do this automatically on return of the init function too and the user wouldn't need to do anything special (other than being aware that a VCPU task runs both normal code and a VM).

dreamliner787-9:
> > > For the record: I'm not entirely convinced this is the best way the API could look/behave, it feels quite ugly.
> > 
> > 
> > Yes I agree, as discussed internally, there isn't a clean way to integrate the way virtualisation works on x86 seL4 with the Microkit event loop.
> 
> Why can't the Microkit event loop be x86 virtualisation aware? Or have a x86 version of the event loop that calls VMEnter instead of recv. You can add a special function that enters this x86 virtualisation aware event loop and never returns. Then the user only needs to do one special call during init once.

I've done some thinking on what you've said and I like it a lot. My take on your idea is that we will have one API that looks like:
```c
static void microkit_x86_vcpu_on(uint64_t rip, uint64_t prim_proc_ctl, uint32_t irq_info)
```

The user calls this before returning from `init()` or `notified()` once during the PD's lifetime. This switches the normal event loop into a x86 virtualisation aware mode. Then the Microkit will automatically resume the VCPU as part of the event loop. No deferred calls like right now.

For argument passing to `seL4_VMEnter()`, that will be initially taken care of with this call. For subsequent VM exits due to notifications or faults, we can easily hook it to the existing read/write VMCS API. That is, any writes to the guest RIP, Primary Processor Controls or IRQ info get written to local variables, then it will be written to message registers right before the event loop calls `seL4_VMEnter()`. Rather than having to do a read/write VMCS syscall.

Under this model, the user just need to be aware that a PD with a VCPU will be executing guest code if `init()` or `notified()` returns and the PD get scheduled.

Edit: it maybe beneficial to have a `microkit_x86_vcpu_off` as well. So that this model is similar to the ARM's VCPU restart and stop APIs.

> If the tooling is aware of all this, it could do this automatically on return of the init function too and the user wouldn't need to do anything special (other than being aware that a VCPU task runs both normal code and a VM).

This would be ideal, but VCPUs are resumed by user code at runtime so I'm unsure how this could fit in.

Indanz:
> > If the tooling is aware of all this, it could do this automatically on return of the init function too and the user wouldn't need to do anything special (other than being aware that a VCPU task runs both normal code and a VM).
> 
> This would be ideal, but VCPUs are resumed by user code at runtime so I'm unsure how this could fit in.

Resuming a VCPU is just calling `seL4_VMEnter()`. That is no different from a code flow point of view than blocking on `seL4_ReplyRecv()`. Either can be called after return of the user callback function.

Considering x86 VCPU tasks can't receive IPC calls anyway, but only notifications, for Microkit it would make sense to model non-notification returns from VMEnter as Protected procedure calls. Or be more specific and add another callback function for those.

(I'm ignoring the practical implementation split between generic Microkit and libvm or whatever virtualisation aware code you have. For a more virtualisation agnostic Microkit it would help to move stuff more into libraries and less in the Microkit framework.)

libmicrokit/include/microkit.h
```diff
@@ -391,28 +393,74 @@ static inline seL4_Uint32 microkit_x86_ioport_read_32(microkit_ioport ioport_id,
 
     return ret.result;
 }
-#endif
 
-#if defined(CONFIG_ARCH_X86_64) && defined(CONFIG_VTX)
+#if defined(CONFIG_VTX)
+/* These value came from seL4 source, see: include/arch/x86/arch/object/vcpu.h */
```
Indanz: These are the identifiers for specific MSRs in VMCS, so this comment is misleading.

libmicrokit/include/microkit.h
```diff
@@ -391,28 +393,74 @@ static inline seL4_Uint32 microkit_x86_ioport_read_32(microkit_ioport ioport_id,
...
+        break;
+    case VMX_CONTROL_ENTRY_INTERRUPTION_INFO:
+        value = microkit_x86_vcpu_irq_info;
+        break;
+    default: {
+        seL4_X86_VCPU_ReadVMCS_t ret;
+        ret = seL4_X86_VCPU_ReadVMCS(BASE_VCPU_CAP + vcpu, field);
+        if (ret.error != seL4_NoError) {
+            microkit_dbg_puts("microkit_x86_read_vmcs: error reading data\n");
+            microkit_internal_crash(ret.error);
```
Indanz, line 442: `field` is user supplied, it is not an internal bug if this syscall fails, but a user error. You should propagate those errors to user space instead of deliberately crashing. (And if you do have such kind of assertions, use actual assertions so they can be disabled at compile time with -DNDEBUG.)

libmicrokit/src/main.c
```diff
@@ -63,10 +73,98 @@ static void run_init_funcs(void)
     }
 }
 
+static seL4_MessageInfo_t receive_event(bool have_reply, seL4_MessageInfo_t reply_tag, seL4_Word *badge)
+{
+    if (have_reply) {
+        return seL4_ReplyRecv(INPUT_CAP, reply_tag, badge, REPLY_CAP);
+    } else if (microkit_have_signal) {
+        return seL4_NBSendRecv(microkit_signal_cap, microkit_signal_msg, INPUT_CAP, badge, REPLY_CAP);
+        microkit_have_signal = seL4_False;
```
Indanz: Well, that's not going to work.

libmicrokit/src/main.c
```diff
@@ -63,10 +73,98 @@ static void run_init_funcs(void)
...
+static seL4_Bool is_fault(seL4_Word badge)
+{
+#if defined(CONFIG_VTX)
+    if (microkit_x86_vmenter_result_valid && microkit_x86_vmenter_result == SEL4_VMENTER_RESULT_FAULT) {
+        return seL4_True;
+    } else {
+        return !!((badge >> 62) & 1);
+    }
+#else
+    return !!((badge >> 62) & 1);
```
Indanz: Please combine this with the previous identical check.

Also, doing !! only makes sense if you check `badge & (1 << 62)` (which would make sense if you had a define for 1 << 62. Why don't you?). If you shift it to bit 0 anyway, there is no need for !! any more.

libmicrokit/src/main.c
```diff
@@ -63,10 +73,98 @@ static void run_init_funcs(void)
...
+    }
+#else
+    return receive_event(have_reply, reply_tag, badge);
+#endif
+}
+
+static seL4_Bool is_endpoint(seL4_Word badge)
+{
+    return !!(badge >> 63);
+}
```
Indanz: This should be inline and why the !!?

And consider moving them to a header file so users can also use them perhaps?

libmicrokit/src/main.c
```diff
@@ -63,10 +73,98 @@ static void run_init_funcs(void)
...
+static seL4_MessageInfo_t send_recv_event(bool have_reply, seL4_MessageInfo_t reply_tag, seL4_Word *badge)
+{
+#if defined(CONFIG_VTX)
+    microkit_x86_vmenter_result_valid = seL4_False;
+    if (microkit_x86_do_vcpu_resume) {
+        /* There isn't a syscall that atomically send signal then perform vmenter while waiting for
+            incoming notifications. So we have to dispatch any deferred signal first. Then switch execution
+            to the bound VCPU. A PD with a VCPU can't receive PPC on x86 so no need to check `have_reply`. */
+        if (microkit_have_signal) {
+            seL4_NBSend(microkit_signal_cap, microkit_signal_msg);
```
Indanz: Why not `seL4_Signal`?

libmicrokit/src/main.c
```diff
@@ -63,10 +73,98 @@ static void run_init_funcs(void)
...
+        return seL4_ReplyRecv(INPUT_CAP, reply_tag, badge, REPLY_CAP);
+    } else if (microkit_have_signal) {
+        return seL4_NBSendRecv(microkit_signal_cap, microkit_signal_msg, INPUT_CAP, badge, REPLY_CAP);
+        microkit_have_signal = seL4_False;
+    } else {
+        return seL4_Recv(INPUT_CAP, badge, REPLY_CAP);
+    }
+}
+
+static seL4_MessageInfo_t send_recv_event(bool have_reply, seL4_MessageInfo_t reply_tag, seL4_Word *badge)
```
Indanz: `send_recv_event()` and `receive_event()` are inside out: The VTX part should go in its own function and that should be called as one of the if-checks in what currently is `receive_event()`. No need to add layers of layers of indirection just for the sake of it.

libmicrokit/src/main.c
```diff
@@ -63,10 +73,98 @@ static void run_init_funcs(void)
...
+    seL4_Bool reply_to_fault = fault(badge & PD_MASK, tag, reply_tag);
+    if (microkit_x86_vmenter_result_valid == seL4_True) {
+        /* There won't be anything to reply to for a VCPU fault. But
+         * if fault() returns false then we shouldn't resume the VCPU. */
+        if (!reply_to_fault) {
+            microkit_x86_do_vcpu_resume = seL4_False;
+        }
+        reply_to_fault = seL4_False;
+    }
+    return reply_to_fault;
```
Indanz: This is horrible and doesn't make sense, please remove this mess. It's harmless to do a `seL4_ReplyRecv` while there is no outstanding call, even if `fault` would tell us to do that for some reason.

And no need to set `microkit_x86_do_vcpu_resume` to false, the user can turn the vcpu off or reset it in the fault handler. Doing things like this behind the user's back is bad design.

tool/microkit/src/sdf.rs
```diff
@@ -2018,6 +2018,26 @@ pub fn parse(
             ));
         }
 
+        // Due to a kernel limitation in `src/arch/x86/c_traps.c`,
+        // only the VMM's bound ntfn is checked for pending ntfn.
```
Indanz: What limitation? For non-VMM only the bound ntfn is also checked, so there is nothing different in that regard.

tool/microkit/src/sdf.rs
```diff
@@ -2018,6 +2018,26 @@ pub fn parse(
...
+        // The endpoint object isn't passed to the kernel and checked for
+        // pending messages so PPC won't work while the VCPU is running.
+        // Technically PPC still works if the VCPU isn't resumed. But
```
Indanz:
```suggestion
        // Technically PPC could still work when the VCPU is off.
```

libmicrokit/include/microkit.h
midnightveil: Yeah, I really wish we would remove all these internal crashes  and just propagate the error.

Indanz: And the same applies for debug prints, for a library that should be a user decision too.

libmicrokit/src/main.c
midnightveil: The layout of the badge bits is an internal representation (at this time). 

Also, maybe not in this context, but compilers can warn about implicit int from bool conversions. Also, stylistically I prefer it.

libmicrokit/src/main.c
midnightveil: (Yes I don't agree with the addition of these functions - my opinion previously stated internally was that the non-x86 code should be kept basically the same...)

libmicrokit/src/main.c
midnightveil: I believe the idea is to match that of ARM: that is that if you don't reply thr VCPU won't start again (in ARM's case because of blocked on fault-ness).

Indanz: Well, in Arm's case you can write all the registers via the reply message too, and you're not emulating that either. It's confusing if you try to emulate part of the behaviour. I can't find anything documenting the meaning of the `fault` function's return value.

tool/microkit/src/sdf.rs
dreamliner787-9: I should have worded that better:
```diff
- // only the VMM's bound ntfn is checked for pending ntfn.
+ // only the VMM's bound ntfn is checked for pending ntfn when seL4_VMEnter() is called.
```

> For non-VMM only the bound ntfn is also checked, so there is nothing different in that regard.

For non x86 VMM, libmicrokit calls `seL4_Recv()` or one of its flavours and pass in the PD's endpoint cap. For x86 VMM, only the notification cap can be passed to `seL4_VMEnter()`.

libmicrokit/src/main.c
dreamliner787-9: Oops that's right will fix.

libmicrokit/src/main.c
dreamliner787-9: I was just trying to make it consistent with how the mainline event loop works, but `seL4_Signal` works as well, will fix.

libmicrokit/src/main.c
dreamliner787-9: > I believe the idea is to match that of ARM: that is that if you don't reply thr VCPU won't start again

Yes that was the intention, the behaviour is documented in the Microkit Manual:
> ## `seL4_Bool fault(microkit_child child, microkit_msginfo msginfo, microkit_msginfo *reply_msginfo)`
> ...
> The returned `seL4_Bool` is whether or not to reply to the fault with the message `reply_msginfo`.
Returning `seL4_True` will reply to the fault. Returning `seL4_False` will not reply to the fault.

libmicrokit/src/main.c
dreamliner787-9: Hmm I must have misunderstood the discussion offline, will fix.

libmicrokit/src/main.c
Indanz: Ah, I didn't realise I could scroll down in the navigation panel. I looked at the [faults](https://docs.sel4.systems/projects/microkit/manual/latest/#fault) section and did some searches here and there, but missed that whole section.

tool/microkit/src/sdf.rs
Indanz: So the limitation is you can't handle endpoints while running the VMM. Nothing to do with `src/arch/x86/c_traps.c` or notifications though. Confusing comment.

libmicrokit/src/main.c
dreamliner787-9: I believe this have been improved now.

libmicrokit/include/microkit.h
dreamliner787-9: I feel like this issue is out of this PR's scope and should be revisited in another PR later.

libmicrokit/include/microkit.h
```diff
@@ -188,13 +197,8 @@ static inline void microkit_vcpu_restart(microkit_child vcpu, seL4_Word entry_po
...
     seL4_Error err;
     seL4_UserContext ctxt = {0};
-#if defined(CONFIG_ARCH_AARCH64)
     ctxt.pc = entry_point;
-#elif defined(CONFIG_ARCH_X86_64)
-    ctxt.rip = entry_point;
-#else
-#error "unknown architecture for 'microkit_vcpu_restart'"
-#endif
+
```
Indanz, line 206: Is this intentional? `microkit_pd_restart()` has the same.

Looking at `libsel4/sel4_arch_include/x86_64/interfaces/object-api-sel4-arch.xml`, there isn't an entry `pc` generated for x86_64?

libmicrokit/include/microkit.h
```diff
@@ -391,28 +393,81 @@ static inline seL4_Uint32 microkit_x86_ioport_read_32(microkit_ioport ioport_id,
...
+     * - RIP,
+     * - Primary Processor Based VM Execution Controls, and
+     * - VM Entry Interruption-Information
+     * from local variables. Since they will be passed to the kernel on `seL4_VMEnter()`,
+     * then the kernel will write them to the VMCS for us.
+     *
+     * Note that this logic will be incorrect if we give a single PD access to multiple VCPU caps,
+     * but I don't think that will be a valid scenario since you can only have 1 bound VCPU
+     * with your TCB. */
+
```
Indanz: - This comment should be outside the function.
- The part "Since they will be passed to the kernel on `seL4_VMEnter()`, then the kernel will write them to the VMCS for us." applies to writing VMCS, not for reading.
- The paragraph comment doesn't belong here.

libmicrokit/include/microkit.h
```diff
@@ -391,28 +393,81 @@ static inline seL4_Uint32 microkit_x86_ioport_read_32(microkit_ioport ioport_id,
...
+        break;
+    case VMX_CONTROL_PRIMARY_PROCESSOR_CONTROLS:
+        value = microkit_x86_vcpu_prim_proc_ctl;
+        break;
+    case VMX_CONTROL_ENTRY_INTERRUPTION_INFO:
+        value = microkit_x86_vcpu_irq_info;
+        break;
+    default: {
+        seL4_X86_VCPU_ReadVMCS_t ret;
+        ret = seL4_X86_VCPU_ReadVMCS(BASE_VCPU_CAP + vcpu, field);
```
Indanz: Either put an empty new line between them, or put them on one line.

libmicrokit/include/microkit.h
```diff
@@ -477,7 +532,20 @@ static inline void microkit_vcpu_x86_write_regs(microkit_child vcpu, seL4_VCPUCo
     }
 }
 
-#endif
+static inline void microkit_vcpu_x86_on(void)
+{
+    /* On x86, a TCB can only have one bound VCPU at any given time.
+     * So we don't take a `microkit_child vcpu` here. */
```
Indanz, line 548: Doesn't that mean that you want to take that argument away from `microkit_vcpu_x86_write_vmcs/microkit_vcpu_x86_read_vmcs` too? (And any other x86 virtualisation functions there may be.)

libmicrokit/include/microkit.h
```diff
@@ -45,6 +48,12 @@ extern char microkit_name[MICROKIT_PD_NAME_LENGTH];
 extern seL4_Bool microkit_have_signal;
 extern seL4_CPtr microkit_signal_cap;
 extern seL4_MessageInfo_t microkit_signal_msg;
+#if defined(CONFIG_VTX)
+extern seL4_Bool microkit_x86_do_vcpu_resume;
+extern seL4_Word microkit_x86_vcpu_rip;
+extern seL4_Word microkit_x86_vcpu_prim_proc_ctl;
+extern seL4_Word microkit_x86_vcpu_irq_info;
```
Indanz: Can't you at least group them together in a structure?

I would do that for all those global Microkit variables and put them all in one structure (and the vcpu ones in a substructure of that one). Then it's also easier to hide implementation details from the user, as you can do a forward declaration of the structure and hide the content.

libmicrokit/src/main.c
```diff
@@ -63,10 +73,94 @@ static void run_init_funcs(void)
...
 
+#if defined(CONFIG_VTX)
+static seL4_MessageInfo_t x86_vcpu_resume(seL4_Word *badge)
+{
+    /* There isn't a seL4 syscall that atomically send signal then perform vmenter while waiting for
+        incoming notifications. So we have to dispatch any deferred signal first. Then switch execution
+        to the bound VCPU. A PD with a VCPU can't receive PPC on x86 so no need to check `have_reply`. */
+    if (microkit_have_signal) {
+        seL4_Send(microkit_signal_cap, microkit_signal_msg);
+        microkit_have_signal = seL4_False;
```
Indanz: Assuming that `microkit_have_signal` is only set by a user calling `microkit_deferred_notify`, I wouldn't bother supporting this corner case. It's okay to detect and warn it, but I wouldn't actually handle it, as it makes no sense.

And what about `microkit_deferred_irq_ack`?

If you do want to deal with this, add a `microkit_deferred_flush()` function which clears out all the deferred stuff and call that here.

libmicrokit/src/main.c
```diff
@@ -63,10 +73,94 @@ static void run_init_funcs(void)
...
+
+static inline bool is_endpoint(seL4_Word badge)
+{
+    return badge >> 63;
+}
+
+static bool is_fault(seL4_Word badge)
+{
+#if defined(CONFIG_VTX)
+    if (microkit_x86_vmenter_result_valid && microkit_x86_vmenter_result == SEL4_VMENTER_RESULT_FAULT) {
```
Indanz: Why not just check `SEL4_VMENTER_RESULT_FAULT`?

libmicrokit/src/main.c
```diff
@@ -63,10 +73,94 @@ static void run_init_funcs(void)
...
+    }
+#endif
+    return (badge >> 62) & 1;
+}
+
+static seL4_Bool handle_fault(seL4_Word badge, seL4_MessageInfo_t tag, seL4_MessageInfo_t *reply_tag)
+{
+#if defined(CONFIG_VTX)
+    seL4_Bool reply_to_fault = fault(badge & PD_MASK, tag, reply_tag);
+    if (microkit_x86_vmenter_result_valid == seL4_True) {
```
Indanz: Why not just check `SEL4_VMENTER_RESULT_FAULT` and get rid of `microkit_x86_vmenter_result_valid`?

libmicrokit/src/main.c
```diff
@@ -63,10 +73,94 @@ static void run_init_funcs(void)
...
+        /* There won't be anything to reply to for a VCPU fault. But
+         * if fault() returns false then we shouldn't resume the VCPU. */
+        if (!reply_to_fault) {
+            microkit_x86_do_vcpu_resume = seL4_False;
+        }
+        reply_to_fault = seL4_False;
+    }
+    return reply_to_fault;
+#else
+    return fault(badge & PD_MASK, tag, reply_tag);
```
Indanz: Please just return `reply_to_fault` instead of duplicating the call.

docs/manual.md
```diff
@@ -984,6 +986,18 @@ virtual CPU with ID `vcpu`.
 Write the registers of a given virtual CPU with ID `vcpu`. The `regs` argument is the pointer to
 the struct of registers `seL4_VCPUContext` that are written from.
 
+## `void microkit_vcpu_x86_on(void)`
+
+Allow the PD to switch to guest execution mode with the bound vCPU every time `init()` or `notified()` return.
```
Indanz: But the main path seems to be the `fault` handler, not those other ones. You should probably mention that here, as it's not obvious. I think it would be clearer to have a special callback for x86 vmexit returns, instead of using the fault one.

docs/manual.md
```diff
@@ -984,6 +986,18 @@ virtual CPU with ID `vcpu`.
...
+
+The caller is responsible for initialising the VCPU's instruction pointer, Primary Processor-Based VM-Execution Controls
+and VM-Entry Interruption-Information Field to the corresponding VMCS fields. For more details,
+please see the 'Virtualisation' section of the [seL4 reference manual](https://sel4.systems/Info/Docs/seL4-manual-latest.pdf).
```
Indanz: yes, but say how, as it's not obvious you are intercepting the vmcs read/write functions.

docs/manual.md
```diff
@@ -649,6 +649,8 @@ To find the full list of possible faults that could occur and details regarding
 kind of fault, please see the 'Faults' section of the
 [seL4 reference manual](https://sel4.systems/Info/Docs/seL4-manual-latest.pdf).
 
+For x86 VCPU faults, please see the 'Virtualisation' section of the [seL4 reference manual](https://sel4.systems/Info/Docs/seL4-manual-latest.pdf).
```
Indanz: The seL4 manual doesn't mention any x86 VCPU faults.

docs/manual.md
```diff
@@ -1154,6 +1168,7 @@ The `end` element has the following attributes:
 * `id`: Channel identifier in the context of the named protection domain. Must be at least 0 and less than 63.
 * `pp`: (optional) Indicates that the protection domain for this end can perform a protected procedure call to the other end; defaults to false.
         Protected procedure calls can only be to PDs of strictly higher priority.
+        On x86-64, PDs with virtual machines cannot receive protected procedure calls.
```
Indanz, line 1089: Probably worth mentioning somewhere that there is a limit of one VCPU per PD for x86.

libmicrokit/src/main.c
```diff
@@ -27,6 +27,16 @@ seL4_Bool microkit_have_signal = seL4_False;
...
 seL4_MessageInfo_t microkit_signal_msg;
 
+#if defined(CONFIG_VTX)
+seL4_Bool microkit_x86_do_vcpu_resume = seL4_False;
+seL4_Word microkit_x86_vcpu_rip;
+seL4_Word microkit_x86_vcpu_prim_proc_ctl;
+seL4_Word microkit_x86_vcpu_irq_info;
+
+seL4_Bool microkit_x86_vmenter_result_valid = seL4_False;
+seL4_Word microkit_x86_vmenter_result;
```
Indanz:
```suggestion
static seL4_Bool microkit_x86_vmenter_result_valid = seL4_False;
static seL4_Word microkit_x86_vmenter_result;
```

libmicrokit/include/microkit.h
dreamliner787-9: Yes the removals are intentional, it's from a mistake introduced in https://github.com/seL4/microkit/pull/337.

On x86 a VCPU does not have its own TCB, and this API is only available when `CONFIG_ARM_HYPERVISOR_SUPPORT` so having x86 path in there is wrong.

dreamliner787-9: Arguably, `microkit_vcpu_restart` and `microkit_vcpu_stop` should be renamed to `microkit_vcpu_arm_restart` and `microkit_vcpu_arm_stop` but that would be a breaking change...

libmicrokit/include/microkit.h
dreamliner787-9: That's a good point, the reason why I haven't done that is because I don't know how VM with multiple VCPUs will look like yet, because there are multiple approaches you can take.

You could allow one VMM PD access to multiple VCPU caps to configure them and/or handle faults, i.e. you will have an implicit "big VMM lock". This is currently how ARM VM are modelled in Microkit. For x86, we will need additional trampoline threads that resume the VCPUs. The downside is that you have more context switching, but you would be able to replicate the ARM model. In this case we would need to pass `microkit_child vcpu` to all the VCPU APIs.

Or, you could adopt a "multi VMM" approach where each PD only have access 1 VCPU. Then the user is responsible for creating 1 PD per VCPU, handle IPIs with explicit channels between the PDs and synchronise access to any global VMM state themselves. This model would incur less context switching and there isn't a "big lock" like the previous one. But there are more complications: you would need to ensure that all the VMM PDs have access to the same resources (memory regions, interrupts, etc) for fault handling, and channels between the VMM PDs for every possible combinations of IPIs (`n(n-1)` channels). In this case we would not need to pass `microkit_child vcpu` to all the VCPU APIs.

Arguably the "multi VMM" approach is already possible on both ARM and x86 in Microkit. But at this point I'm not sure whether supporting the "big VMM lock" model on x86 is appropriate so I'm keeping the option open.

libmicrokit/include/microkit.h
dreamliner787-9: I will group the x86 VCPU ones in a structure, then we can do the other ones in a different PR.

libmicrokit/src/main.c
dreamliner787-9: We should deal with this case, since it is valid for the VMM to signal another component to service some request from the guest then resume the VCPU and let that happen in the background.

I don't see how a `microkit_deferred_flush()` would help though. For non VTX cases, "flushing deferred" then waiting for the next event happens within 1 syscall. So having the highlighted change is necessary for the VTX case since we can't refactor it with the non VTX cases.

docs/manual.md
dreamliner787-9: Oops yes that slipped my mind.

libmicrokit/include/microkit.h
Indanz: What you say argues against dropping the parameter for microkit_vcpu_x86_on/off too though.

I strongly advise against some trampoline construction that shuffles all VCPU handling to one VMM task: The reason to have multiple VCPUs is to use multiple cores, so they all will be on different cores. If you handle all VCPUs with one VMM task, all operations become cross-core with huge overhead: First the fault/trap needs to be propagated to the core the VMM is on, and then any reply needs to be funnelled back.

The added value of handling multiple, independent virtual machines with one VMM is approximately zero.

This means that also on Arm, you do not want to use the model with one central VMM handling multiple VCPUs, even though it's simpler.

> But there are more complications: you would need to ensure that all the VMM PDs have access to the same resources (memory regions, interrupts, etc) for fault handling,

Why? A VMM doesn't need access to the virtual machine's memory, interrupts etc.

> and channels between the VMM PDs for every possible combinations of IPIs (n(n-1) channels). 

Why do you need IPIs? Why would one VMM care about what others do?

If you need some kind of central control, then have a global VMM manager which communicates with all the VMMs handling VCPUs. Then you need only n channels and have only one decision point.

libmicrokit/src/main.c
Indanz: > We should deal with this case, since it is valid for the VMM to signal another component to service some request from the guest then resume the VCPU and let that happen in the background.

But you can't let that happen in the background on x86. So the only argument for supporting `microkit_deferred_flush()` is to make (VMM) code portable between Arm and x86.

> I don't see how a `microkit_deferred_flush()` would help though.

It helps that you can call this function in other places when you do need to flush deferred signals, instead of replicating the same logic everywhere. It limits the poking of Microkit internals to one place. It's a matter of time till this is needed elsewhere. I would actually argue that this should be a user API function, in case of e.g. shutdown or restart or other exceptional corner cases.

> For non VTX cases, "flushing deferred" then waiting for the next event happens within 1 syscall. So having the highlighted change is necessary for the VTX case since we can't refactor it with the non VTX cases.

You already don't support multiple deferred signals, whether to support "deferred signals" for a VMM is an implementation choice you make. It is not necessary.

libmicrokit/include/microkit.h
dreamliner787-9: > I strongly advise against some trampoline construction that shuffles all VCPU handling to one VMM task: The reason to have multiple VCPUs is to use multiple cores, so they all will be on different cores. If you handle all VCPUs with one VMM task, all operations become cross-core with huge overhead: First the fault/trap needs to be propagated to the core the VMM is on, and then any reply needs to be funnelled back.

I agree on the cross-core overhead. I wouldn't recommend it for anything production grade. The reason I brought it up is outlined below.

> The added value of handling multiple, independent virtual machines with one VMM is approximately zero.

If you mean handling multiple independent VMs, I agree. However, if you mean handling multiple VCPUs of a single VM with one VMM, I do see value in it from a prototyping perspective.

We currently have the single-VMM + multi-VCPU model working on our ARM VMM, and it was a non-invasive evolution from the single-VCPU code. While x86 would require a trampoline approach, some Microkit changes for multiple VCPUs in one PD, and incur a performance hit. It offers a much simpler development path to a working prototype. It allows the user to get a basic multi-core VM booting first, and then mature their implementation into a performant multi-VMM design later.

That said, I see the argument for keeping the x86 VCPU API aligned with the current kernel implementation. If you feel the prototyping benefit isn't worth accommodating, I'm happy to remove the `microkit_child vcpu` arguments from all the x86 VCPU APIs. However, since dropping them across the rest of the x86 VCPU APIs would be a breaking change, I'd prefer to handle that API restructuring in a separate, dedicated PR so we can keep this one unblocked and non-breaking.

> Why? A VMM doesn't need access to the virtual machine's memory, interrupts etc.

On x86, the VMM needs access to the VM's memory primarily for fault handling. For example, consider an EPT VM Exit when the guest accesses an emulated I/O APIC. The architecture tells you the faulting guest physical address and whether it was a read or a write, but unlike ARM, it does not provide the read/write width or the target CPU registers. To emulate the operation, the VMM must take the instruction pointer, walk the guest page tables, fetch the faulting instruction from guest RAM, and decode it. If multiple VMMs are managing different VCPUs for a single VM, all of them need access to guest RAM to do this.

Additionally, if the VMM is emulating a storage or network device (e.g., VirtIO), it naturally needs access to guest RAM to process DMA requests.

Regarding interrupts, this is necessary for pass-through devices. Since mainline seL4 does not yet support posted interrupts, the physical interrupt must be routed to the VMM so the injection process can be emulated in software.

> Why do you need IPIs? Why would one VMM care about what others do?

The guest OS will send IPIs between any pair of VCPUs. If a single VM is split across multiple VMMs, they must coordinate to inject these IPIs into the target VCPUs. Connecting the VMMs with an $n(n-1)$ channel mesh is the brute-force way to handle this. The central control approach you mentioned would solve this problem at a cost of more context switching.

libmicrokit/src/main.c
dreamliner787-9: >But you can't let that happen in the background on x86.

Could you please elaborate on what you mean? The kernel can pre-empt the VCPU and schedule other tasks in the system, which could be servers that can service requests from the guest OS (e.g. reading a block from the storage device).

>It helps that you can call this function in other places when you do need to flush deferred signals, instead of replicating the same logic everywhere. It limits the poking of Microkit internals to one place. It's a matter of time till this is needed elsewhere.

Ok I see, I will add that.

>I would actually argue that this should be a user API function, in case of e.g. shutdown or restart or other exceptional corner cases.

I don't think we should support this case, because if the user calls a "deferred" API then they should know that the underlying operation won't happen immediately. This is also documented in the manual:
```
The same as microkit_notify but will instead not actually perform the notify
until the entry point where microkit_deferred_notify was called returns.
```

> whether to support "deferred signals" for a VMM is an implementation choice you make. It is not necessary.

I've decided to support it for VMMs here since it keep the API consistent and avoid "catches" such as you can't use X API on Y configuration.

libmicrokit/include/microkit.h
Indanz: Is Microkit production grade or a prototype toolkit? That's the choice that you need to make, and be very clear about what your design goals are.

> On x86, the VMM needs access to the VM's memory primarily for fault handling.

Wonderful, that explains where virtio came from.

> Additionally, if the VMM is emulating a storage or network device (e.g., VirtIO), it naturally needs access to guest RAM to process DMA requests.

That's why I hate virtio, because it forces write access.

> the physical interrupt must be routed to the VMM so the injection process can be emulated in software.

The seL4 manual doesn't explain how IRQ injection works on x86, but I assume it's by manipulating VMCS and the return message for `seL4_VMEnter()`. You don't need access to the interrupts for that, you just need to know when the real interrupt happened, and you do that via notifications, not the IRQ objects themselves. You also don't want to handle all interrupts on all cores, so this would be partitioned anyway.

> The guest OS will send IPIs between any pair of VCPUs.

Fair enough, that's a good reason.

> Connecting the VMMs with an $n(n-1)$ channel mesh is the brute-force way to handle this. The central control approach you mentioned would solve this problem at a cost of more context switching.

You don't need full-blown channels for just IPIs though, you could limit it to one shared memory page and $n$ notification caps.

But anyway, IPIs are just passing info through, it's pretty stateless. I was talking more about VMMs needing to communicate between them for internal reasons.

libmicrokit/src/main.c
dreamliner787-9: > It helps that you can call this function in other places when you do need to flush deferred signals, instead of replicating the same logic everywhere. It limits the poking of Microkit internals to one place. It's a matter of time till this is needed elsewhere. I would actually argue that this should be a user API function, in case of e.g. shutdown or restart or other exceptional corner cases.

I have added that here: https://github.com/seL4/microkit/pull/523, once that is merged I'll rebase this PR. Edit: done.

docs/manual.md
```diff
@@ -826,6 +863,20 @@ virtual CPU with ID `vcpu`.
 Write the registers of a given virtual CPU with ID `vcpu`. The `regs` argument is the pointer to
 the struct of registers `seL4_VCPUContext` that are written from.
 
+## `void microkit_vcpu_x86_on(void)`
+
+Allow the PD to switch to guest execution mode with the bound vCPU every time `init()`, `notified()` or `fault()` return.
```
midnightveil:
```suggestion
Start running the vCPU bound to this PD.

The vCPU runs after every return from a Microkit entrypoint; it does not run concurrently with the native thread.
```

Thoughts?

(and similar for vcpu off).

I don't think we need to discuss implementation details in the documentation, only what it means.

libmicrokit/include/microkit.h
```diff
@@ -393,28 +398,85 @@ static inline seL4_Uint32 microkit_x86_ioport_read_32(microkit_ioport ioport_id,
...
+#define VMX_GUEST_RIP 0x0000681E
+#define VMX_CONTROL_PRIMARY_PROCESSOR_CONTROLS 0x00004002
+#define VMX_CONTROL_ENTRY_INTERRUPTION_INFO 0x00004016
+
+/* Serve the guest's:
+ * - RIP,
+ * - Primary Processor Based VM Execution Controls, and
+ * - VM Entry Interruption-Information
+ * from local variables that were written to by `microkit_vcpu_x86_write_vmcs()`.
+ * The kernel will service reading from other fields.
```
midnightveil: "Serve" is a bit of a weird verb here. I'd say "Read".

Also, I'd start with "Read a guest VMCS field" as the primary thing, then mention just a comment that certain values are served from local variables? (That is after all, an implementation detail.)

libmicrokit/include/microkit.h
```diff
@@ -393,28 +398,85 @@ static inline seL4_Uint32 microkit_x86_ioport_read_32(microkit_ioport ioport_id,
...
+        }
+        value = ret.value;
+        break;
+    }
+    }
+
+    return value;
+}
+
+/* These fields will be written to local variables rather than the actual VMCS:
```
midnightveil: Ditto here. This is mostly talking about implementation details, which should belong inside the function. Describe that it writes a VMCS field.

Reading "microkit.h" it's somewhat obvious that the comment immediately above about the reference in seL4/the intel SDM, but if you use an IDE or we ever added doxygen compiling them the comments about what is "field" would be lost, so I'd recommend adding it to these two function's docs.

libmicrokit/include/microkit.h
```diff
@@ -393,28 +398,85 @@ static inline seL4_Uint32 microkit_x86_ioport_read_32(microkit_ioport ioport_id,
...
+        break;
+    case VMX_CONTROL_ENTRY_INTERRUPTION_INFO:
+        microkit_x86_vcpu_state.irq_info = value;
+        break;
+    default: {
+        seL4_X86_VCPU_WriteVMCS_t ret = seL4_X86_VCPU_WriteVMCS(BASE_VCPU_CAP + vcpu, field, value);
+        if (ret.error != seL4_NoError) {
+            microkit_dbg_puts("microkit_x86_write_vmcs: error writing data\n");
+            microkit_internal_crash(ret.error);
+        }
```
midnightveil, line 478: No. This clearly can fail if the user makes a mistake (such as an invalid field, or if seL4 doesn't allow writing to that field), so it should actually return seL4_Error, not void.

docs/manual.md
```diff
@@ -826,6 +863,20 @@ virtual CPU with ID `vcpu`.
...
+
+The caller is responsible for initialising the VCPU's instruction pointer, Primary Processor-Based VM-Execution Controls,
+VM-Entry Interruption-Information Field and other architectural states to the corresponding VMCS fields via
```
midnightveil:
```suggestion
VM-Entry Interruption-Information Field, and other architectural states to the corresponding VMCS fields via
```

This sentence doesn't really make sense. "responsible for initialising the other architectural states to the corresponding VMCS fields".

Why not just say "The caller should initialise the VMCS fields per (the SDM/seL4))". Do they all initialise to zero by default? If so, should mention that.

libmicrokit/include/microkit.h
```diff
@@ -265,7 +270,7 @@ static inline void microkit_vcpu_arm_write_reg(microkit_child vcpu, seL4_Word re
         microkit_internal_crash(err);
     }
 }
-#endif
+#endif /* CONFIG_ARM_HYPERVISOR_SUPPORT */
```
midnightveil, line 273: These changes to the arm stuff and the ifdefs; these should be part of a separate commit.

Feel free to merge the manual changes and API changes and rust changes into one commit (they are related, and standing alone doesn't really make snse), but these can be standalone.

libmicrokit/include/microkit.h
```diff
@@ -479,7 +541,20 @@ static inline void microkit_vcpu_x86_write_regs(microkit_child vcpu, seL4_VCPUCo
...
+static inline void microkit_vcpu_x86_on(void)
+{
+    /* On x86, a TCB can only have one bound VCPU at any given time.
+     * So we don't take a `microkit_child vcpu` here. */
+    microkit_x86_vcpu_state.do_resume = seL4_True;
+}
+
+static inline void microkit_vcpu_x86_off(void)
+{
+    microkit_x86_vcpu_state.do_resume = seL4_False;
```
midnightveil: Might as well call this "is_on" given that's what the API is called.

Also, we can use `stdbool.h` bool, so true/false. We don't need to use `seL4_False`.

libmicrokit/src/main.c
```diff
@@ -71,10 +77,89 @@ static void deferred_flush(void)
     }
 }
 
+#if defined(CONFIG_VTX)
+static seL4_MessageInfo_t x86_vcpu_resume(seL4_Word *badge)
+{
+    /* There isn't a seL4 syscall that atomically send signal then perform vmenter while waiting for
+        incoming notifications. So we have to dispatch any deferred signal first. Then switch execution
+        to the bound VCPU. A PD with a VCPU can't receive PPC on x86 so no need to check `have_reply`. */
+    deferred_flush();
```
midnightveil: It doesn't really matter that it's atomic. That being said:


```suggestion
     
    /* We should never have a reply message from the `protected()` endpoint,
     * as on x86 a PD with a bound vCPU cannot receive PPCs. */
    assert(!have_reply);
    /* There is no seL4 invocation which combines a VMEnter and a non-blocking send.
     * Thus, we must perform any deferred signals from `microkit_deferred_notify()` /
     * `microkit_deferred_irq_ack()` before invoking VMEnter. */
    deferred_flush();
```

libmicrokit/src/main.c
```diff
@@ -71,10 +77,89 @@ static void deferred_flush(void)
...
+
+    microkit_mr_set(SEL4_VMENTER_CALL_EIP_MR, microkit_x86_vcpu_state.rip);
+    microkit_mr_set(SEL4_VMENTER_CALL_CONTROL_PPC_MR, microkit_x86_vcpu_state.prim_proc_ctl);
+    microkit_mr_set(SEL4_VMENTER_CALL_INTERRUPT_INFO_MR, microkit_x86_vcpu_state.irq_info);
+
+    microkit_x86_vmenter_result = seL4_VMEnter(badge);
```
midnightveil: Ugh, I feel that this really should have been a `seL4_VMEnter(badge, eip, control, interrupt)` API... Mind opening an issue?

libmicrokit/src/main.c
```diff
@@ -71,10 +77,89 @@ static void deferred_flush(void)
...
+    if (microkit_x86_vmenter_result == SEL4_VMENTER_RESULT_FAULT) {
+        return seL4_MessageInfo_new(0, 0, 0, SEL4_VMENTER_NUM_FAULT_MSGS);
+    } else {
+        /* VCPU got interrupted due to a notification, no tag. */
+        return seL4_MessageInfo_new(0, 0, 0, 0);
+    }
+}
+#endif
+
+static seL4_MessageInfo_t receive_event(bool have_reply, seL4_MessageInfo_t reply_tag, seL4_Word *badge)
```
midnightveil: This refactoring into `receive_event` (and others) should be split out into a separate commit that just does the refactor without changing the behaviour. That way, if it does accidentally change the behaviour, we can easily bisect to that commit, and also, we can see the VMEnter difference much easier.

But I think actually, unless I'm missing something, we can remove that. We shouldn't need anywhere near as much refactoring as was done. (After you fixup my other comments, which will simplify it if I'm not wrong in my understanding)

libmicrokit/src/main.c
```diff
@@ -71,10 +77,89 @@ static void deferred_flush(void)
...
+    microkit_mr_set(SEL4_VMENTER_CALL_CONTROL_PPC_MR, microkit_x86_vcpu_state.prim_proc_ctl);
+    microkit_mr_set(SEL4_VMENTER_CALL_INTERRUPT_INFO_MR, microkit_x86_vcpu_state.irq_info);
+
+    microkit_x86_vmenter_result = seL4_VMEnter(badge);
+
+    microkit_x86_vcpu_state.rip = microkit_mr_get(SEL4_VMENTER_CALL_EIP_MR);
+    microkit_x86_vcpu_state.prim_proc_ctl = microkit_mr_get(SEL4_VMENTER_CALL_CONTROL_PPC_MR);
+    microkit_x86_vcpu_state.irq_info = microkit_mr_get(SEL4_VMENTER_CALL_INTERRUPT_INFO_MR);
+
+    /* Create a dummy tag so that we can call `fault()`, as VM Exits on x86 isn't modelled as an IPC like ARM. */
```
midnightveil:
```suggestion
    /* Create a dummy msgInfo so that we can call `fault()`, as on x86 a VMExit is not an IPC as on other architectures. */
```

libmicrokit/src/main.c
```diff
@@ -71,10 +77,89 @@ static void deferred_flush(void)
...
+        /* VCPU got interrupted due to a notification, no tag. */
+        return seL4_MessageInfo_new(0, 0, 0, 0);
+    }
+}
+#endif
+
+static seL4_MessageInfo_t receive_event(bool have_reply, seL4_MessageInfo_t reply_tag, seL4_Word *badge)
+{
+#if defined(CONFIG_VTX)
+    microkit_x86_vmenter_result = UINT64_MAX;
```
midnightveil: This is unnecessary, yes: it's immediately overwrriten

libmicrokit/src/main.c
```diff
@@ -71,10 +77,89 @@ static void deferred_flush(void)
...
+}
+
+static bool is_fault(seL4_Word badge)
+{
+#if defined(CONFIG_VTX)
+    if (microkit_x86_vmenter_result == SEL4_VMENTER_RESULT_FAULT) {
+        return true;
+    }
+#endif
+    return (badge >> 62) & 1;
```
midnightveil: Instead of doing this, if we see that the `VMEnter` was a fault, we should instead update the badge so that it appears as a fault.

From the syscall implementation it doesn't get set if it was a fault, so we can safely change it.

```
    if (!fault && sender) {
        *sender = badge;
    }
```

libmicrokit/src/main.c
```diff
@@ -71,10 +77,89 @@ static void deferred_flush(void)
...
+}
+
+static seL4_Bool handle_fault(seL4_Word badge, seL4_MessageInfo_t tag, seL4_MessageInfo_t *reply_tag)
+{
+    seL4_Bool reply_to_fault = fault(badge & PD_MASK, tag, reply_tag);
+#if defined(CONFIG_VTX)
+    if (microkit_x86_vmenter_result == SEL4_VMENTER_RESULT_FAULT) {
+        /* There won't be anything to reply to for a VCPU fault. But
+         * if fault() returns false then we shouldn't resume the VCPU. */
```
midnightveil: Is this behaviour documented in the manual?

libmicrokit/src/main.c
```diff
@@ -71,10 +77,89 @@ static void deferred_flush(void)
...
+    }
+#endif
+    return (badge >> 62) & 1;
+}
+
+static seL4_Bool handle_fault(seL4_Word badge, seL4_MessageInfo_t tag, seL4_MessageInfo_t *reply_tag)
+{
+    seL4_Bool reply_to_fault = fault(badge & PD_MASK, tag, reply_tag);
+#if defined(CONFIG_VTX)
+    if (microkit_x86_vmenter_result == SEL4_VMENTER_RESULT_FAULT) {
```
midnightveil: Surely if we return from fault then the vmenter result must be fault already? 

This seems unnecessary.

tool/microkit/tests/sdf/vm_x86_ppc_invalid_1.system
```diff
@@ -0,0 +1,39 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+ Copyright 2026, UNSW (ABN 57 195 873 179)
```
midnightveil: I don't think we normally put the ABN in, but I'm also not the expert here.

libmicrokit/src/main.c
midnightveil: For now, we can make this better:  just call `x64_sys_send_recv` directly, which is what VMEnter does anyway...

```
#ifdef CONFIG_VTX
LIBSEL4_INLINE_FUNC seL4_Word seL4_VMEnter(seL4_Word *sender)
{
    seL4_Word fault;
    seL4_Word badge;
    seL4_Word mr0 = seL4_GetMR(0);
    seL4_Word mr1 = seL4_GetMR(1);
    seL4_Word mr2 = seL4_GetMR(2);
    seL4_Word mr3 = seL4_GetMR(3);

    x64_sys_send_recv(seL4_SysVMEnter, 0, &badge, 0, &fault, &mr0, &mr1, &mr2, &mr3, 0);

    seL4_SetMR(0, mr0);
    seL4_SetMR(1, mr1);
    seL4_SetMR(2, mr2);
    seL4_SetMR(3, mr3);
    if (!fault && sender) {
        *sender = badge;
    }
    return fault;
}
```

midnightveil (changes requested):

tool/microkit/src/sdf.rs
```diff
@@ -2151,6 +2151,25 @@ pub fn parse(
             ));
         }
 
+        // When seL4_VMEnter() is called, the kernel only checks the VMM's bound
+        // notification for pending signals. Because the endpoint object isn't passed
+        // or checked for pending messages, PPC won't work while the VCPU is on.
+        // Technically, PPC could still work when the VCPU is off, but we shouldn't
+        // expose this footgun to users.
```
midnightveil: If there's no endpoint, we should also disallow child PDs.

(`assert needs_ep() == false`?)

tool/microkit/src/sdf.rs
```diff
@@ -2151,6 +2151,25 @@ pub fn parse(
...
+            && ((ch.end_a.pp && pd_b.virtual_machine.is_some())
+                || (ch.end_b.pp && pd_a.virtual_machine.is_some()))
+        {
+            return Err(format!(
+                "Error: It is not possible to PPC to PD '{}' with a bound vCPU from PD '{}' on x86_64 @ {}:{}:{}",
+                    if ch.end_a.pp {&pd_b.name} else {&pd_a.name},
+                    if ch.end_a.pp {&pd_a.name} else {&pd_b.name},
+                    filename.display(),
+                    pd_a.text_pos.unwrap().row,
+                    pd_a.text_pos.unwrap().col
```
midnightveil, line 2177: use loc_to_string

libmicrokit/src/main.c
Indanz: > > But you can't let that happen in the background on x86.
> 
> Could you please elaborate on what you mean? The kernel can pre-empt the VCPU and schedule other tasks in the system, which could be servers that can service requests from the guest OS (e.g. reading a block from the storage device).

The point of the deferred variant is to avoid one syscall. But with VCPU on x86, you can't avoid the extra syscall, so calling the deferred version makes no sense. Ergo, supporting deferred calls for x86 VMMs makes no sense.

The usual argument that code should be portable doesn't apply for x86 VMM code, as that can't be portable by definition.

> > I would actually argue that this should be a user API function, in case of e.g. shutdown or restart or other exceptional corner cases.
> 
> I don't think we should support this case, because if the user calls a "deferred" API then they should know that the underlying operation won't happen immediately.

Yes, but they may not know at the time they call the deferred version that later on, something exceptional happens. Example: Restart command via message queue. The messages may be deferred ACKed via a signal after receiving it, but when the command gets processed and the restart happens, that deferred signal will disappear. The restart code and message queue code is independent and not aware of each other's details.

> > whether to support "deferred signals" for a VMM is an implementation choice you make. It is not necessary.
> 
> I've decided to support it for VMMs here since it keep the API consistent and avoid "catches" such as you can't use X API on Y configuration.

But such catches are minor compared to the "can't defer more than one signal" catch. And the only reason to have such catches at all is to avoid extra overhead, which you are adding here for a nonsensical use case.

At the very least make the deferred functions safe to be called multiple times by flushing pending signals automatically.

docs/manual.md
dreamliner787-9: I think we should keep the old version since it is the exact behaviour, whereas the new version can implies that you lose control once you call `microkit_vcpu_x86_on()` which is not the case.

midnightveil: But I state that?

This assumes one knows how the handler loop work, I don't think itm akes sense to rephrase that.

dreamliner787-9: Oh whoops that's right sorry I did not read the second sentence...

libmicrokit/src/main.c
dreamliner787-9: Yes it is documented here:

> ## `seL4_Bool fault(microkit_child child, microkit_msginfo msginfo, microkit_msginfo *reply_msginfo)`
>...
>The returned `seL4_Bool` is whether or not to reply to the fault with the message `reply_msginfo`.
Returning `seL4_True` will reply to the fault. Returning `seL4_False` will not reply to the fault.

midnightveil: No, that documents "replies to fault", i.e. the internal behaviour (implementation).

We should document the *behaviour*, i.e.: replying false will instead make the PD/vCPU not run anymore, and replying true will make it resume from the fault (at what `$pc`? the fault restart pc).

dreamliner787-9:
Hmm not sure what to do about the Intel SDM link check failing, even https://www.intel.com/content/www/us/en/developer/articles/technical/intel-sdm.html does not work, presumably due to some anti bot protection?

midnightveil:
We can add it to linkcheck-ignore: https://github.com/seL4/microkit/blob/main/.linkcheck-ignore.yml

libmicrokit/include/microkit.h
```diff
@@ -9,7 +9,11 @@
 
 #pragma once
 
+#include <stdbool.h>
```
midnightveil: why add this to this file if not used in this file?

libmicrokit/include/microkit.h
```diff
@@ -393,28 +399,86 @@ static inline seL4_Uint32 microkit_x86_ioport_read_32(microkit_ioport ioport_id,
...
+#if defined(CONFIG_VTX)
+/* Architecturally defined identifiers for a x86 VCPU's VMCS fields,
+ * see seL4 source: `include/arch/x86/arch/object/vcpu.h`
+ * or Intel® 64 and IA-32 Architectures Software Developer’s Manual
+ * Combined Volumes: 1, 2A, 2B, 2C, 2D, 3A, 3B, 3C, 3D, and 4
+ * Order Number: 325462-084US June 2024
+ * "Table B-8. Encodings for 32-Bit Control Fields (0100_00xx_xxxx_xxx0B)" and
+ * "Table B-14. Encodings for Natural-Width Guest-State Fields (0110_10xx_xxxx_xxx0B) (Contd.)".
+ * */
+#define VMX_GUEST_RIP 0x0000681E
```
midnightveil: these should be namespaced (I hate C)

libmicrokit/src/main.c
```diff
@@ -71,10 +78,46 @@ static void deferred_flush(void)
...
+{
+    /* There is no seL4 invocation which combines a VMEnter and a non-blocking send.
+     * Thus, we must perform any deferred signals from `microkit_deferred_notify()` /
+     * `microkit_deferred_irq_ack()` before invoking VMEnter. */
+    deferred_flush();
+
+    seL4_Word is_fault, fault_reason;
+
+    x64_sys_send_recv(seL4_SysVMEnter, 0, badge, 0, &is_fault,
+                      &microkit_x86_vcpu_state.rip, &microkit_x86_vcpu_state.prim_proc_ctl, &microkit_x86_vcpu_state.irq_info, &fault_reason,
```
midnightveil: split onto a new line. that second line is getting long, split into two lines of two args each.

libmicrokit/src/main.c
```diff
@@ -71,10 +78,46 @@ static void deferred_flush(void)
...
+    /* We want to follow the documented kernel behaviour where these 4 values are always populated in the message registers.
+     * However, due to `setMR()`'s behaviour in include/object/tcb.h of the kernel, these 4 values will only be set in
+     * CPU registers. So we need to copy them to the IPC buffer to conform to the documented kernel behaviour.
+     *
+     * The other reason why we want to manage these values is so that the user does not have to worry about saving/updating them
+     * when servicing a notification. */
+    microkit_mr_set(SEL4_VMENTER_CALL_EIP_MR, microkit_x86_vcpu_state.rip);
+    microkit_mr_set(SEL4_VMENTER_CALL_CONTROL_PPC_MR, microkit_x86_vcpu_state.prim_proc_ctl);
+    microkit_mr_set(SEL4_VMENTER_CALL_INTERRUPT_INFO_MR, microkit_x86_vcpu_state.irq_info);
+    /* The 4th value (the exit reason) is only valid when fault is true. */
```
midnightveil: fault_reason / SEL4_VMENTER_FAULT_REASON_MR

libmicrokit/src/main.c
```diff
@@ -97,7 +140,16 @@ static void handler_loop(void)
         seL4_Word badge;
         seL4_MessageInfo_t tag;
 
+#if defined(CONFIG_VTX)
```
midnightveil, line 146: This looks much nicer now!

libmicrokit/src/main.c
Indanz: Or add a local pointer to `microkit_x86_vcpu_state` named `s` and use that...

libmicrokit/include/microkit.h
```diff
@@ -479,7 +543,20 @@ static inline void microkit_vcpu_x86_write_regs(microkit_child vcpu, seL4_VCPUCo
     }
 }
 
-#endif
+static inline void microkit_vcpu_x86_on(void)
+{
+    /* On x86, a TCB can only have one bound VCPU at any given time.
+     * So we don't take a `microkit_child vcpu` here. */
+    microkit_x86_vcpu_state.is_on = true;
```
Indanz: Wouldn't you want to add some check to see if this PD actually has a VCPU and at least print a debug error?

libmicrokit/src/main.c
```diff
@@ -71,10 +78,46 @@ static void deferred_flush(void)
...
+
+    x64_sys_send_recv(seL4_SysVMEnter, 0, badge, 0, &is_fault,
+                      &microkit_x86_vcpu_state.rip, &microkit_x86_vcpu_state.prim_proc_ctl, &microkit_x86_vcpu_state.irq_info, &fault_reason,
+                      0);
+    /* We want to follow the documented kernel behaviour where these 4 values are always populated in the message registers.
+     * However, due to `setMR()`'s behaviour in include/object/tcb.h of the kernel, these 4 values will only be set in
+     * CPU registers. So we need to copy them to the IPC buffer to conform to the documented kernel behaviour.
+     *
+     * The other reason why we want to manage these values is so that the user does not have to worry about saving/updating them
+     * when servicing a notification. */
```
Indanz: Text too wide.

libmicrokit/src/main.c
```diff
@@ -97,7 +140,16 @@ static void handler_loop(void)
...
+#if defined(CONFIG_VTX)
+        if (microkit_x86_vcpu_state.is_on) {
+            /* We should never have a reply message from the `protected()` endpoint,
+            * as on x86 a PD with a bound vCPU cannot receive PPCs.*/
+            // assert(!have_reply);
+            tag = x86_vcpu_resume(&badge);
+        } else if (have_reply) {
+#else
         if (have_reply) {
+#endif
```
Indanz:
```suggestion
        } else
#endif
        if (have_reply) {
```

libmicrokit/src/main.c
```diff
@@ -107,13 +159,21 @@ static void handler_loop(void)
...
 
         have_reply = false;
 
         if (is_fault) {
             seL4_Bool reply_to_fault = fault(badge & PD_MASK, tag, &reply_tag);
+#if defined(CONFIG_VTX)
+            /* If fault() returns false then we shouldn't resume the VCPU. */
+            if (!reply_to_fault) {
+                microkit_x86_vcpu_state.is_on = false;
+            }
```
Indanz, line 176: This points out another limitation of x86 PDs having a VCPU: They are not allowed to be fault handlers of other, non-VCPU tasks.

A VMM could do an IPC call to something where it's a fault handler for safely, even on x86, as long as no one else calls that server.

tool/microkit/tests/sdf/vm_x86_ppc_invalid_1.system
Indanz: Should be explained in https://www.library.unsw.edu.au/content/dam/secure/library/staff-and-students/2023-00-copyright/Author-agreement-to-upload-to-UNSWorks-when-UNSW-owns-the-copyright.pdf

Need to log in to access it, so no idea what that says. Found on https://subjectguides.library.unsw.edu.au/copyright/unsworks.

midnightveil: That doesn't explain it, also it assumes you licens eunder cc-by, and also is for unswworks which is the University's thesis hosting thing.

Indanz: Well, universities usually say how to attribute copyright to them if that's the case, I expected it to be there. Just email them and ask.

It's the copyright owner bit we're interested in, so doesn't matter what licence is used.

lsf37: I'm sure you'll get all possible different answers depending on who you ask in UNSW. Australian authorities want the ABN on any official communication with them, because it uniquely identifies the business. In the end, it really doesn't matter, there is no question which UNSW is meant here. It doesn't particularly hurt to add it, but nobody else in the world does or cares.

libmicrokit/include/microkit.h
dreamliner787-9: It is used though https://github.com/au-ts/microkit/blob/355e52f7cee9c778fd98a459c5197174742b78c3/libmicrokit/include/microkit.h#L550

midnightveil: Using "true" when the type of is_on is seL4_Bool is weird... Should be consistent.

But I might have changed my mind..m since we have a bunch of other stuff using seL4_Bool we should just use seL4_True and not include the header.

Unless we wanna change everything but...

dreamliner787-9: > Using "true" when the type of is_on is seL4_Bool is weird

Ah yea my bad.

> But I might have changed my mind..m since we have a bunch of other stuff using seL4_Bool we should just use seL4_True and not include the header.

I think we should just use seL4_Bool to be consistent

tool/microkit/tests/sdf/vm_x86_ppc_invalid_1.system
midnightveil: Gernot's said we don't need the ABN, so that's definitive enough for me :)

libmicrokit/include/microkit.h
dreamliner787-9: I'll make an issue about this, since it is relevant for ARM as well.

libmicrokit/include/microkit.h
dreamliner787-9: We could resolve this with https://github.com/seL4/microkit/issues/362.

midnightveil (approved):

midnightveil:
Explicitly ignoring the style check, since it is broken.

EDIT: Actually this is going to bug us every time we change main.c.... Hmm. Maybe that's a bad idea. We should fix that.

dreamliner787-9:
> Explicitly ignoring the style check, since it is broken.
> 
> EDIT: Actually this is going to bug us every time we change main.c.... Hmm. Maybe that's a bad idea. We should fix that.

I think it is better if we go with what I had originally:
```diff
-        } else
-#endif
-        if (have_reply) {
+        } else if (have_reply) {
+#else
+        if (have_reply) {
+#endif
```

Indanz:
> > Explicitly ignoring the style check, since it is broken.
> > EDIT: Actually this is going to bug us every time we change main.c.... Hmm. Maybe that's a bad idea. We should fix that.
> 
> I think it is better if we go with what I had originally:

The style checker hates me. The feeling is mutual.

PR #433, opened 2026-03-10
Fix x86 MSI PCI parsing + add x86 IRQ vector range check

dreamliner787-9:


Ivan-Velickovic (commented):
Not really to do with this PR but if/when we refactor SDF to do better error reporting etc for use as a library we should probably have things like `range_error` etc. Sometimes we use `[x..x]` notation, sometimes just describe the range.

Issue #91, opened 2024-01-17
Fix approach used for making PDs passive at run-time


Ivan-Velickovic:
The approach of using `seL4_NBSendRecv` in the event handler loop of libmicrokit when PDs essentially 'make' themselves passive by sending a message to the monitor is not ideal for multiple reasons.

https://github.com/seL4/microkit/blob/ad271b5c1b4df2f16d43cafc1bfe1f7faf6d2de5/libmicrokit/src/main.c#L60

1. It will not scale/work in multi-core setups. Given that it is a non-blocking send, a scenario may occur where the monitor is already dealing with a message from one PD (meaning that it is not waiting on an endpoint) and another PD that wants to make itself passive on another core which means the message will be lost.
2. It affects verification as it leads to assumptions needing to be made about the scheduling of the system.

This issue was found in the process of verifying the passive PD changes in Microkit.

The current proposed solution is to have synchronisation using notifications so that all PDs block until they receive a notification from the monitor that all PDs that needed to be converted to passive have completed. This means that no PDs start their `init` procedure until everything else in the system is also ready to start their `init` procedure. This should remove any concurrency issues such as what happens when a PD is in the middle of being converted to passive and another PD on a separate core tries to communicate with it, I believe this may lead to lost notifications.

This solution is not final as we have not implemented and evaluated it, however, from our internal discussions it I think it is the cleanest solution and should require minimal changes to Microkit.

gernotheiser:
+1

midnightveil:
@dreamliner787-9 Do you remember what was still remaining after the lazy SC rebind?

(Did we ever get the lazy rebind merged into the capDL initialiser?)

dreamliner787-9:
> (Did we ever get the lazy rebind merged into the capDL initialiser?)

No I don't think we have. I can make a patch for it for the July release

PR #440, opened 2026-03-18
monitor: lazy Scheduling Context rebind

dreamliner787-9:
Applied https://github.com/seL4/seL4/pull/523 to workaround https://github.com/seL4/seL4/issues/1617.

To fix the problem properly this should be performed in the capDL initialiser. But that would take a while to get merged so we will merge this first.

Related: https://github.com/seL4/microkit/issues/91

### DIFF
--- a/monitor/src/main.c
+++ b/monitor/src/main.c
@@ -733,18 +733,13 @@ static void monitor(void)
 
         if (label == seL4_Fault_NullFault && pd_id < MAX_PDS) {
             /* This is a request from our PD to become passive */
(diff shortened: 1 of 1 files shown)


monitor/src/main.c
```diff
@@ -732,19 +732,13 @@ static void monitor(void)
         seL4_Word tcb_cap = BASE_PD_TCB_CAP + pd_id;
 
         if (label == seL4_Fault_NullFault && pd_id < MAX_PDS) {
-            /* This is a request from our PD to become passive */
```
Ivan-Velickovic, line 735: This comment is still valid

PR #445, opened 2026-03-24
Domain schedules

Kswin01:
This adds support for domain schedules to the Microkit. It defines syntax that looks like:

```xml
    <!-- This element is used to declare the list of domains, to give
         each domain a human-readable name. -->
    <domains>
        <!-- The first domain contains the Microkit Monitor PD, but can contain others. -->
        <domain name="monitor" id="0" />
        <!-- You give each domain a name and its domain id. -->
        <domain name="emitter" id="1" />
        <!-- The domain ID is optional and can be auto-assigned if desired.
             It will be assigned to the smallest unused ID, and must be less
             than KernelNumDomains. -->
        <domain name="collector" />

        <!-- This is a domain schedule. Note that the same domain can appear twice
             in the schedule.
             Also supports optional attributes "start_index" and "index_shift".
         -->
        <domain_schedule>
            <schedule_entry domain="monitor" duration="6000 us" />
            <schedule_entry domain="emitter" duration="8000 us" />
            <schedule_entry domain="collector" duration="2000 us" />
            <!-- This element is optional, and is implicitly there anyway. -->
            <schedule_end_marker />
        </domain_schedule>
    </domains>

    <protection_domain name="emitter" domain="emitter">
        <program_image path="emitter.elf" />
    </protection_domain>
```

The standard, non-SMP config is changed to always have the domain schedule configured. This is only a minor memory cost to users who do not want to use domains.

Depends on https://github.com/seL4/rust-sel4/pull/360 being merged.

<details><summary>Original description</summary>

This PR adds support for defining domain schedules in Microkit, and relies on the run-time domain scheduler changes due to be merged into seL4.

It also depends on the following branch of rust-sel4: https://github.com/au-ts/rust-sel4/tree/domain_set?branch=domain_set

# Changes for users

In `build_sdk.py` we define two new configurations, `debug_domains` and `release_domains`. These configurations build the kernel with a max of 256 domains, and 256 entries into the domain schedule. The user can change this to values as they see fit. I have separated these from the regular configs as the users may not wish to have the extra memory overhead (although mostly negligible). I can merge these configs into one if its more desirable. 

A domain schedule is defined in the sdf as:
```
  <domain_schedule>
      <domain name="domain_1" length="1000" />
      <domain name="domain_2" length="1000" />
      <domain name="domain_3" length="2000" />
  </domain_schedule>
```
The length is defined in milliseconds, and reads the `TIMER_FREQUENCY` from the kernel config to covert to timer ticks that the kernel requires. This is the case for aarch64 and riscv64. However, on x86, we don't have a static definition of the timer frequency, and is instead set at runtime by the kernel. I'm not sure of the best way to do it here, whether we should do this conversion in the capDL initialiser instead.

The above will insert the schedule using the new `DomainSet` invocations, beginning at index 0 by default. We also provide capDL a start index of 0 by default.

The users can optionally set these values like so:
```
    <domain_schedule>
        <domain name="domain_1" length="1000" />
        <domain name="domain_2" length="1000" />
        <domain name="domain_3" length="2000" />
        <domain_start index="2" />
        <domain_idx_shift shift="10" />
    </domain_schedule>
```

The start index is relative to the schedule list that the user provides not an absolute index. It must be in bounds of the length of schedule that the user defines. We also can have a domain shift, which means that will start inserting the schedule at that shifted index (in the above case 10). The shift + length of the schedule must be less than the kernel configured max number of domain schedules.

We also now build one monitor per domain. The monitor is responsible for handling faults, and also making threads passive if requested. The decision to create a monitor per domain was that these faults/requests can be serviced within that domains timeslice, and not have to wait for a singular domain to be scheduled again.

</details>


Kswin01:
I've switched the default maximums to 64 domains, and 128 domain schedule entries. Would this be sufficient?

Indanz (commented):
Wouldn't it make more sense to make `idx_shift` and `domain_start` attributes of `domain_schedule` instead? Adding it as an attribute to an entry makes no sense.

And I would also make it standard integers, not some custom weird hexadecimal format (assuming that's supported) which can't easily defined with XSD and is less user friendly anyway. If you don't support that already, then perhaps don't use characters that would be valid within hexadecimal, it's ambiguous now.

docs/manual.md
```diff
@@ -1135,6 +1148,26 @@ The `end` element has the following attributes:
...
 The `id` should be passed to the `microkit_notify` and `microkit_ppcall` functions.
 
+## `domain_schedule`
+
+The `domain_schedule` element, by default, has a list of up to 64 `domain` child elements, and can describe 32 different domains. (bounded by the kernel config). This can be configured by changing the number of schedule entries defined in `elaborate_all_board_configs` in `build_sdk.py`. Each child specifies a particular timeslice in the domain schedule and the order of the child elements specifies the order in which the timeslices will be scheduled. A domain may be named more than once in the schedule, in which case the domain will have multiple timeslices in the schedule.
+
+The `domain` element has the following attributes:
+
+* `name`: Name of the domain.
+* `length`: Length of time the domain will run each time it is active, in microseconds on Aarch64 and Riscv64. On x86 this length is currently defined in ticks as we don't have a static definition of the timer frequency.
```
Indanz, line 1158: You know the x86 frequency at boot time, there is no good excuse to not have the same API for x86 as for the others.

docs/manual.md
```diff
@@ -1135,6 +1148,26 @@ The `end` element has the following attributes:
...
+
+The `name` attribute of each `domain` element can be referenced in the `domain` attribute of a `protection_domain` element.
+
+The `domain_idx_shift` element has the following attribute:
```
Indanz, line 1162: This is not an element any more.

docs/manual.md
```diff
@@ -1135,6 +1148,26 @@ The `end` element has the following attributes:
...
+* `shift`: This is the start index of where we will construct our schedule. So if we shift by 5, we will begin constructing the domain schedule from index 5 onwards. If not specified, no shift will be applied, and we will start constructing the schedule from index 0.
+
+
+The `domain_start` element has the following attribute:
```
Indanz, line 1166: Idem.

docs/manual.md
```diff
@@ -1135,6 +1148,26 @@ The `end` element has the following attributes:
...
+
+The `domain` element has the following attributes:
+
+* `name`: Name of the domain.
+* `length`: Length of time the domain will run each time it is active, in microseconds on Aarch64 and Riscv64. On x86 this length is currently defined in ticks as we don't have a static definition of the timer frequency.
+
+The `name` attribute of each `domain` element can be referenced in the `domain` attribute of a `protection_domain` element.
+
+The `domain_idx_shift` element has the following attribute:
+* `shift`: This is the start index of where we will construct our schedule. So if we shift by 5, we will begin constructing the domain schedule from index 5 onwards. If not specified, no shift will be applied, and we will start constructing the schedule from index 0.
```
Indanz, line 1163: I don't think this is comprehensible for people that don't know already.

What about saying something like that the kernel has a fixed array of domain schedules, and that if `domain_idx_shift` is unspecified, the Microkit schedules will be configured starting from index 0. If a shift is specified, then the first schedule entry will start at index `domain_idx_shift` of the kernel domain schedule array.

docs/manual.md
```diff
@@ -1135,6 +1148,26 @@ The `end` element has the following attributes:
...
+
+
+The `domain_start` element has the following attribute:
+* `index`: This is the start index of the domain schedule. When the kernel reaches the end of the domain schedule, it will wrap around to this index again. This index is an offset into the user defined schedule, and not an absolute index. If no domain start index is set, then the Microkit tool will set this to 0 by default.
```
Indanz, line 1167: I'd move this higher up, before domain_idx_shift. Then you can mention the absolute index concept there and this one is less confusing.

docs/manual.md
```diff
@@ -1135,6 +1148,26 @@ The `end` element has the following attributes:
...
+
+The `domain_schedule` element is only valid of using non-smp configs.
```
Indanz, line 1169: if? Or better, "for non-smp configs" or "Domains are not supported for SMP configs."

docs/manual.md
```diff
@@ -1818,6 +1851,8 @@ which PD caused an exception and details on the PD's state at the time of the fa
 Other than printing fault details, the monitor does not do anything to handle
 the fault, it will simply go back to sleep waiting for any other faults.
 
+When domains are enabled, we will build one monitor for each domain.
```
Indanz, line 1854: But it's confusing that you call tasks "protection domains", so can't really just say domain without it being confusing in some way.

example/hello_domains/Makefile
Indanz, line 1: It seems that 90% of the file is just copied from elsewhere.

What about making the common part a separate file that is included by the top-level Makefile? That would avoid a lot of duplication...

example/hello_domains/Makefile
```diff
@@ -0,0 +1,70 @@
...
+IMAGE_FILE = $(BUILD_DIR)/loader.img
+REPORT_FILE = $(BUILD_DIR)/report.txt
+
+all: $(IMAGE_FILE)
+
+$(BUILD_DIR)/%.o: %.c Makefile
+	$(CC) -c $(CFLAGS) $< -o $@
+
+$(BUILD_DIR)/hello.elf: $(addprefix $(BUILD_DIR)/, $(HELLO_OBJS))
+	$(LD) $(LDFLAGS) $^ $(LIBS) -o $@
```
Indanz, line 67: This could be a generic rule for `%.elf` instead of a specific one.

example/hello_domains/Makefile
```diff
@@ -0,0 +1,70 @@
...
+
+$(IMAGE_FILE) $(REPORT_FILE): $(addprefix $(BUILD_DIR)/, $(IMAGES)) hello.system
+	$(MICROKIT_TOOL) hello.system --search-path $(BUILD_DIR) --board $(MICROKIT_BOARD) --config $(MICROKIT_CONFIG) -o $(IMAGE_FILE) -r $(REPORT_FILE)
```
Indanz, line 70: If you add a variable for hello.system, the rest could go in the common file.

tool/microkit/src/capdl/builder.rs
```diff
@@ -393,109 +400,129 @@ pub fn build_capdl_spec(
...
+    for (monitor_idx, monitor_elf) in monitor_elfs.iter().enumerate() {
+        let monitor_name = format!("{MONITOR_PD_NAME}_{monitor_idx}");
+        // Make sure that no other PD has the same name as this monitor
+        for pd in system.protection_domains.iter() {
+            if pd.name == monitor_name {
+                return Err(format!(
+                    "Error: User defined PD name clashes with monitor name '{monitor_name}'",
+                ));
+            }
+        }
```
Indanz, line 416: Why other? It's just a name.

Either you support duplicate names, and then there is no problem, or you don't, in which case you already have duplicate name detection somewhere.

tool/microkit/src/capdl/builder.rs
```diff
@@ -661,15 +700,16 @@ pub fn build_capdl_spec(
         ));
 
         // Step 3-5 Create fault Endpoint cap to parent/monitor
-        let pd_fault_ep_cap = if let Some(pd_parent) = pd.parent {
-            assert!(pd_global_idx > pd_parent);
+        let pd_fault_ep_cap = if let Some(parent) = pd.parent {
+            // @kwinter: Should we enforce domain boundaries here?
```
Indanz, line 704: Probably not, that would be quite restrictive.

tool/microkit/src/capdl/builder.rs
```diff
@@ -1079,7 +1123,35 @@ pub fn build_capdl_spec(
...
+    // *********************************
+
+    if let Some(sys_domain_schedule) = &system.domain_schedule {
+        let mut domain_schedule: Vec<DomainSchedEntry> = Vec::new();
+        // We want to convert from the microseconds that the user defines in the
+        // sdf to the kernel scheduler ticks. If we are on x86, we won't necessarily
+        // have a static definition of the timer frequency. We will express the
+        // domain timeslices in ticks.
+        let ticks_in_ms = if let Some(timer_freq) = kernel_config.timer_freq {
+            timer_freq / US_IN_S
```
Indanz, line 1136: Shouldn't it be `ticks_in_us` now?

This assumes a round number of MHz for the timer frequency. If it's not, you really should to do the rounding at the end instead of multiplying the rounding error.

tool/microkit/src/main.rs
```diff
@@ -650,8 +664,17 @@ fn main() -> Result<(), String> {
         }
     }
 
-    // The monitor is just a special PD
-    system_elfs.push(monitor_elf);
+    // Create a copy of the monitor elf for every domain in the system
```
Indanz, line 667: This seems a bit silly, but it works I suppose.

Why do you have a monitor per domain again? That's a bit unclear to me.

If monitors are merely fault handlers that handle unexpected faults, then having one in domain 0 would be nice, as you can separate it from the real-time code, so one task faulting doesn't affect other execution in that domain. They seem to do a lot of slow UART printing, and they're not passive, but at a higher priority it seems.

(They also convert a task to passive, but that's not needed with lazy SC unbind.)

tool/microkit/src/sdf.rs
```diff
@@ -987,7 +1013,8 @@ impl ProtectionDomain {
                     checked_add_setvar(&mut setvars, setvar, xml_sdf, &child)?;
                 }
                 "protection_domain" => {
-                    let child_pd = ProtectionDomain::from_xml(config, xml_sdf, &child, true)?;
+                    let child_pd =
+                        ProtectionDomain::from_xml(config, xml_sdf, &child, true, domain_schedule)?;
```
Indanz, line 1017: Can't you fit this on one line? This is extremely ugly.

If that breaks style, update your style: Rust seems to need a lot of horizontal space.

tool/microkit/src/sdf.rs
```diff
@@ -273,6 +274,23 @@ pub struct ProtectionDomain {
...
+    pub id: u8,
+    pub length: u64,
+}
+
+#[derive(Debug, PartialEq, Eq, Clone)]
+pub struct DomainSchedule {
+    pub domain_ids: HashMap<String, u8>,
+    pub schedule: Vec<DomainTimeslice>,
+    pub domain_start_idx: Option<u64>,
+    pub domain_idx_shift: Option<u64>,
```
Indanz, line 293: Why are these u64? That seems excessive.

tool/microkit/src/sel4.rs
```diff
@@ -289,6 +292,9 @@ pub struct Config {
     /// The two remaining fields are only valid on ARM and RISC-V
     pub device_regions: Option<Vec<PlatformConfigRegion>>,
     pub normal_regions: Option<Vec<PlatformConfigRegion>>,
+    /// Number of domain and domain schedules the kernel is configure for
```
Indanz, line 295:
```suggestion
    /// Number of domains and domain schedules the kernel is configured for
```

midnightveil:
I'm inclined to leave the number of domains at the verification default of 16.

(The verification is generic over the number of domains, right, though?)

lsf37:
> I'm inclined to leave the number of domains at the verification default of 16.
> 
> (The verification is generic over the number of domains, right, though?)

Yes, it can be configured to anything <=255.

midnightveil:
The original branch is here: https://github.com/au-ts/microkit/tree/archive/krishan-runtime-domains

I'm going to restart this branch and reorder a lot of the commits here, because upstream has changed enough it's annoying to rebase now.

docs/manual.md
midnightveil: Ack.

tool/microkit/src/sdf.rs
midnightveil: It's what rust-seL4 and capDL uses.

example/domains/Makefile
midnightveil: The examples are supposed to be standalone so they can be copied by users, as they are packaged in the SDK. We could add some script that autogenerates them or something so this stays true.

docs/manual.md
```diff
@@ -1118,6 +1156,50 @@ The `iomap` element supports the following attributes:
...
+
+It supports two types of child element, the `<schedule_entry>` and the
+`<schedule_end_marker>`.
+
+The `<schedule_entry>` requires the following attributes:
+
+* `domain`: The name of the domain to be scheduled.
+* `duration`: The duration for which the domain should be active.
+              This contains a value and a unit, i.e. `1000 us`.
+              The unit can either be 'us' or 'ticks'.
```
Furao, line 1202: I think a `duration_unit` or `duration_type` attribute in the `domain_schedule` element would be better than making this `duration` a value and a unit. It would make parsing easier and I can't imagine wanting a schedule that mixes units.

docs/manual.md
```diff
@@ -1118,6 +1156,50 @@ The `iomap` element supports the following attributes:
...
            Defaults to read-write.
 
+## `domains` {#sdf-domains}
+
+The `domains` element describes the (security) domains and their schedule within
+a microkit system. seL4 only supports the domain scheduler on non-SMP (unicore)
+configurations. There is a fixed upper limit on the number of domains and the
+number of schedule entries, determined by the kernel configurations `KernelNumDomains`
+and `KernelNumDomainSchedules`. In the default SDK build, we support 16 domains and
+100 schedule entries.
```
Furao, line 1166: This probably makes more sense in the seL4 manual, but it would be nice to know the actual memory trade-off of increasing these numbers.

Indanz: Memory usage of various data structures scales by the number of priorities multiplied by the number of domains. If you want to reduce the memory usage, reducing the number of priories from 256 to something more sane is the best way of doing it. So if you reduce the number of priorities to 16, you can have 16 domains with the same memory usage as no domains and 256 priorities.

`KernelNumDomainSchedules` doesn't affect anything else, it uses 8 bytes per entry. I picked 100 as a default to make it clear there that the number is arbitrary and has no special power-of-two properties, while being large enough for most practical purposes. Microkit uses 100 because that is the kernel default.

docs/manual.md
midnightveil: That's fair. My primary reason for doing it like this is that I like the way it reads better, that is, `duration="1000"` is confusing but `duration="1000 us"` reads like how you would write a value in English.

Parsing wise it's not that hard, not any more than I would have to do to add another attribute.

Not sure what's "best".

I can definitely see a few situations where for the most care you are OK with the us conversions but where some domains are shorter or you want exact control.

docs/manual.md
```diff
@@ -1118,6 +1156,50 @@ The `iomap` element supports the following attributes:
 * `perms`: Identifies the permissions with which to map the memory region. Can be a combination of `r` (read), and `w` (write).
            Defaults to read-write.
 
+## `domains` {#sdf-domains}
+
+The `domains` element describes the (security) domains and their schedule within
+a microkit system. seL4 only supports the domain scheduler on non-SMP (unicore)
```
dreamliner787-9:
```suggestion
a Microkit system. seL4 only supports the domain scheduler on non-SMP (unicore)
```

docs/manual.md
```diff
@@ -1118,6 +1156,50 @@ The `iomap` element supports the following attributes:
...
+* `name`: A unique name for the domain.
+* `id`: (optional) The domain ID.
+
+The `domain_schedule` element specifies the domain schedule.
+Please see the [Domains](#domains) reference for details on these.
+It has the following attributes:
+
+* `start_index` (optional, defaults to 0) defines the 'start index' of the schedule
+* `index_shift` (optional) the offset of the Microkit domain schedule within
+                the kernel's domain schedule array.
```
dreamliner787-9: Could you please elaborate on this a bit more? As someone who isn't familiar with the domain scheduler I don't understand why we need this. If `index_shift` is optional then how does the offset of the domain schedule within the kernel's data structure get worked out, is it just equals to the ID or something? Is this similar to the "index" in the RFC?

tool/microkit/tests/test.rs
```diff
@@ -1223,6 +1232,104 @@ mod channel {
     }
 }
 
+#[cfg(test)]
+mod domains {
+    use super::*;
```
dreamliner787-9, line 1237: Do you have time to add tests for valid cases too? Especially cases that are not covered by the example, e.g. specifying ticks instead of us, etc

dreamliner787-9 (commented):
Just nitpicks from me.

docs/manual.md
midnightveil: The initialiser adds this shift to all the other numbers. It's something that was added in the initialiser. I don't actually know *why* this was added.

midnightveil: Right, this was why: https://github.com/seL4/capdl/pull/81#issuecomment-4070884008

... which, now that I think about it, means it should default to 1, which the rust-seL4 work doesn't do.

EDIT: no, nevermind it shouldn't

tool/microkit/tests/test.rs
midnightveil: invalid shifts and stuff is tested. I don't have heaps of time right now.

docs/manual.md
midnightveil: (done, updated the comment too)

midnightveil: Actually, unless we give the domain cap to any PD, I think these are both useless from the Microkit side.

(but we should also rename start_index/whatever to the set_start etc)

midnightveil: I suppose I will keep them.

docs/manual.md
midnightveil: Back ref: https://github.com/seL4/capdl/pull/95#issuecomment-5019341312

Furao: Makes sense to leave it as is.

docs/manual.md
Furao: thanks for the info.

tool/microkit/src/main.rs
midnightveil: I should document this.

midnightveil:
I've tested this works locally on RISCV, AArch64, and x86 so once CI is done I will merge this.

midnightveil (approved):

PR #454, opened 2026-03-26
Fix IOAPIC Polarity.

cazb2:
The kernel syscall for configuring ioapic pins bitwise or's the polarity value passed in the syscall. If a user wants a low polarity pin they pass the value 1, and if a user wants a high polarity pin they pass 0.
https://github.com/seL4/seL4/blob/82a1a8f48f3d16fecb5398be9200007d59675ff9/src/plat/pc99/machine/ioapic.c#L190-L194

Currently the microkit tool does the opposite i.e. passing 0 for  a low polarity configuration and passing 1 for a high polarity configuration. The enum definition is what this code changes. The value is actually used here:
https://github.com/seL4/microkit/blob/cbc4c9dbba20787badaf90cc852bb18064564d48/tool/microkit/src/capdl/irq.rs#L73-L87

### DIFF
--- a/tool/microkit/src/sel4.rs
+++ b/tool/microkit/src/sel4.rs
@@ -573,15 +573,15 @@ impl X86IoapicIrqTrigger {
 #[repr(u64)]
 #[derive(Debug, Copy, Clone, Eq, PartialEq)]
 pub enum X86IoapicIrqPolarity {
-    LowTriggered = 0,
-    HighTriggered = 1,
+    HighTriggered = 0,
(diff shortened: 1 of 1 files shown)


Ivan-Velickovic:
@cazb2 I think you are right but when making a PR like this you should point to the actual kernel source code or documentation that backs up your argument. When I checked the manual it didn't make it clear what argument values relate to what actual polarity, so please point to source code and explain why the Microkit tool is doing the wrong thing.

PR #455, opened 2026-03-26
tool: abstract functions in Microkit Tool for use as a library

sfulham:
In order to allow use of Microkit Tool as a library, SDF structures need constructors, and their validation moved from the XML parsing step. Additionally, the `main` entrypoint contains a lot of parts that are useful for other applications, and can also be abstracted out to their own functions.

TODO:
- [ ] Move SDF validation into constructors
- [x] Move kernel configuration, and final binary image generation into their own functions.


Ivan-Velickovic:
@sfulham I think this should be split into multiple PRs. I think we should start with just making the tool accessible as a library first, then restructure `sdf.rs` for when you don't want to generate the system file directly.

I've fixed the overflow issue on main (https://github.com/seL4/microkit/pull/468) so you can just rebase on main.

PR #464, opened 2026-04-01
aarch64: Restrict address mappings in loader.rs

bruelc:
Hello,

This is a rework of the existing support committed in elfloader:
https://github.com/seL4/seL4_tools/pull/244
and discussed here:
https://lists.sel4.systems/hyperkitty/list/devel@sel4.systems/thread/DJNDO5CUQBKGIA4SQHXCNQ3T6SKZEY4A/

This patch limits the 1:1 address mapping in the loader to cover only the necessary blocks using level-2 translation tables with 2MB blocks. The UART MMIO address is now exposed as a global uart_addr variable.

This prevents speculative accesses to device or secure memory, which could otherwise cause faults or unwanted side effects.

This has been tested only on STM32MP2. Testers on other platforms would be greatly appreciated.

Thank you for your review.

### DIFF
--- a/tool/microkit/src/loader.rs
+++ b/tool/microkit/src/loader.rs
@@ -487,20 +487,64 @@ impl<'a> Loader<'a> {
         let (boot_lvl0_upper_addr, boot_lvl0_upper_size) = elf
             .find_symbol("boot_lvl0_upper")
             .expect("Could not find 'boot_lvl0_upper' symbol");
+        let (boot_lvl2_lower_addr, boot_lvl2_lower_size) = elf
+            .find_symbol("boot_lvl2_lower")
+            .expect("Could not find 'boot_lvl2_lower' symbol");
+        let (start_addr, _) = elf
+            .find_symbol("_loader_start")
+            .expect("Could not find 'loader_start' symbol");
+        let (end_addr, _) = elf
+            .find_symbol("_loader_end")
+            .expect("Could not find 'loader_end' symbol");
+
+        if Aarch64::lvl1_index(start_addr) != Aarch64::lvl1_index(end_addr) {
+            panic!("We only map 1GiB, but elfloader paddr range covers multiple GiB");
+        }
 
         let mut boot_lvl0_lower: [u8; PAGE_TABLE_SIZE] = [0; PAGE_TABLE_SIZE];
         boot_lvl0_lower[..8].copy_from_slice(&(boot_lvl1_lower_addr | 3).to_le_bytes());
 
         let mut boot_lvl1_lower: [u8; PAGE_TABLE_SIZE] = [0; PAGE_TABLE_SIZE];
-        for i in 0..512 {
+
+        // map optional UART MMIO in l1 1GB page, only available if CONFIG_PRINTING
+        if let Ok((uart_addr, _)) = elf.find_symbol("uart_addr") {
+            let data = elf
+                .get_data(uart_addr, 8)
+                .expect("uart_addr not initialized");
+            let uart_base = u64::from_le_bytes(data[0..8].try_into().unwrap());
+
+            let lvl1_idx = Aarch64::lvl1_index(uart_base);
             #[allow(clippy::identity_op)] // keep the (0 << 2) for clarity
-            let pt_entry: u64 = ((i as u64) << AARCH64_1GB_BLOCK_BITS) |
+            let pt_entry: u64 = ((lvl1_idx as u64) << AARCH64_1GB_BLOCK_BITS) |
                 (1 << 10) | // access flag
                 (0 << 2) | // strongly ordered memory
-                (1); // 1G block
+                (1 << 0); // 1G block
+            let start = 8 * lvl1_idx;
+            let end = 8 * (lvl1_idx + 1);
+            boot_lvl1_lower[start..end].copy_from_slice(&pt_entry.to_le_bytes());
+        }
+
+        let mut boot_lvl2_lower: [u8; PAGE_TABLE_SIZE] = [0; PAGE_TABLE_SIZE];
+
+        // 1GB lvl1 Table entry
+        let pt_entry = (boot_lvl2_lower_addr | 3).to_le_bytes();
+        let lvl1_idx = Aarch64::lvl1_index(start_addr);
+        let start = 8 * lvl1_idx;
+        let end = 8 * (lvl1_idx + 1);
+        boot_lvl1_lower[start..end].copy_from_slice(&pt_entry);
+
+        // map the loader 1:1 access into 2MB lvl2 Block entries for a 4KB granule
+        let lvl2_idx = Aarch64::lvl2_index(start_addr);
+        for i in lvl2_idx..=Aarch64::lvl2_index(end_addr) {
+            let entry_idx = (i - Aarch64::lvl2_index(start_addr)) << AARCH64_2MB_BLOCK_BITS;
+            let pt_entry: u64 = (entry_idx as u64 + start_addr) |
+                (1 << 10) | // Access flag
+                (3 << 8) | // Sharable
+                (0 << 2) |
+                (1 << 0); // 2M block
             let start = 8 * i;
             let end = 8 * (i + 1);
-            boot_lvl1_lower[start..end].copy_from_slice(&pt_entry.to_le_bytes());
+            boot_lvl2_lower[start..end].copy_from_slice(&pt_entry.to_le_bytes());
         }
 
         let boot_lvl0_upper: [u8; PAGE_TABLE_SIZE] = [0; PAGE_TABLE_SIZE];
@@ -538,6 +582,7 @@ impl<'a> Loader<'a> {
             (boot_lvl0_upper_addr, boot_lvl0_upper_size, boot_lvl0_upper),
             (boot_lvl1_upper_addr, boot_lvl1_upper_size, boot_lvl1_upper),
             (boot_lvl2_upper_addr, boot_lvl2_upper_size, boot_lvl2_upper),
+            (boot_lvl2_lower_addr, boot_lvl2_lower_size, boot_lvl2_lower),
         ]
     }
 }
--- a/tool/microkit/src/loader.rs
+++ b/tool/microkit/src/loader.rs
@@ -498,7 +498,7 @@ impl<'a> Loader<'a> {
             .expect("Could not find 'loader_end' symbol");
 
         if Aarch64::lvl1_index(start_addr) != Aarch64::lvl1_index(end_addr) {
-            panic!("We only map 1GiB, but elfloader paddr range covers multiple GiB");
+            panic!("We only map 1GiB, but loader paddr range covers multiple GiB");
         }
 
         let mut boot_lvl0_lower: [u8; PAGE_TABLE_SIZE] = [0; PAGE_TABLE_SIZE];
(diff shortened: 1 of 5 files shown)


tool/microkit/src/loader.rs
```diff
@@ -485,20 +485,65 @@ impl<'a> Loader<'a> {
...
+            .find_symbol("boot_lvl2_lower")
+            .expect("Could not find 'boot_lvl2_lower' symbol");
+
+        let (text_addr, _) = elf
+            .find_symbol("_text")
+            .expect("Could not find 'text' symbol");
+        let (end_addr, _) = elf
+            .find_symbol("_bss_end")
+            .expect("Could not find 'end' symbol");
+
```
midnightveil, line 497: I'd add a brief comment just saying what this is doing.

Specifically: it's mapping in just the loader itself; not all of RAM.

It might make sense to add extra linker script symbols for 'loader_start' and 'loader_end' rather than relying on text/bss end implicitly being at start/end.

Also `text_addr` -> `start_addr`

tool/microkit/src/loader.rs
```diff
@@ -485,20 +485,65 @@ impl<'a> Loader<'a> {
...
+        if Aarch64::lvl1_index(text_addr) != Aarch64::lvl1_index(end_addr) {
+            eprintln!("ERROR: We only map 1GiB, but elfloader paddr range covers multiple GiB");
```
midnightveil: I think we tend to use `panic!("INTERNAL: <message>")` here.

I'd also reword this: specifically the issue is the microkit loader (not elfloader) crosses a GiB boundary [or really that it requires multiple level 1 page tables].

It likely makes sense to do this change to both AArch64 and Riscv64, no sense it being different for either.

tool/microkit/src/loader.rs
```diff
@@ -485,20 +485,65 @@ impl<'a> Loader<'a> {
...
 
         let mut boot_lvl1_lower: [u8; PAGE_TABLE_SIZE] = [0; PAGE_TABLE_SIZE];
-        for i in 0..512 {
+
+	// map optional UART MMIO in l1 1GB page, only available if CONFIG_PRINTING
+        if let Ok((uart_addr, _)) = elf.find_symbol("uart_addr") {
+	    let data = elf
+	        .get_data(uart_addr, 8)
+		.expect("uart_addr not initialized");
+	    let uart_base = u64::from_le_bytes(data[0..8].try_into().unwrap());
```
midnightveil: (Not something you need to fix, just making this comment so I can link to it later):

It'd be nice if the kernel exposed the virtual address of where it wants its UART so we can map that in and get [this feature working again](https://github.com/seL4/seL4/pull/1513).

tool/microkit/src/loader.rs
```diff
@@ -485,20 +485,65 @@ impl<'a> Loader<'a> {
...
+        boot_lvl1_lower[start..end].copy_from_slice(&pt_entry);
+
+	let lvl2_idx = Aarch64::lvl2_index(text_addr);
+
+        for i in lvl2_idx ..= Aarch64::lvl2_index(end_addr) {
+            let entry_idx = (i - Aarch64::lvl2_index(text_addr)) << AARCH64_2MB_BLOCK_BITS;
+            let pt_entry: u64 = (entry_idx as u64 + text_addr) |
+       	        (1 << 10) | // access flag
+                (3 << 8)  | // inner sharable
+	        (3 << 2)  | // MT_NORMAL memory
```
midnightveil: alignment here and elsewhere is off

tool/microkit/src/loader.rs
```diff
@@ -485,20 +485,65 @@ impl<'a> Loader<'a> {
...
                 (0 << 2) | // strongly ordered memory
                 (1); // 1G block
+            let start = 8 * lvl1_idx;
+            let end = 8 * (lvl1_idx + 1);
+            boot_lvl1_lower[start..end].copy_from_slice(&pt_entry.to_le_bytes());
+        }
+
+        let mut boot_lvl2_lower: [u8; PAGE_TABLE_SIZE] = [0; PAGE_TABLE_SIZE];
+
+        let pt_entry = (boot_lvl2_lower_addr | 3).to_le_bytes();
```
midnightveil, line 528: can we be consistent with the other examples and split out the meanings of the `3` flags into the individual bits?

midnightveil (approved):
Fine. Just some minor code cleanups to do.

tool/microkit/src/loader.rs
bruelc: 3 means the L1 entry points to a 2MB block instead of a 1GB  table, it is not individual bits (like access flags, or others)

tool/microkit/src/loader.rs
bruelc: I am not familiar with the Riscv64 MMU tables, and I do not have access to a testing platform. Is there any volunteer to do the adaptation?

tool/microkit/src/loader.rs
bruelc: This this is a 2 bit  marker for the 1GB L1 table entry here, not individual bits

tool/microkit/src/loader.rs
midnightveil: That's fair. We're planning on changing a few things after these are good to go, anyway, and having to retest.

tool/microkit/src/loader.rs
midnightveil: Ah, right, we do `|3` in a bunch of places below. I'll change that at a later point.

tool/microkit/src/loader.rs
```diff
@@ -485,20 +485,63 @@ impl<'a> Loader<'a> {
...
+        let lvl1_idx = Aarch64::lvl1_index(start_addr);
+        let start = 8 * lvl1_idx;
+        let end = 8 * (lvl1_idx + 1);
+        boot_lvl1_lower[start..end].copy_from_slice(&pt_entry);
+
+	// map the loader 1:1 access into 2MB blocks
```
midnightveil: the indentation on this line is still off

tool/microkit/src/loader.rs
bruelc: Thanks, better to handle it globally yes. I’ll add a comment since “3” means different things depending on the granule (page or table)

tool/microkit/src/loader.rs
bruelc: thank you. I noticed the CI doesn’t build for riscv64 without uart_addr. I’ll guard it to fix the build before your changes.

tool/microkit/src/loader.rs
```diff
@@ -485,20 +485,64 @@ impl<'a> Loader<'a> {
...
+        boot_lvl1_lower[start..end].copy_from_slice(&pt_entry);
+
+        // map the loader 1:1 access into 2MB lvl2 Block entries for a 4KB granule
+        let lvl2_idx = Aarch64::lvl2_index(start_addr);
+        for i in lvl2_idx ..= Aarch64::lvl2_index(end_addr) {
+            let entry_idx = (i - Aarch64::lvl2_index(start_addr)) << AARCH64_2MB_BLOCK_BITS;
+            let pt_entry: u64 = (entry_idx as u64 + start_addr) |
+                (1 << 10) | // Access flag
+                (3 << 8) | // Sharable
+                (3 << 2) | // MT_NORMAL memory
```
midnightveil: Looking at this more. Did you intentionally make this `3 << 2`? That would correspond to `MT_NORMAL_NC`, not `MT_NORMAL`.

bruelc: Argh no! That’s an early debugging trace… thanks for noticing.

bruelc: Fixed: Note that I also had to update https://github.com/seL4/microkit/pull/465 to flush the cache before jumping to the secondary core

midnightveil (approved):

midnightveil:
~~OK, I'll note that this appears to have broken most of our boards even in non-SMP mode. They all seem to error out with: (e.g. for the MaaXboard)~~

EDIT: no this was just a dirty build. rerunning it everything works in debug mode (for non-smp)

<details><summary>log</summary>

```
Bootstrapping kernel
available phys memory regions: 1
  [40000000..c0000000)
reserved virt address space regions: 2
  [8040000000..8040246000)
  [8040246000..804029e000)
Booting all finished, dropped to user space
panicked at /home/julia/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/rkyv-0.8.12/src/option.rs:38:17:
called `ArchivedOption::unwrap()` on a `None` value
(aborted)
Found thread has no fault handler while trying to handle:
unknown fault
in thread 0x80bfc07400 "rootserver" at address 0x2219e8
With stack:
0x24dc80: 0x221bd8
0x24dc88: 0x24dcd8
0x24dc90: 0x221b5c
0x24dc98: 0x221b5c
0x24dca0: 0x221b30
0x24dca8: 0x24dcd8
0x24dcb0: 0x22d550
0x24dcb8: 0x24f2ec
0x24dcc0: 0x0
0x24dcc8: 0x2055c8
0x24dcd0: 0x67
0x24dcd8: 0x24dcc8
0x24dce0: 0x205600
0x24dce8: 0x1
0x24dcf0: 0x0
0x24dcf8: 0x221118
```

</details>

~~This is just building the hello world example in debug mode.~~
~~I'm somewhat unsure how, though.~~

Ivan-Velickovic:
Rebased on main + pushed a temporary fix for the Raspberry Pi 4B so we don't regress that platform with this change. Also changed the mapping code so that it still maps the loader as strongly ordered, we did this because we don't want to worry about flushing caches for this PR (without flushing caches we'd still have issues on Raspberry Pi 4B).

Ivan-Velickovic:
> temporary fix for the Raspberry Pi 4B 

I should add that the reason I say 'temporary' is since @midnightveil has in-progress patches to re-write all of this loader mapping code to be easier to read/better structured so we'll revisit as part of that.

midnightveil:
I'm also going to note down here why this was necessary, as it's not completely obvious from the fact, as the 1GB mapping the loader used was already `DEVICE_nGnRnE` ('strongly ordered').

---

See this discussion: https://github.com/seL4/seL4_tools/pull/244#discussion_r2800315222

The conclusion is that ARM seems to allow speculative reads of *instructions* even in `DEVICE_nGnRnE` memory unless the XN bit is set. This contrasts to data loads/stores which are not (speculatively) permitted.

References in ARM ARM DDI 0487L.B, under §D8.4.4 "Effects on instruction execution permissions and restrictions on instruction fetch":

> $I_{PTMPK}$ If Device memory is not execute-never, then speculative instruction fetch from that memory region is permitted.
> $I_{ZZSZZ}$ For all Exception levels from which a read-sensitive memory region can be accessed, to avoid the possibility of a speculative fetch affecting that memory region, software is expected to define that memory region as execute-never.

In contrast, under §B2.10.2 "Device memory", it states:

> Speculative data accesses are not permitted to any memory location with any Device memory attribute. This means that each memory access to any Device memory type must be one that would be generated by a simple sequential execution of the program.
> [...]
> Hardware does not prevent speculative instruction fetches from a memory location with any of the Device memory attributes unless the memory location is also marked as execute-never for all Exception levels.

bruelc:
Just noticed that the commit log was copied  from the CAmKes tool loader and the sentence:

> Don't map device memory at all, therefore, printf must not be used in the elfloader after the MMU is enabled.

doesn't apply for Microkit elf loader.

Could you please amend the committed log and remove this sentence ?

sorry about this, thank you.

midnightveil:
Generally we can't easily force push to main, so unfortunately that will have to stay.

No big issue though, it's OK :)

Issue #463, opened 2026-03-31
broken atomic lock when starting multiple cores


bruelc:
Hello,

Synchronization between cores in SMP mode relies on STLR/LDAR instructions to implement release/acquire semantics on the print_lock flag. These instructions depend on proper cache coherency and on memory being mapped as Normal, Inner Shareable.

Therefore, all cores must have the MMU (and caches) enabled with consistent attributes in order to use locks across cores0

In loader.c:main(), the MMU on logical core 0 is enabled after the secondary cores have already been started.
As a result, core 0 does not participate in the coherency domain when the other cores begin execution. Consequently, updates to the print_lock flag performed by secondary cores are not observed by core 0, leading to an infinite loop.

Workarounds are:  Disabling the data cache via SCTLR.C, or using  Normal_NC in MAIR for the translation tables

The MMU on core 0 should be enabled before releasing secondary cores. 

_Originally posted by @bruelc in https://github.com/seL4/microkit/issues/446#issuecomment-4156815830_

bruelc:
Only reproducible on STMP32MP2 port, but probably latent on other aarch64 platforms

Indanz:
The lock should never be used with MMU disabled, just like any other shared data may not be used by MMU disabled code.

PR #465, opened 2026-04-01
loader,smp: Enable caches on all cores for shareability

bruelc:
Hello,

This fixes an SMP boot issue observed on the STM32MP2 platform: core 0 does not start because print_lock is never seen as released by the secondary cores. The issue is related to cachability, which is not yet enabled at this stage.

Ensure all cores are in the same coherency domain before using cached shared data by enable MMU on all cores before using shared datas
Thus moved arch_mmu_enable() out of start_kernel().

### DIFF
--- a/loader/src/aarch64/cpus.c
+++ b/loader/src/aarch64/cpus.c
@@ -8,6 +8,7 @@
 #include <stdint.h>
 
 #include "smc.h"
+#include "../arch.h"
 #include "../cpus.h"
 #include "../cutil.h"
 #include "../loader.h"
@@ -110,6 +111,14 @@ void arm_secondary_cpu_entry(int logical_cpu, uint64_t mpidr_el1)
 
     plat_save_hw_id(logical_cpu, mpidr_el1);
 
+    int r = arch_mmu_enable(logical_cpu);
+    if (r != 0) {
+        LDR_PRINT("ERROR", logical_cpu, "failed to enable MMU: ");
+        puthex32(r);
+        puts("\n");
+        goto fail;
+    }
+
     start_kernel(logical_cpu);
 
 fail:
@@ -174,6 +183,10 @@ int plat_start_cpu(int logical_cpu)
     sp[0] = logical_cpu;
     /* zero out what was here before */
     sp[1] = 0;
+
+    /* clean up the cache line containing sp before use by secondary core */
+    asm volatile("dc cvac, %0" :: "r"(sp) : "memory");
+    asm volatile("dsb sy" ::: "memory");
 
     /* Arguments as per 5.1.4 CPU_ON of the PSCI spec.
 
--- a/loader/src/loader.c
+++ b/loader/src/loader.c
@@ -100,15 +100,6 @@ static int print_lock = 0;
 
 void start_kernel(int logical_cpu)
 {
-    LDR_PRINT("INFO", logical_cpu, "enabling MMU\n");
-    int r = arch_mmu_enable(logical_cpu);
-    if (r != 0) {
-        LDR_PRINT("ERROR", logical_cpu, "failed to enable MMU: ");
-        puthex32(r);
-        puts("\n");
-        for (;;) {}
-    }
-
     LDR_PRINT("INFO", logical_cpu, "jumping to kernel\n");
 
 #ifdef CONFIG_PRINTING
@@ -174,6 +165,13 @@ int main(void)
     LDR_PRINT("INFO", 0, "active CPUs to start: ");
     puthex32(plat_get_active_cpus());
     puts("\n");
+
+    r = arch_mmu_enable(0);
+    if (r != 0) {
+        LDR_PRINT("ERROR", 0, "failed to enable MMU: ");
+        puthex32(r);
+        fail();
+    }
 
     for (int cpu = 1; cpu < plat_get_active_cpus(); cpu++) {
         r = plat_start_cpu(cpu);


midnightveil (approved):
This seems like it would be correct to me, but I'd like to confirm it against the ARM ARM first. 

What is the shareability of these mappings, do you know? Hopefully we set it up as ISH.

Maybe I misunderstood @KurtWu10 but I was under the impression that this would be fine as we only disabled the MMU & L1 cache (or actually thinking about it I guess we also disable the cache itself. I was asking about how it works it you just disable the MMU, not the cache too)).

This probably would affect spin table booting too, though I'm inclined to leave it as is as nothing but raspberry pi uses it (looking at Linux it does a cache clean & invalidate to PoC there)

(Not something to fix here) this does make me realise I could have switched to the kernel page tables instead of disabling the uboot ones, as that would stop the loader loading things over the current page tables since it knows to not load over itself. Although we probably want to clean caches then.

KurtWu10:
Initially, I thought this was correct, but now I am not sure whether enabling MMU is required for cache coherency.

According to the Arm ARM [_D8.2.12 The effects of disabling an address translation stage_](https://developer.arm.com/documentation/ddi0487/maa/-Part-D-The-AArch64-System-Level-Architecture/-Chapter-D8-The-AArch64-Virtual-Memory-System-Architecture/-D8-2-Translation-process/-D8-2-12-The-effects-of-disabling-an-address-translation-stage), assuming that the effective value of `HCR_EL2.DC` is 0, it is said that
> ($\textsf{R}_\textsf{WFZPW}$) ... the stage 1 translation assigns one of the following attributes to memory accesses:
> - For a data access, the __Device__-nGnRnE memory attribute.

then according to [_B2.3 Ordering requirements defined by the formal concurrency model_](https://developer.arm.com/documentation/ddi0487/maa/-Part-B-The-AArch64-Application-Level-Architecture/-Chapter-B2-The-AArch64-Application-Level-Memory-Model/-B2-3-Ordering-requirements-defined-by-the-formal-concurrency-model),
> In this section ..., all of the following are assumed:
> - One of the following applies:
>   - ...
>   - All PEs are part of the Full System domain and all Memory Effects are to Normal, Inner Non-Cacheable, Outer Non-Cacheable memory or __Device__ memory as long as all barriers specify the Full System domain and there are no TLB invalidations.
> - All Memory Effects are to Normal, Inner Write-Back, Outer Write-Back memory. Behaviors in Ordering requirements defined by the formal concurrency model also apply to Memory Effects to memory with attributes other than Normal, Inner Write-Back, Outer Write-Back (for example, __Device__ or Inner Write-Through, Outer Write-Through) as long as the behavior of the instruction which generated the Memory Effect is guaranteed for the memory type.

My interpretation is, with MMU disabled, data memory accesses use the device attribute, and cache coherency for them is provided for load-acquire and store-release instructions.

---
The reasoning above assumes that the effective value of `HCR_EL2.DC` is 0. This might not be always true, as only the `switch_to_el1` function [sets](https://github.com/seL4/microkit/blob/2.2.0/loader/src/aarch64/util64.S#L220-L221) the value. By default, the field is said to reset to an unknown value.

bruelc:
> My interpretation is, with MMU disabled, data memory accesses use the device attribute, and cache coherency for them is provided for load-acquire and store-release instructions.

I think you are correct here: cacheability and coherency can indeed apply regardless of MMU, as long as the cores are in the same shareability domain and cache attributes are consistent.

My mistake in the description was that I confused “enabling the caches” with “enabling the MMU”, because both are currently done inside enable_mmu().

A possible refinement would be to separate cache activation from the enable_mmu() with a new function called before accessing shared data.  Which makes now 2 possible fixes :

- Keep the existing patch, update the comments accordingly, and retitle the PR to mention cache instead of MMU
- Keep the current implementation but split enable_mmu() with a new enable_cache() called before accessing the sync datas.

I am fine with either option

Thank you for raising this inconsistency!

bruelc:
> This seems like it would be correct to me, but I'd like to confirm it against the ARM ARM first.
> 
> What is the shareability of these mappings, do you know? Hopefully we set it up as ISH.

In https://github.com/seL4/microkit/pull/464 I set inner shareability in the memory translation attributes

> 
> Maybe I misunderstood @KurtWu10 but I was under the impression that this would be fine as we only disabled the MMU & L1 cache (or actually thinking about it I guess we also disable the cache itself. I was asking about how it works it you just disable the MMU, not the cache too)).

That the point, we could disable the MMU and only enable the cache. But in this situation, I don't know what the default shareability settings. I tested on STM32MP and this works, but I'm afraid we could fall into SoC or undefined behavior.  

Making sure attributes in the translation tables are set (thus with MMU enabled) is IMHO the best way to avoid ambiguity 

> 
> This probably would affect spin table booting too, though I'm inclined to leave it as is as nothing but raspberry pi uses it (looking at Linux it does a cache clean & invalidate to PoC there)
> 
> (Not something to fix here) this does make me realise I could have switched to the kernel page tables instead of disabling the uboot ones, as that would stop the loader loading things over the current page tables since it knows to not load over itself. Although we probably want to clean caches then.

midnightveil:
Yes, I agree. I think it makes sense to turn MMU and caches on.

midnightveil (approved):

midnightveil:
FYI, you'll need to fix the git commit title length, since there is gitlint setup on seL4 repos. No more than 50 chars in the title and 70 in the body. Same for the other PR.

loader/src/aarch64/cpus.c
```diff
@@ -175,6 +184,10 @@ int plat_start_cpu(int logical_cpu)
     /* zero out what was here before */
     sp[1] = 0;
 
+    /* clean up the cache line containing sp before use by secondary core */
+    asm volatile("dc cvac, %0" :: "r"(sp) : "memory");
```
midnightveil, line 188: I guess this is an argument that we should make the loader use outer shareable instead of inner shareable, then I think this should be fine.

Indanz: I don't think that would solve any coherency problems when the caches are disabled on the secondary cores.

bruelc: > I don't think that would solve any coherency problems when the caches are disabled on the secondary cores.

Caches are enabled early in arm_secondary_cpu_entry(). Before, the only data that must be coherent is the parameter read from the stack by arm_secondary_cpu_entry_asm() (without this I read 0 as logical_cpu parameter, which fails). Instructions should be fine.

We could conservatively clean the entire cache here, but that seems like overkill. Am I missing anything, or do you see any other possible source of incoherence before caches are enabled in this sequence?

(edit: arm_secondary_cpu_entry_asm() read the stack, not the PSCI call)

midnightveil: Apologies, I didn't notice this was for the stack, which probably needs a cache clean here yes.


What I had thought of was that the print_lock is a variable that exists across cores that won't be in the same shareability domain. I probably need to recheck the manual here, it might be that a "dsb sy" (full system barrier) would be OK when the memory is mapped ISH as opposed to OSH,  but I also don't know if that's what the compiler atomics expect and whether the instructions they emit work for that case.

Admittedly this isn't an issue with this PR at all...

bruelc: No problem. It's best to discuss the details now :). I think the print_lock exclusive accesses are in the same shareability domain, because they are accessed after enable_mmu() on all cores, with the same MMU/cache attributes. Also, we enable both ISH and OSH shareability in the translation tables

midnightveil: > Also, we enable both ISH and OSH shareability in the translation tables

I'm not sure what you mean by this?

---

Anyway, it's a Friday evening and I'm mixing everything up 🙃. For some reason I'm thinking that ISH means "non-shareable"; I'm mixing things up.  ISH should be fine for cores; all cores are (usually) in the same inner shareable domain; at least that's what the seL4 SMP code assumes. Non-shareable would be a different matter, it was just the comment in here that the shareability matches that of seL4 sees to be not true as seL4 is NSH on the non-SMP configurations.

tldr: no issue I mixed things up .

Indanz: > > I don't think that would solve any coherency problems when the caches are disabled on the secondary cores.
> 
> Caches are enabled early in arm_secondary_cpu_entry(). Before, the only data that must be coherent is the parameter read from the stack by arm_secondary_cpu_entry_asm() (without this I read 0 as logical_cpu parameter, which fails). Instructions should be fine.
> 
> We could conservatively clean the entire cache here, but that seems like overkill. Am I missing anything, or do you see any other possible source of incoherence before caches are enabled in this sequence?

My comment was in reply to Julia's suggestion to have the memory as outer-shareable instead of inner-shareable, not your cache clean code. The code you have is to handle the case before caches are enabled, so it doesn't matter whether it's mapped inner or outer shareable on the other core.

Anyway, considering you know that the other core starts at the correct exception level, and the MMU is already configured, you could just enable it immediately at the start of `arm_secondary_cpu_entry_asm` instead. That seems better than ad-hoc cache flushing.

That, or start the other cores with cache disabled. But this weird mix of having cache enabled on one core and disabled on the others seems precarious.

bruelc: > Anyway, considering you know that the other core starts at the correct exception level, and the MMU is already configured, you could just enable it immediately at the start of `arm_secondary_cpu_entry_asm` instead. That seems better than ad-hoc cache flushing.

Do you mean enabling the caches (not mmu) in arm_secondary_cpu_entry_asm? I just tried it, but it doesn’t work. It seems that ARM requires cache lines to be invalidated first to prevent the use of stale data. Thus even with cache enabled in secondary core we still need to clean it first. If we don’t want a single word ad-hoc fix, all lines could be cleaned but that’s a shame.

@Julia:
> > Also, we enable both ISH and OSH shareability in the translation tables
> 
> I'm not sure what you mean by this?

Sorry, I thought that enabling inner shareable implied outer shareable. Forget this. But anyway since we are discussing code before enable_mmu() in core1 that does not matter.

Indanz: > Do you mean enabling the caches (not mmu) in arm_secondary_cpu_entry_asm?

Well, I would enable the MMU too while you're there.

> I just tried it, but it doesn’t work.

Did you do this early enough and without reading any stack data?

> It seems that ARM requires cache lines to be invalidated first to prevent the use of stale data. Thus even with cache enabled in secondary core we still need to clean it first.

I thought if caches were disabled, they still participate in coherency, so this doesn't make sense to me.

But if you only enabled caches, but not the MMU, then it makes sense that it doesn't work, as e.g. ASIDs don't match.

> If we don’t want a single word ad-hoc fix, all lines could be cleaned but that’s a shame.

You introduced this problem yourself by enabling caches on the first core before starting the other cores, as a work-around for the UART lock problem.

I propose either getting it working with caching and MMU enabled immediately on secondary cores, or fixing the UART code, instead of having dodgy stuff like mixed caching enabled between cores.

bruelc: > > Do you mean enabling the caches (not mmu) in arm_secondary_cpu_entry_asm?
> 
> Well, I would enable the MMU too while you're there.

logical_cpu is used by arch_mmu_enable() for print outputs, so we still need to pop it before enabling the MMU. 

> 
> > I just tried it, but it doesn’t work.
> 
> Did you do this early enough and without reading any stack data?
> 

sure, :

.macro enable_caches sctlr tmp
    mrs     \tmp, \sctlr
    orr     \tmp, \tmp, #(1 << 2)
    orr     \tmp, \tmp, #(1 << 12)
    msr     \sctlr, \tmp
    isb
 .endm

BEGIN_FUNC(arm_secondary_cpu_entry_asm)
    mov sp, x0

    enable_caches sctlr_el1, x0

    /* grab the core ID off the stack */
    ldr x0, [sp]
    mrs x1, mpidr_el1
    /* x0 and x1 are arguments */
    b arm_secondary_cpu_entry
END_FUNC(arm_secondary_cpu_entry_asm)

<snip>

Additively, the behavior of cache synchronization with the PoU are quite difficult to read. 
But I found this in the ARM ARM armv8-a :

"
An implementation can require the use of a specific cache initialization routine to invalidate its storage array
before caching is enabled. The exact form of any required initialization routine is IMPLEMENTATION DEFINED,
"

Which should be sufficient to not trust the cache content when we enable it.

> 
> I propose either getting it working with caching and MMU enabled immediately on secondary cores, 

This requires some larger refactoring to avoid the use of logical_cpu early in arch_mmu_enable().  But worse I'm not even sure this is safe if the cache status in unpredictable at the time of initialization

Indanz: All in all it sounds it's much easier to keep the current code that enables the other cores with caches disabled and abandon this PR, and to fix the UART code, which is non-critical only anyway.

What about adding a cache clean to the print_lock code instead?

midnightveil: We likely wanted to keep the MMU+caches enabled for the main core, FWIW: https://github.com/seL4/microkit/issues/466. Having it off was a sort of hack.

FWIW: I thought it was responsibility of the TF-A/PSCI interface to handle all the required cache management. Maybe that's just power off, not power on?

---

Also, feel free to rearrange any of the prints and such, they're for debugging. If it makes the code simpler it's fine to remove some in certain locations.

bruelc: > All in all it sounds it's much easier to keep the current code that enables the other cores with caches disabled and abandon this PR, and to fix the UART code, which is non-critical only anyway.
> 
> What about adding a cache clean to the print_lock code instead?

This would be a workaround. The root cause for the print_lock code is that the lock was taken with cache off and released with cache on. I think doing a cache clean when releasing would work (I haven't tried) but this should not be required by the ACQUIRE/RELEASE semantic.

bruelc: I tried to enable the MMU early and it is OK, but we lose the traces and sanity from arch_mmu_enable()

To summarize when MMU and caches are enabled on the main core:

0 Enabling caches in arm_secondary_cpu_entry_asm() FAIL
1 Enabling caches and MMU in arm_secondary_cpu_entry_asm(): OK but we lose the ensure_correct_el() sanity, and a few traces
2 Enabling I Bit only and keep Dcache disabled is OK. Sounds like a W/A, no impact on performance
3 Enabling All caches and doing a clean before jumping to secondary core is OK, but Indan is not confortable with  it

At this point, I don’t have further proposals. Any of those would be acceptable to me.

Regards,

Indanz: (From the name I assumed `print_lock` was in UART or generic printing code, but it's a boot serialising lock used to serialise the loader startup code, not just prints.)

> This would be a workaround. The root cause for the print_lock code is that the lock was taken with cache off and released with cache on.

Code with caches on is trying to communicate via memory with code running with caches off. Doing cache clean operations on that piece of memory is not a work-around, but the proper fix if you want to keep this cross-caching-enabled communication.

Whether you do that on `print_lock` or a part of stack is equally work-around-hacky.

Doing it on `print_lock` is safer because it is easy to analyse for correctness.

In the other case, things are a lot more fuzzy, hence my discomfort.

E.g. `arm_secondary_cpu_entry()` calls `plat_save_hw_id()` before enabling the MMU and caching. That's not obviously safe to do. The first core could have a dirty cache line which contains the other core's `cpu_mpidrs[]` entry, and once that line gets evicted it will overwrite the secondary core's write. In this case it's easy to fix by moving the `plat_save_hw_id()` to after enabling the MMU, but you have to review all code for subtle interactions like these. I don't think there is anything else, but it's easy to add something by mistake in the future as this is very non-obvious.

(On AArch64, `plat_get_hw_id()` doesn't seem to be called, so it doesn't matter anyway. But it could be another object adjacent to cpu_mpidrs, which means that unrelated code or toolkit changes can trigger the bug).

In the end, it all relies on the dcache clean by set/way working, which is called when enabling the MMU by `el1/el2_mmu_disable()`. But if that MMU disable call gets removed, things may break in mysterious and very hard to debug ways.

It all seems to be asking for subtle problems in the future, even if it's made to work now. But that's true for the current code as well as this PR, as cores will transition to cache on at some point and then theoretically may touch cache lines containing data used by cache-off booting cores. Except that it's less likely to trigger such problems, as it enables the caching just before jumping to the kernel and the kernel won't touch the loader's data.

But prefetching on another core could load cache lines to L2 that causes uncached writes to (temporarily) disappear when enabling the MMU. This is especially problematic if that write was to the stack and contains e.g. the return address.

In conclusion, my discomfort has expanded and to me the only truly safe approach is to not write any data until the MMU and caching is enabled, including the stack. Basically write the code in such way that it works without a cache flush on the secondary core. This would be a variant of version of 1. If PSCI was used, the started core runs at the same exception level, but that may not be true for spin tables.

bruelc: OK. Enforcing caches enable on both cores before writing (for the print_lock) and before reading (for the stack) is a good choice to make the shared datas robust for once. 
I tried with my initial attempt to be less intrusive, as I thought there was more usages of logical_cpu in ensure_correct_el(), but checking the code it's only for traces. Just refactoring it is manageable. (eventually keeping the debug traces using mpidr_el1 instead of logical_cpu)

Indanz: I'm not saying you should fix this right now, it's an existing problem and Elfloader probably has the same issue of arbitrary L2 data prefetch reads done by other cores being able to "corrupt" the stack, but it's generally not safe to write to your own stack either, not even if you do a clean before enabling the MMU (probably okay if it's a broadcast clean and invalidate, if there are no writes between that and the MMU enable). It's only safe to use the stack for e.g. calls if all called functions have returned at the time that the MMU gets enabled and you don't depend on any uncached writes to stack to persist.

Scenario:

Core 0: Start core 1.
Core 1: Bootup and launch kernel, causing a prefetch read of core 2's cache.
Core 0: Start core 2.
Core 2: Call function that cleans and invalidates cache and enables the MMU.
Core 2: Pops registers, but reads wrong cached data from L2 brought in by core 1.
Core 2: ???

bruelc: as you wish we can handle the early enable_mmu in another follow-up PR if it is not a blocker. 

Note that I have tested only on dual-core, so your scenario above might be unnoticed with the  W/A

Indanz: Actually, if my analysis is correct, it gets triggered by core 0 too:

Core 0: Write data to core 1's stack, bringing it into L1/L2.
Core 0: Clean, but not invalidate cache line (this PR) and start core 1.
Core 1: Call function that cleans and invalidates cache and enables the MMU.
Core 1: Pops registers, but reads wrong cached data from L2 brought in by core 0.
Core 1: ???

Either you got lucky, or the clean and invalidate of the cache clears out L2 too, or I'm wrong.

bruelc: I'm lost. is the scenario you describe  with my dc cvac w/a or with the early enable_mmu fix ?

using my w/a  the sequence is:

Core 0: Write data to core 1's stack, bringing it into L1/L2.
Core 0: Clean, but not invalidate cache line (this PR) and start core 1.
Core 1: Pops registers, reads cached data from L2 brought in by core 0.
Core 1: Call function that cleans and invalidates cache and enables the MMU.

that's probably why I see it OK at my end;

now if there is an issue about the cache not being invalidated after the clean, I can use "dc civac" instead of "dc cvac"

Indanz: > I'm lost. is the scenario you describe is with my dc cvac, w/a or with the early enable_mmu fix ?

I meant this PR as it is at this moment, without early MMU enable or other changes.

My question is why that works, as it triggers the corner case I'm concerned about. (Cached data shadowing uncached writes.)

> using my w/a the sequence is:
[...]
> that's probably why I see it OK at my end;

I don't know what you mean with "w/a", but the pop I mean is the return from the function call, not retrieving the arguments passed from core 0, that's always fine.

If you mean enabling caches without enabling the MMU, that should work for L2 as that's PIPT. Not entirely sure how that interacts with a VIPT L1 cache though, that might be implementation defined.

> now if there is an issue about the cache not being invalidated after the clean, I can use "dc civac" instead of "dc cvac"

No, that would only hide the issue, as speculative reads at the wrong moment have the same effect. Better to always trigger the scenario so it breaks immediately when something changes, instead of once in a blue moon.

Indanz: > My question is why that works, as it triggers the corner case I'm concerned about. (Cached data shadowing uncached writes.)

It should only work if, when returning from the last call that enables the MMU, no (important) registers were pushed to stack and only LR is used to return. If LR got pushed when entering the MMU enable function, on leave, if we pop that value from cache, LR should contain an invalid (cached) value. That's okay as long as we don't do another return, but seems somewhat precarious. E.g. it may only work when compiling with optimisations enabled.

Same for other registers, but with so many registers on AARch64 that's unlikely, so it's mostly the return stack that's of concern.

That's why I said above that writing to stack isn't save with MMU disabled. I don't mean the arguments passed from core 0 to secondary cores, I meant stack writes done by the secondary cores themselves.

bruelc: OK, sorry,  You were talking about the ABI pop, I was still with the 'logical_cpu' pop... I see your point now.

It might be a stupid question, but how can we have a cache conflict between the main core stack and other core's ? they are not at the same address. (not even in the same cache line), and if they did because of cache associativity, one of them would be evicted.

bruelc: >  ... or the clean and invalidate of the cache clears out L2 too, ???

I think this the case : doing a full clean cache enforces what ARM calls PoC,  and this is called from el2_mmu_enable->el2_mmu_disable->flush_dcache

So I see the sequence as your sequence now as

Core 0: Write data to core 1's stack, bringing it into L1/L2.
Core 0: Clean, but not invalidate cache line (this PR) and start core 1.
Core 1: Call function that cleans and invalidates cache and enables the MMU. -> PoC (*)
Core 1: Pops registers, 

(*) Any data cleaned to PoC is guaranteed to be visible

Indanz: > It might be a stupid question, but how can we have a cache conflict between the main core stack and other core's ? they are not at the same address. (not even in the same cache line).

It's not a conflict between the main core's stack and other core's stack. It's a conflict in value between the cached view and the uncached view of the other core's stacks.

> Looks like just doing a full clean cache before jumping to secondary core is the way to enforce what ARM calls PoC

Yes, and that's fine. That's part of what I meant with "retrieving the arguments passed from core 0, that's always fine.", as that's easily made coherent with the cache clean.

The incoherence is introduced by the other core writing any data to uncached memory, while the same cache line is in any other cache. Those writes bypass the cache altogether. The writes that are most concerning are pushes to stack.

If this happens, after enabling the cache, the core will suddenly see the cached value.

Now, I think the reason it works in practice is because `flush_dcache` does a clean *and* invalidate by set/way for all cache levels in `el1/2_mmu_disable`, which is the deepest function. At that point all values stored on stack are made coherent. Very importantly, it doesn't do any other memory writes (directly or via push) between the invalidate and the cache and MMU enable.

If it did just a clean, we should hit this issue and it probably would crash when returning from one of the functions called.

midnightveil (approved):
I think the conclusion from the discussion is that because we do a cache clean & invalidate as part of the enable MMU sequence then there was unlikely to be any issues we encountered?

Approving this: we ran tests with both this and #464 on all our boards and they work.

(Same with our sDDF tests on the boards we use regularly).

Caveat: #464 seems to have broken the Raspberry Pi as the spin table entries are not mapped in, but I have some other loader rework changes to merge after this and I'll fix it with them.

Plan is we'll merge this (&stm32 support) tomorrow morning, hopefully.

midnightveil:
OK. We didn't merge this PR yet (unlike #464), because we currently don't believe it is necessary. Specifically:

- The previous edition of #464 includes mapping the memory for the loader as NORMAL-cacheable. This broke SMP boot on all our boards (not just the STM32). Previously this was OK because all the loader memory was mapped as 'strongly ordered' (`DEVICE_nGnRnE`) and so per §B2.10.2 "Device memory" of ARM ARM DDI 0487 L.b, 

    > Data accesses to memory locations *[marked as device memory]* are coherent for all observers in the system, and correspondingly are treated as being Outer Shareable.

    This meant that around SMP core booting we did not need to perform cache maintenance. (Only I think `DMB`/`DSB` barriers which we do for the Raspberry Pi spin tables but not for writing the stack pointer for the PSCI boot).

- Changing #464 to be strongly ordered device memory again made SMP boot work on all our boards.

Therefore, this seems to be unnecessary from a 'does it work' standpoint on all our platforms (but we currently cannot test STM32).

---

However: in #463 you do note that the `STLR` instructions which the `atomic_store` and `atomic_load` compiler intrinsics compile to are at the very least questionable on DEVICE memory. We *should* probably change this, but it does work on all our platforms. Is this the root cause for this case behind this change, or was it the change to make the loader memory cacheable that was the issue (especially given you needed an extra cache line flush operation here).

bruelc:
> OK. We didn't merge this PR yet (unlike #464), because we currently don't believe it is necessary. Specifically:
> 
> * The previous edition of [aarch64: Restrict address mappings in loader.rs #464](https://github.com/seL4/microkit/pull/464) includes mapping the memory for the loader as NORMAL-cacheable. This broke SMP boot on all our boards (not just the STM32). Previously this was OK because all the loader memory was mapped as 'strongly ordered' (`DEVICE_nGnRnE`) and so per §B2.10.2 "Device memory" of ARM ARM DDI 0487 L.b,
>   > Data accesses to memory locations _[marked as device memory]_ are coherent for all observers in the system, and correspondingly are treated as being Outer Shareable.
>   
>   
>   This meant that around SMP core booting we did not need to perform cache maintenance. (Only I think `DMB`/`DSB` barriers which we do for the Raspberry Pi spin tables but not for writing the stack pointer for the PSCI boot).
> * Changing [aarch64: Restrict address mappings in loader.rs #464](https://github.com/seL4/microkit/pull/464) to be strongly ordered device memory again made SMP boot work on all our boards.
> 
> Therefore, this seems to be unnecessary from a 'does it work' standpoint on all our platforms (but we currently cannot test STM32).
> 
> However: in #463 you do note that the `STLR` instructions which the `atomic_store` and `atomic_load` compiler intrinsics compile to are at the very least questionable on DEVICE memory. We _should_ probably change this, but it does work on all our platforms. Is this the root cause for this case behind this change, or was it the change to make the loader memory cacheable that was the issue (especially given you needed an extra cache line flush operation here).

I have checked it, The atomic intrinsic 'does work' as it on the stm32 with the pages mapped DEVICE_nGnRnE.  So #463 is not necessary with this setting.
Since the values are not held in cache anymore, the cache status discrepancy between the main core and secondaries does not seem to be impacting, But that leaves a taste of strange unbalanced core settings here. That could be very similar to leave the pages in NORMAL mode and just do not enable the cache.
But OK, dissociate this from the stm32 support. That was just a side finding.

PR #478, opened 2026-04-22
tool: add --override-kernel option

Ivan-Velickovic:
Motivated by the comment here [1].

The main thing this will be useful for is adding debug prints/asserts to the kernel without having to re-compile the whole Microkit SDK.

Using kernels that are not compiled with the same configuration
or same version that the SDK was compiled will likely cause issues
and not work at all.

[1]: https://github.com/seL4/seL4/issues/1645#issuecomment-4267174595

### DIFF
--- a/tool/microkit/src/main.rs
+++ b/tool/microkit/src/main.rs
@@ -42,7 +42,7 @@ const KERNEL_COPY_FILENAME: &str = "sel4.elf";
 const KERNEL32_COPY_FILENAME: &str = "sel4_32.elf";
 
 fn print_usage() {
-    println!("usage: microkit [-h] [-o OUTPUT] [--image-type {{binary,elf,uimage}}] [-r REPORT] --board BOARD --config CONFIG [--capdl-json CAPDL_SPEC] --search-path [SEARCH_PATH ...] system")
+    println!("usage: microkit [-h] [OPTIONS] --board BOARD --config CONFIG [--search-path SEARCH_PATH ...] system")
 }
 
 fn print_help(available_boards: &[String]) {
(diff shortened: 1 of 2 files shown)


tool/microkit/src/main.rs
```diff
@@ -42,7 +42,7 @@ const KERNEL_COPY_FILENAME: &str = "sel4.elf";
 const KERNEL32_COPY_FILENAME: &str = "sel4_32.elf";
 
 fn print_usage() {
-    println!("usage: microkit [-h] [-o OUTPUT] [--image-type {{binary,elf,uimage}}] [-r REPORT] --board BOARD --config CONFIG [--capdl-json CAPDL_SPEC] --search-path [SEARCH_PATH ...] system")
+    println!("usage: microkit [-h] [-o OUTPUT] [--image-type {{binary,elf,uimage}} --override-kernel KERNEL] [-r REPORT] --board BOARD --config CONFIG [--capdl-json CAPDL_SPEC] --search-path [SEARCH_PATH ...] system")
```
midnightveil: the formatting of this is off, you need another pair of []

also it's usually okay for the usage to be slightly less verbose then --help

Indanz:
Thanks for adding this! This will save me a lot of time reproducing problems in the future.

Now my much less important request is to add a microkit option that spits out the config used for a board, so I don't need to go digging into build_sdk.py to figure that out.

midnightveil:
> Now my much less important request is to add a microkit option that spits out the config used for a board, so I don't need to go digging into build_sdk.py to figure that out.

What do you mean by this? Do you mean the values in `build_sdk` (the `cmake -D<blah>` command?) or the values of the config (.. which is there, under `gen_config.json`)?

Ivan-Velickovic:
> What do you mean by this? Do you mean the values in build_sdk (the cmake -D<blah> command?)

I think the options we explicitly pass when building the kernel so it's easier to reproduce it. The `gen_config.json` works but it has too many options to easily reproduce the same kernel just by looking at it.

I think we can store the exact CMake command or at least the explicit kernel options in the SDK and then get the tool to spit them out.

Something I've been meaning to add is put kernel build info etc in the report, so we should do that as well. That way if there's an issue with a particular system you can just look at the report to get more details.

PR #482, opened 2026-04-28
Loader MMU Cleanups (AArch64)

midnightveil:
This is a series of cleanups, except for one commit:

We were previously setting the MemAttr to the Stage 1 AttrIndex values from MAIR_EL1, but since we boot in hyp/stage 2 we were using the reserved 0b100 value (which is constrained unpredictable to behave as one of the other values).

### DIFF
--- a/tool/microkit/src/loader.rs
+++ b/tool/microkit/src/loader.rs
@@ -22,6 +22,12 @@ macro_rules! grab_symbol {
 const PAGE_TABLE_SIZE: usize = 4096;
 
 pub mod aarch64 {
+    //! For AArch64, our page tables can be both Stage 2 (if we run in EL2) or Stage
+    //! 1 (if we run in EL1, for non-hyp seL4). Generally, most attributes of the
+    //! page tables for the features we use are compatible in the layouts between
+    //! the Stage 2 descriptors and the Stage 1 descriptors. When modifying values
+    //! of the descriptors, please ensure that the values are valid for both stages
+    //! of the translation scheme.
     use crate::util::mask;
 
     pub const LVL0_BITS: u64 = 9;
@@ -43,15 +49,64 @@ pub mod aarch64 {
         idx as usize
     }
 
-    /// These match those in util64.S configured in the MAIR register
+    /// The one main difference between the Stage 1 and Stage 2 translation
+    /// tables is that bit[4:2] contains AttrIndex[2:0] (Stage 1) or
+    /// MemAttr[3:0] (Stage 2).
+    /// For Stage 1, the AttrIndex depends on our configured value of `MAIR_EL1`
+    /// done in util64.S.
+    /// For Stage 2, the values are fixed in Table D8-96 "Stage 2 MemAttr[3:0] encoding"
+    #[derive(Copy, Clone)]
+    pub enum MemAttr {
+        Stage1 { attr_index: u64 },
+        Stage2 { mem_attr: u64 },
+    }
 
+    impl MemAttr {
+        pub fn value(&self) -> u64 {
+            match *self {
+                MemAttr::Stage1 { attr_index } => attr_index,
+                MemAttr::Stage2 { mem_attr } => mem_attr,
+            }
+        }
+        pub fn is_normal_cacheable(&self) -> bool {
+            match *self {
+                MemAttr::Stage1 { attr_index } => attr_index == s1_mair_attr_index::MT_NORMAL,
+                MemAttr::Stage2 { mem_attr } => mem_attr == s2_mem_attr::NORMAL_INNER_WBC_OUTER_WBC,
+            }
+        }
+    }
+
+    /// These match those in util64.S configured in the MAIR_EL1 register,
+    /// which also needs to match the values that seL4 uses.
     #[allow(non_upper_case_globals, reason = "matching ARM naming convention")]
-    pub const MT_DEVICE_nGnRnE: u64 = 0;
+    pub mod s1_mair_attr_index {
+        pub const MT_DEVICE_nGnRnE: u64 = 0b000;
+        pub const MT_DEVICE_nGnRE: u64 = 0b001;
+        pub const MT_DEVICE_GRE: u64 = 0b010;
+        pub const MT_NORMAL_NC: u64 = 0b011;
+        pub const MT_NORMAL: u64 = 0b100;
+    }
+
+    /// See fixed values in Table D8-96 "Stage 2 MemAttr[3:0] encoding"
     #[allow(non_upper_case_globals, reason = "matching ARM naming convention")]
-    pub const MT_DEVICE_nGnRE: u64 = 1 << 2;
-    pub const MT_DEVICE_GRE: u64 = 2 << 2;
-    pub const MT_NORMAL_NC: u64 = 3 << 2;
-    pub const MT_NORMAL: u64 = 4 << 2;
+    pub mod s2_mem_attr {
+        pub const DEVICE_nGnRnE: u64 = 0b0000;
+        pub const DEVICE_nGnRE: u64 = 0b0001;
+        pub const DEVICE_nGRE: u64 = 0b0010;
+        pub const DEVICE_GRE: u64 = 0b0011;
+
+        pub const NORMAL_INNER_NC_OUTER_NC: u64 = 0b0101;
+        pub const NORMAL_INNER_WTC_OUTER_NC: u64 = 0b0110;
+        pub const NORMAL_INNER_WBC_OUTER_NC: u64 = 0b0111;
+
+        pub const NORMAL_INNER_NC_OUTER_WTC: u64 = 0b1001;
+        pub const NORMAL_INNER_WTC_OUTER_WTC: u64 = 0b1010;
+        pub const NORMAL_INNER_WBC_OUTER_WTC: u64 = 0b1011;
+
+        pub const NORMAL_INNER_NC_OUTER_WBC: u64 = 0b1101;
+        pub const NORMAL_INNER_WTC_OUTER_WBC: u64 = 0b1110;
+        pub const NORMAL_INNER_WBC_OUTER_WBC: u64 = 0b1111;
+    }
 
     pub mod descriptor_type {
         //! The translation table descriptor formats, as per §D8.3 "Translation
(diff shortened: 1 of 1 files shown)


tool/microkit/src/loader.rs
```diff
@@ -6,75 +6,73 @@
...
+    pub(crate) const PTE_PPN0_SHIFT: u64 = 10;
 
     /// Due to RISC-V having various virtual memory setups, we have this generic function to
     /// figure out the page-table index given the total number of page table levels for the
     /// platform and which level we are currently looking at.
     pub fn pt_index(pt_levels: usize, addr: u64, level: usize) -> usize {
-        let pt_index_bits = Self::PAGE_TABLE_INDEX_BITS * (pt_levels - level) as u64;
-        let idx = (addr >> (pt_index_bits + Self::PAGE_SHIFT)) % 512;
+        let pt_index_bits = PAGE_TABLE_INDEX_BITS * (pt_levels - level) as u64;
+        let idx = (addr >> (pt_index_bits + PAGE_SHIFT)) % 512;
```
Indanz, line 321: Shouldn't this be `% PAGE_TABLE_INDEX_BITS`?

tool/microkit/src/loader.rs
```diff
@@ -6,75 +8,328 @@
...
+    /// specifically subfigure "4KB, 16KB, and 64KB granules, 48-bit OA"
+    pub fn table_descriptor(addr: u64) -> [u8; 8] {
+        // Per Table D8-48, Condition for descriptor_type::TABLE is level != 3.
+
+        // We don't set any of these attributes, most are hardware-feature conditional
+        let attributes: u64 = 0;
+
+        // address must not have bits above 47 or below 12 set
+        assert!(addr & mask(12) == 0x0);
+        assert!(addr & mask(47) == addr);
```
Indanz:
```suggestion
        assert!(addr & mask(48) == addr);
```
?

tool/microkit/src/loader.rs
```diff
@@ -416,42 +414,42 @@ impl<'a> Loader<'a> {
 
         let mut boot_lvl1_pt: [u8; PAGE_TABLE_SIZE] = [0; PAGE_TABLE_SIZE];
         {
-            let text_index_lvl1 = Riscv64::pt_index(num_pt_levels, text_addr, 1);
-            let pt_entry = Riscv64::pte_next(boot_lvl2_pt_elf_addr);
+            let text_index_lvl1 = riscv64::pt_index(num_pt_levels, text_addr, 1);
```
Indanz, line 677: I'm surprised that the differences aren't be abstracted away and that there is so much Arch specific code left.

(This puzzlement also applies to the seL4 kernel code, but no reason to duplicate that approach elsewhere.)

tool/microkit/src/loader.rs
```diff
@@ -37,6 +35,191 @@ mod aarch64 {
...
+    pub const MT_DEVICE_GRE: u64 = 2 << 2;
+    pub const MT_NORMAL_NC: u64 = 3 << 2;
+    pub const MT_NORMAL: u64 = 4 << 2;
+
+    pub mod descriptor_type {
+        //! The translation table descriptor formats, as per §D8.3 "Translation
+        //! table descriptor formats" of ARM DDI 0487 L.b. Specifically,
+        //! as per "Table D8-48 Determination of descriptor type"
+
+        /// Descriptor type: Table. Condition is lookup level != 3.
```
Indanz, line 116: Instead of hard-coding your 3-level assumption, perhaps word it as "For intermediate levels" and "For last level".

tool/microkit/src/loader.rs
```diff
@@ -647,6 +647,82 @@ impl<'a> Loader<'a> {
         ]
     }
 
+    /// AArch64 loader page tables when the loader exists in EL2 (hypervisor config)
+    /// starts at the TTBR0_EL2 table. In non-hypervisor mode, we have
+    /// two Level 0 tables, Level 0 Lower (TTBR0_EL1) and Lower 0 Upper (TTBR1_EL1);
```
Indanz: "Lower 0 Upper" -> "Level 0 Upper"

But it's not a very useful comment in the sense it doesn't explain which table is selected. That's done via bit 56 of the address.

tool/microkit/src/loader.rs
```diff
@@ -647,6 +647,82 @@ impl<'a> Loader<'a> {
...
+    /// and these contain the Level 1 Lower and Level 1 Upper tables respectively.
```
Indanz: Assuming you meant "and these contain the EL0 Lower and EL1 Upper tables respectively":

If you use the word "level" for both page table levels and exception levels, things get confusing fast. Better to avoid the word perhaps and say ELn and "page table" instead.

tool/microkit/src/loader.rs
```diff
@@ -43,15 +51,63 @@ pub mod aarch64 {
...
+        }
+        pub fn is_normal_cacheable(&self) -> bool {
+            match *self {
+                MemAttr::Stage1 { attr_index } => attr_index == s1_mair_attr_index::MT_NORMAL,
+                MemAttr::Stage2 { mem_attr } => mem_attr == s2_mem_attr::NORMAL_INNER_WBC_OUTER_WBC,
+            }
+        }
+    }
 
+    /// These match those in util64.S configured in the MAIR_EL1 register.
```
Indanz: Maybe a good place to also mention that this should match what the kernel expects.

Indanz (commented):
I think comment _"loader: cleanup, pte functions return [u8]"_ makes things worse, not better. The only reason it needs to be an array is because `copy_from_slice` is used. This comments spreads that local detail to other code with no gain.

Better solution would be to make `boot_lvl*` u64 arrays or something instead of this weird byte array stuff.

tool/microkit/src/loader.rs
midnightveil: Yes, it is true that a lot of it is duplicated. This is by no means done: I didn't even start cleaning up RISC-V much.

But Riscv64 and Aarch64 has different kernel page tables so some has to be different, and the internal layouts need to be different.

tool/microkit/src/loader.rs
midnightveil: This is a direct quote from the ARM manual.

tool/microkit/src/loader.rs
midnightveil: No, I am referring to Level 0 pagw tables, i.e. the "root" (for a 4-stage walk at least).

There's no EL0 lower table, it's only by convention that TTBR0_EL1 is used for userspace and TTBR1_EL1 for kernel.

tool/microkit/src/loader.rs
midnightveil: Do you mean (1 << PAGE_TABLE_INDEX_BITS)?

But yes I haven't touched RISC-V much.

tool/microkit/src/loader.rs
```diff
@@ -159,7 +159,7 @@ pub mod aarch64 {
     /// specifically subfigure "4KB, 16KB, and 64KB granules, 48-bit OA"
     /// Also "Table D8-53 Stage 2 VMSAv8-64 Block and Page descriptor fields"
     /// for the Stage 2 attributes.
-    pub fn block_descriptor(level: usize, addr: u64, mem_attr: MemAttr) -> u64 {
+    pub fn block_descriptor(level: usize, addr: u64, mem_attr: MemAttr) -> [u8; 8] {
```
midnightveil: @Indanz said

> I think comment "loader: cleanup, pte functions return [u8]" makes things worse, not better. The only reason it needs to be an array is because copy_from_slice is used. This comments spreads that local detail to other code with no gain.

This might be true. I'm not happy with the code we use to actually set up the page tables. It's annoying doing those bitshifts manually and messing around with the indices doing `8 * i` etc. For some silly reason `copy_from_slice` expects the dest and source to have precisely the same length which is just annoying.

Unfortunately we can't just transmute `Vec<u64>` to `Vec<u8>`, as it is UB (some allocators have type-alignment aware free() calls). Though we could make everything a `Box<[u64: 512]>` (i.e. a heap allocated fixed size array, instead of a growable vector) and then it would be OK I think.

tool/microkit/src/loader.rs
Indanz: Sure, top level is different and low level is different. But that doesn't mean that all the intermediate code can't be shared...

tool/microkit/src/loader.rs
Indanz: Oh well....

tool/microkit/src/loader.rs
Indanz: So you're saying the same thing twice?

"In non-hypervisor mode, we have two Level 0 tables, Level 0 Lower (TTBR0_EL1) and Level 0 Upper (TTBR1_EL1); and these contain the Level 0 Lower and Level 0 Upper tables respectively."

That's confusing and redundant.

tool/microkit/src/loader.rs
Indanz: Yes, that's what I meant.

tool/microkit/src/loader.rs
midnightveil: I agree. I just haven't done it yet.

tool/microkit/src/loader.rs
midnightveil: No, I'm saying (confusingly and probably can be reworded):

For Stage 2, there's a singular TTBR0_EL2 (Level 0). It contains PTE pointing to the tabled drawn in the diahram as "Level 1 Upper" and "Level 1 Lower".

For Stage 1, there's a both TTBR0_EL1 and TTBR1_EL1 (both Level 0). TTBR0_EL1 contains the PTE pointed to "Level 1 Lower" as shown in the diagram, whereas TTBR_EL1 contains the table descriptor that points at "Level 1 Upper". Other than this difference in the setup of the Level 0 tables and the table descriptors pointing to Level 1, there is no difference between Stage 1 and 2.

tool/microkit/src/loader.rs
midnightveil: [0001-loader-cleanup-pte-functions-return-u8.patch](https://github.com/user-attachments/files/27191367/0001-loader-cleanup-pte-functions-return-u8.patch)

Removed that commit.

PR #483, opened 2026-04-29
tool: fix kernel mappings for RISC-V loader

### LINKED ISSUE #481: riscv64: memory paging doesn't seem to work if physBase()>0
Hello,
I'm currently in trying to get microkit running on the [BananaPI-F3](https://docs.sel4.systems/Hardware/bananapi-f3.html).
Please keep in mind that I'm just starting in the seL4 ecosystem and that the issue could be rooted in a misunderstanding on my part.

The first 512 KiB of the physical address range is reserved and cannot be used for the kernel.
Therefore, `physBase()` starts at `0x80000`.
When setting up the virtual address tables, the microkit tool skips the third level and uses 2MiB "megapages" instead.
It adds `physBase()` as an offset to the page address of the 2nd level page table (512KiB in the case of the BananaPi-F3).

[Chapter 12.4.1 of the RISC-V Volume II: Privileged ISA Extension](https://docs.riscv.org/reference/isa/v20250508/priv/supervisor.html#addressing-and-memory-protection) states:
> Any level of PTE may be a leaf PTE, so in addition to 4 KiB pages, Sv39 supports 2 MiB megapages and 1 GiB gigapages, each of which must be virtually and physically aligned to a boundary equal to its size. A page-fault exception is raised if the physical address is insufficiently aligned.

Unsurprisingly the processor traps on a "Load Page fault" as soon as the kernel address (`0xffffffff80080000`) is loaded (due to the invalid alignment) and the system can't boot.
One fix I've tried is to round the addresses in the 2nd level page table down to 2MiB.
That allows seL4 to start partially, but seL4 fails to start fully due to a failed assertion in `sel4/src/arch/riscv/kernel/vspace.c:205` (`lu_ret.ptBitsLeft == seL4_PageBits`).
I've also tried to increase the reserved area up to 2MiB with the same assertion error.

If you don't have other suggestions, I'll  try to add a 3rd level to the generated page table.

I'm using the `15.0.0` brach (`881de507fe528490dc5e570c7810a149bad5880f`) of seL4 with the current `main` branch (`dd818c11cabc86420c3a882c6730c77e47c6d030`) of microkit.

Greetings
Paul

Ivan-Velickovic:
Done on top of https://github.com/seL4/microkit/pull/482.

See commit message for details.

Closes https://github.com/seL4/microkit/issues/481.

### DIFF
--- a/loader/src/riscv/mmu.c
+++ b/loader/src/riscv/mmu.c
@@ -14,7 +14,7 @@
 uint64_t boot_lvl1_pt[1 << 9] ALIGN(1 << 12);
 uint64_t boot_lvl2_pt[1 << 9] ALIGN(1 << 12);
 /* Paging structures for identity mapping */
-uint64_t boot_lvl2_pt_elf[1 << 9] ALIGN(1 << 12);
+uint64_t boot_lvl2_pt_loader[1 << 9] ALIGN(1 << 12);
 
 
 /*
--- a/loader/src/riscv/mmu.c
+++ b/loader/src/riscv/mmu.c
@@ -13,6 +13,7 @@
 /* Paging structures for kernel mapping */
 uint64_t boot_lvl1_pt[1 << 9] ALIGN(1 << 12);
 uint64_t boot_lvl2_pt[1 << 9] ALIGN(1 << 12);
+uint64_t boot_lvl3_pt[1 << 9] ALIGN(1 << 12);
 /* Paging structures for identity mapping */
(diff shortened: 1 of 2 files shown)


midnightveil:
View diff here: https://github.com/au-ts/microkit/compare/julia/loader-mmu...seL4:microkit:riscv_mappings_fix

midnightveil (approved):
Once #482 is merged.

PR #484, opened 2026-04-30
loader: set exception handler for RISC-V

### LINKED ISSUE #480: riscv64: feature-request: page fault handler
Hello,
I'm currently in trying to get microkit running on the [BananaPI-F3](https://docs.sel4.systems/Hardware/bananapi-f3.html).
Please keep in mind that I'm just starting in the seL4 ecosystem and that the issue could be rooted in a misunderstanding on my part.

During my testing I've noticed that contrary to the comment in [`loader/src/riscv/exceptions.c:arch_set_exception_handler`](https://github.com/seL4/microkit/blob/dd818c11cabc86420c3a882c6730c77e47c6d030/loader/src/riscv/exceptions.c#L12) the M-Mode doesn't seem to catch all exceptions (especially page faults).
I think that at least the debug config should have a page fault handler so that one can at least see scause, stval and sepc if the boot fails. 

I'm running the vendor supplies OpenSBI and U-Boot.

Greetings
Paul

Ivan-Velickovic:
Closes https://github.com/seL4/microkit/issues/480.

### DIFF
--- a/loader/src/riscv/exceptions.c
+++ b/loader/src/riscv/exceptions.c
@@ -6,9 +6,113 @@
  */
 
 #include "../arch.h"
+#include "../cutil.h"
+#include "../uart.h"
+
+#if __riscv_xlen == 32
+#define SCAUSE_IRQ BIT(31)
+#define SCAUSE_CODE MASK(31)
+#else
+#define SCAUSE_IRQ BIT(63)
+#define SCAUSE_CODE MASK(63)
+#endif
+
+static inline uint64_t read_stval(void)
+{
+    uint64_t temp;
+    asm volatile("csrr %0, stval" : "=r"(temp));
+    return temp;
+}
+
+static inline uint64_t read_scause(void)
+{
+    uint64_t temp;
+    asm volatile("csrr %0, scause" : "=r"(temp));
+    return temp;
+}
+
+static inline uint64_t read_sepc(void)
+{
+    uint64_t temp;
+    asm volatile("csrr %0, sepc" : "=r"(temp));
+    return temp;
+}
+
+static inline void write_stvec(void *value)
+{
+    asm volatile("csrw stvec, %0" :: "rK"(value) : "memory");
+}
+
+static char *scause_to_string(uint64_t scause)
+{
+    if (scause & SCAUSE_IRQ) {
+        switch (scause & SCAUSE_CODE) {
+        case 1:
+            return "Supervisor software interrupt";
+        case 5:
+            return "Supervisor timer interrupt";
+        case 9:
+            return "Supervisor external interrupt";
+        case 13:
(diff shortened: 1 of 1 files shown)


loader/src/riscv/exceptions.c
```diff
@@ -6,9 +6,104 @@
  */
 
 #include "../arch.h"
+#include "../cutil.h"
+#include "../uart.h"
+
+#if __riscv_xlen == 32
```
midnightveil, line 12: Why do we bother doing this when everything else assumes 64-bit?

Ivan-Velickovic: I just did it since it's trivial

loader/src/riscv/exceptions.c
```diff
@@ -6,9 +6,104 @@
...
+static inline uint64_t read_scause(void)
+{
+    uint64_t temp;
+    asm volatile("csrr %0, scause" : "=r"(temp));
+    return temp;
+}
+
+static inline void write_stvec(void *value)
+{
+    asm volatile("csrw stvec, %0" :: "rK"(value));
```
midnightveil: "rK" here means `A 5-bit unsigned immediate for CSR access instructions.`. Was there a reason why rK was used?

Also, don't these need a `: "memory"` barrier otherwise the compile is free to reorder them?

midnightveil (approved):

loader/src/riscv/exceptions.c
midnightveil: Answer: "rK" means register *or* 5-bit immediate; binutils in `opcodes/riscv-opc.c` defines:

```c
{"csrw",       0, INSN_CLASS_ZICSR, "E,s",   MATCH_CSRRW, MASK_CSRRW|MASK_RD, match_opcode, INSN_ALIAS },
{"csrw",       0, INSN_CLASS_ZICSR, "E,Z",   MATCH_CSRRWI, MASK_CSRRWI|MASK_RD, match_opcode, INSN_ALIAS },
```

so it is clever enough to encode a `csrw` in *either* the `csrrw` (register) variant or the `csrrwi` (immediate) variants based on which kind the `%0` fits in.

```
	    case 'Z': /* CSRRxI immediate.  */
	      my_getExpression (imm_expr, asarg, force_reloc);
	      check_absolute_expr (ip, imm_expr, false);
	      if ((unsigned long) imm_expr->X_add_number > 31)
		as_bad (_("improper CSRxI immediate (%"PRIu64")"),
			imm_expr->X_add_number);
	      INSERT_OPERAND (RS1, *ip, imm_expr->X_add_number);
	      imm_expr->X_op = O_absent;
	      asarg = expr_parse_end;
	      continue;
```

So this is GCC magically handling immediates as an optimisation.

PR #486, opened 2026-05-05
tool: Separate argument parsing

zaklogician:
As part of a series of patches which upstream Viper export to Microkit, we separate argument parsing from `main.rs` into `argparse.rs`.

The code was refactored a bit to remove manual index manipulation. The user interface remains compatible/identical to the previous implementation.

### DIFF
--- a/tool/microkit/src/main.rs
+++ b/tool/microkit/src/main.rs
@@ -7,6 +7,8 @@
 // we want our asserts, even if the compiler figures out they hold true already during compile-time
 #![allow(clippy::assertions_on_constants)]
 
+use microkit_tool::argparse;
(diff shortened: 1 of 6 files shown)


tool/microkit/src/main.rs
```diff
@@ -7,6 +7,8 @@
 // we want our asserts, even if the compiler figures out they hold true already during compile-time
 #![allow(clippy::assertions_on_constants)]
 
+use microkit_tool::argparse;
+use microkit_tool::argparse::{Args, ArgsError, RequestedImageType};
```
Indanz, line 11: One of these is redundant?

PR #488, opened 2026-05-12
tool: migrate to new environment info gathering

zaklogician:
As part of a series of patches which upstream Viper export to Microkit, we separate SDK environment discovery from `main.rs` into `sdk.rs`.

SDK discovery is the following process:
1. finding the location of the Microkit directory based on the `MICROKIT_SDK` env variable, or if that's not set then based on the location of the tool executable itself)
2. finding the list of available boards and configs (`debug`/`release`/etc.)

This has to happen before argument parsing, and lets us give better error messages (e.g. we can report available configs if an invalid `--config` argument is given).

### DIFF
--- a/tool/microkit/src/argparse.rs
+++ b/tool/microkit/src/argparse.rs
@@ -4,6 +4,7 @@
 // SPDX-License-Identifier: BSD-2-Clause
 //
 
+use crate::sdk::Sdk;
 use std::fmt;
 use std::iter::Peekable;
 use std::path::PathBuf;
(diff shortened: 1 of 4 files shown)


zaklogician:
This is ready for review, the failing check is a wontfix gitlint issue resulting from my long email address.

midnightveil:
> This is ready for review, the failing check is a wontfix gitlint issue resulting from my long email address.

FYI: We fixed this so it will no longer happen. Now it complains for other reasons.

PR #489, opened 2026-05-12
loader: fix mismatched type for inline assembly

KurtWu10:
This fixes the following warning when using the llvm toolchain
```
src/aarch64/init.c:140:45: error: value size does not match register size specified by the constraint and modifier [-Werror,-Wasm-operand-widths]
  140 |     asm volatile("msr TPIDR_EL1, %0" :: "r"(logical_cpu));
      |                                             ^
src/aarch64/init.c:140:34: note: use constraint modifier "w"
  140 |     asm volatile("msr TPIDR_EL1, %0" :: "r"(logical_cpu));
      |                                  ^~
      |                                  %w0
1 error generated.
```


Ivan-Velickovic:
Thanks, I noticed that recently but forgot to fix.

PR #498, opened 2026-05-22
ci: add a script for running the CI locally

midnightveil:
This is useful for local testing as sometimes the GitHub CI can be annoying to test.

Needs https://github.com/seL4/ci-actions/pull/462 to be useful, as well as a few other changes internally to the scripts to handle cancellation more gracefully.

### DIFF
--- a/ci/local_run.sh
+++ b/ci/local_run.sh
@@ -1,0 +1,99 @@
+#!/usr/bin/env bash
+
+# Copyright 2026, UNSW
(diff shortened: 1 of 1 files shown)


midnightveil:
Unblocked by output PR being merged now.

However, at the moment cancellation with SIGINT doesn't really work as it keeps trying to run the next test, so I think we need to plumb handling off error code 130 through (130 conventionally means "interrupted by sigint").

Or perhaps make Python handle the SIGINT and kill the children.

PR #511, opened 2026-06-04
loader: Map all of RAM in the loader

### LINKED ISSUE #463: broken atomic lock when starting multiple cores

### LINKED ISSUE #466: fix aarch64 MMU fixes - don't disable it
In #353, I [disabled the MMU and caches at AArch64 startup](https://github.com/seL4/microkit/commit/fb120de58153fc0868a3c60224e31f2f178a4349) because the loader regions could be loaded into a location that overlapped with U-Boot's initial page tables. In hindsight, the better solution here would have been to enable MMU mappings using the loader's page tables. This would improve boot performance for large images, as copy without cache is somewhat slow. #465 also claims that the MMU is necessary for cross-core atomic shared data.

With #464 happening for STM32, where we only map the loader data, we should likely also get upstream seL4 to provide us the uart [it wants for early printf](https://github.com/seL4/microkit/pull/464#discussion_r3026119892), and we'll need to map all the loader regions in too.

midnightveil:
This will fix #463 (closing #465) and #466.


midnightveil:
Ok, this breaks SMP, because I'm an absolute buffoon, and changed the loader mappings to be NORMAL, when of course that is wrong when we don't care about caching.

https://github.com/seL4/microkit/pull/465#issuecomment-4310869367

I can now fix this booting by keeping everything as `DEVICE_nGnRnE` as it was before. (But I will change this to care about caching properly, just not *immediately*).

midnightveil:
I'm debating whether it makes sense to instead allocate a fixed amount of slots in the loader, and do this code in the loader. Then in (some future world) where we have one build of the loader for every platform on AArch64 (once seL4 has that) then this will just work like that, too.

The sel4_tools elfloader also works like that.

I think I see many positives for doing it at runtime, and the only one I can maybe see for build-time is just that we can know it works at build time.

Indanz:
Yes, of course do it runtime. If you go through the trouble making your own loader, at least have one that supports all boards of the same platform. :-)

midnightveil:
Indan's previously made the comment that we should just turn on the cache-enabled bit in SCTLR. If that speeds up boot appropriately I think that's probably a much simpler solution, as one only has to map a few things instead of all of RAM.

Although.... that doesn't fix #463 properly, so maybe I won't.

Indanz:
> Indan's previously made the comment that we should just turn on the cache-enabled bit in SCTLR.

Could, not should. Just an option.

But I'm not convinced that mapping all memory in the loader is a good idea, instead of mapping the minimum necessary.

midnightveil:
This doesn't appear to be that much faster :pensive:

midnightveil:
OK, I'm just silly. One of course needs to map ram as MT_NORMAL not MT_DEVICE for it to be faster. Once I do that a 100MiB image is practically instant to copy. 

Woo.

PR #515, opened 2026-06-04
manual: don't list out all MSRs and VMCS fields

dreamliner787-9:
~~The kernel now permits userspace to read `IA32_VMX_MISC_MSR` MSR to operate the VMX preemption timer in commit 7441046. This commit updates the manual to reflect that. Since the seL4 version in Microkit had been recently bumped which included that commit.~~

The previous approach is too unwieldy and required constant updates.
We should just link the user to the kernel instead and have one
source of truth.

### DIFF
--- a/docs/manual.md
+++ b/docs/manual.md
@@ -778,93 +778,9 @@ Read an 8, 16, or 32 bits value at port address `port_addr` from I/O Port with I
 Read a Virtual Machine Control Structure field specified by `field` for a given virtual CPU
 with ID `vcpu` using the `vmread` instruction.
 
-All the VMCS fields are listed in the seL4 source at `include/arch/x86/arch/object/vcpu.h`.
-
-The following fields can be read from:
-
-* `VMX_GUEST_RIP`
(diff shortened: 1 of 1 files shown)


midnightveil (approved):
IDK why we need a long list of everything supported tbh.

midnightveil:
Actually, yeah. I think I changed my mind. I'm guessing there's an x86 reference for these registers? I don't think we should list all of them here, it's unwieldy.

dreamliner787-9:
You could consult the Intel SDM, but the kernel doesn't allow you to read/write to all of the fields. So I thought it would be best if we list out which fields the user can touch.

midnightveil:
I think we should do something along the lines of "seL4 only allows certain MSRs to be written by userspace; for a full list of supported ones, see <sel4 documentation location that is kept in sync with sel4 implementation>".

midnightveil (approved):
It might be worth making it so that seL4 documents this somewhere outside of the source, so that we don't have to link there, but this is good enough for now.

Issue #520, opened 2026-06-11
Cap Sharing Improvements


midnightveil:
Now that we have cap sharing with #406, some improvements that can be made:

- [ ] Experiment with removing the PD hierarchy feature entirely. 
   -  Specifically, we need a replacement for setting up the fault endpoints, but this
      does have a 1:1 mapping. (the cap_sharing is not good enough for this)
   - I would want the ability to automatically migrate users from hierarchy to cap sharing, i.e. run a program that converts system descriptor files.
- [ ] Make the "user caps" (`MICROKIT_MAX_USER_CAPS=128`) its own child cnode that is dynamically sized, there's no reason to limit this (see also: the untypeds PR).
     - This might end up with the `microkit_cspace_slot_to_cptr` function needing to be reworked...
     - **Important**: decide on this before the release, to avoid unnecessary future breakage.
- [ ] Auto-allocation of slots? (edit: maybe not?)
- [ ] Permissions on caps
- [ ] Other cap types

_Originally posted by @midnightveil in https://github.com/seL4/microkit/issues/406#issuecomment-4629492200_

midnightveil:
Zoltan also happened to have agreed with removing the parent-child hierarchy as currently implemented in favour of some other way to ask for fault handler endpoints.

PR #521, opened 2026-06-11
GDB / Child VSpace

midnightveil:
From Krishnan.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -249,6 +249,7 @@ dependencies = [
  "sel4-capdl-initializer-types",
  "serde",
  "serde_json",
+ "zerocopy",
 ]
 
 [[package]]
@@ -962,4 +963,24 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "cbc04313cab124e498ab1724e739720807b6dc405b9ed0edc5860164d2e4ff70"
 dependencies = [
  "xml",
+]
+
+[[package]]
+name = "zerocopy"
+version = "0.8.33"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "668f5168d10b9ee831de31933dc111a459c97ec93225beb307aed970d1372dfd"
+dependencies = [
+ "zerocopy-derive",
+]
+
+[[package]]
+name = "zerocopy-derive"
+version = "0.8.33"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "2c7962b26b0a8685668b671ee4b54d007a67d4eaf05fda79ac0ecf41e32270f1"
+dependencies = [
+ "proc-macro2",
+ "quote",
+ "syn",
 ]
--- a/build_sdk.py
+++ b/build_sdk.py
@@ -443,7 +443,8 @@ SUPPORTED_CONFIGS = (
         kernel_options={
             "KernelDebugBuild": True,
             "KernelPrinting": True,
-            "KernelVerificationBuild": False
+            "KernelVerificationBuild": False,
+            "HardwareDebugAPI": True,
         },
         kernel_options_arch={},
     ),
--- a/libmicrokit/include/microkit.h
+++ b/libmicrokit/include/microkit.h
@@ -16,6 +16,7 @@ typedef unsigned int microkit_child;
 typedef unsigned int microkit_ioport;
 typedef seL4_MessageInfo_t microkit_msginfo;
 
+#define VSPACE_CAP 3
 #define MONITOR_EP 5
 /* Only valid in the 'benchmark' configuration */
 #define TCB_CAP 6
@@ -25,9 +26,11 @@ typedef seL4_MessageInfo_t microkit_msginfo;
 #define BASE_ENDPOINT_CAP 74
 #define BASE_IRQ_CAP 138
 #define BASE_TCB_CAP 202
-#define BASE_VM_TCB_CAP 266
-#define BASE_VCPU_CAP 330
-#define BASE_IOPORT_CAP 394
+#define BASE_VSPACE_CAP 266
+#define BASE_VM_TCB_CAP 330
+#define BASE_VCPU_CAP 394
+#define BASE_IOPORT_CAP 458
+#define BASE_FRAME_CAP 522
 
 #define MICROKIT_MAX_CHANNELS 62
 #define MICROKIT_MAX_CHANNEL_ID (MICROKIT_MAX_CHANNELS - 1)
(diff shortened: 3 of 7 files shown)


cazb2:
These thoughts relate to the Vspace/Frame handling rather than the GDB logic. It would be great if we could implement a more general solution to accessing frame capabilities, both one's own and another PD's.

Another approach that I have experimented with is the idea that you can share frames on a per memory region basis i.e. if PD A wants the frames that PD B has for memory region X it could unambiguously specify this in the SDF. Some potential caveats being the stack, maybe also the IPC buffer frame and intermediate page table caps....
 
This would then give PD A essentially an array of frame caps ordered on the corresponding virtual addresses that PD B has them mapped at. Then by providing the base frame capability (slot in PD A's CNode), base virtual address, maybe the base physical address if needed, memory region size and frame type (small, large, huge) etc PD A can do a constant time lookup to any frame based on the virtual address PD B has mapped it at. This would also assume a memory region only uses one type of frame.

Then based on the user application e.g. GDB a PD could allocate itself an appropriately sized memory region to build any application specific data structures such as page tables or hash map or trees etc.

midnightveil:
Yes, this is one of the improvements I want to make with #520. It should be possible to ask for capabilities at a more fine grained level than current.

> This would then give PD A essentially an array of frame caps ordered on the corresponding virtual addresses that PD B has them mapped at. 

This also relates to #422. `setvar` has limitations that have become apparent, especially on x86 where many physical addresses are determined at runtime.

We should handle this generically and in a uniform manner, rather than various adhoc mechanisms for specific usecases.

midnightveil:
We need to make sure this change only gives extra caps in the case where we actually want this feature to be active, and not unconditionally.

midnightveil:
Idle thought: I'm wondering if this actually ever needs to give capabilities. Mentioned by @joep [here](https://github.com/au-ts/microkit/pull/10#issuecomment-5293458203), one can simulate the current necessary behaviour by just converting a program image into a bunch of memory regions and mapping them in.

So in essence what we could actually do is have the ability to do `<map mr` using some autogenerated regions based on the section names of the `<program_image>`. And this can map in at mostly arbitrary vaddr in the "parent" and then some metadata tracking mechanism. Do we actually *need* the caps - and if we did, surely this should be a generic mechanism for memory regions.

This wouldn't be changing any capability features; however, we would lose the guarantee that nobody can touch program code, which is an assumption in many of our proofs.

joep:
> Idle thought: I'm wondering if this actually ever needs to give capabilities. Mentioned by @joep [here](https://github.com/au-ts/microkit/pull/10#issuecomment-5293458203), one can simulate the current necessary behaviour by just converting a program image into a bunch of memory regions and mapping them in.
> 
> So in essence what we could actually do is have the ability to do `<map mr` using some autogenerated regions based on the section names of the `<program_image>`. And this can map in at mostly arbitrary vaddr in the "parent" and then some metadata tracking mechanism. Do we actually _need_ the caps - and if we did, surely this should be a generic mechanism for memory regions.
> 
> This wouldn't be changing any capability features; however, we would lose the guarantee that nobody can touch program code, which is an assumption in many of our proofs.

I would argue against this approach (yes it is what we did but we didn't really want to).  It is simpler than cap sharing but it is also more costly. Every debuggee PD needs its own disjoint VA range in the debugger's address space and the page-table objects created and populated by the CapDL initializer at boot. Always mapped, whether or not anyone ever inspects it because access is entirely driven by unpredictable user requests. Debug support becomes a cost that grows linearly with the whole system's memory.

I think cap frame sharing is a much better and lighter weight alternative.  Debugger can hold cheap cap table slots and map/unmap on demand into one reusable scratch VA. Boot-time work and static kernel-object count stays flat regardless of how many PDs or how much memory the system has and page table entries are created only when the debugger chooses to look at something and can then be torn back down after. It only has access to other PDs during the brief peek/poke mapping window.

midnightveil:
> I would argue against this approach (yes it is what we did but we didn't really want to). It is simpler than cap sharing but it is also more costly. Every debuggee PD needs its own disjoint VA range in the debugger's address space and the page-table objects created and populated by the CapDL initializer at boot. Always mapped, whether or not anyone ever inspects it because access is entirely driven by unpredictable user requests. Debug support becomes a cost that grows linearly with the whole system's memory.

True, though Microkit is for static systems, so it's possible this is not actually an issue.

---

I'll also note (I thought I had mentioned it here), but whilst providing the capabilities would cause issues, so would providing non-RO access to memory, as you could either overwrite the instructions a PD is executing from, or if you have TCB access make it jump to some shared memory and execute instructions. Manipulating out the instructions involved would violate any verification we do (regardless of whether or not you have access to TCB caps)

Obviously, GDB things are a debug-only feature, but not all of the uses of this are "debug" only. Possibly not actually an issue - surely if you are using this feature you know and intend this - but it does have implications, and one would have to make sure to disallow these systems from any verification of the children. (Or at least assuming that the parent/controller is trusted and doesn't do that).

joep:
> > I would argue against this approach (yes it is what we did but we didn't really want to). It is simpler than cap sharing but it is also more costly. Every debuggee PD needs its own disjoint VA range in the debugger's address space and the page-table objects created and populated by the CapDL initializer at boot. Always mapped, whether or not anyone ever inspects it because access is entirely driven by unpredictable user requests. Debug support becomes a cost that grows linearly with the whole system's memory.
> 
> True, though Microkit is for static systems, so it's possible this is not actually an issue.
> 
> I'll also note (I thought I had mentioned it here), but whilst providing the capabilities would cause issues, so would providing non-RO access to memory, as you could either overwrite the instructions a PD is executing from, or if you have TCB access make it jump to some shared memory and execute instructions. Manipulating out the instructions involved would violate any verification we do (regardless of whether or not you have access to TCB caps)
> 
> Obviously, GDB things are a debug-only feature, but not all of the uses of this are "debug" only. Possibly not actually an issue - surely if you are using this feature you know and intend this - but it does have implications, and one would have to make sure to disallow these systems from any verification of the children. (Or at least assuming that the parent/controller is trusted and doesn't do that).

I agree with all this. In particular, given the size of the systems we are going to be producing, carrying the added heft to the system for the debug builds and any implications for verification aren't an issue at all for us specifically because it will never be included in a production system. It seems to me, however, that if cap sharing becomes a generally available feature, folding this into that is the more desirable approach. If ultimately cap sharing is rejected or significantly delayed (IIRC, there was something mentioned about that but I may be wrong) then there are really many different ways to support GDB in a way that would work just as well, if not better.

The fragility of what we do is that we must follow the drift of the microkit changes and it has broken a couple of times since we implemented it. A more stable approach is simply to formalize what we already do. If a hook path is configured (CLI flag or a field in the SDF itself, e.g. `<protection_domain ... debug_hook="path">`), Microkit shells out to it, passing a small versioned purpose-built JSON blob on stdin:
```json
{
  "schema_version": 1,
  "protection_domains": [
    {"name": "pd...", "regions": [
      {"name": "elf_segment_0", "phys_addr": "0x...", "size": 4096, "perms": "rw"},
      ...
    ]}
  ]
}
```
and the hook could return `<memory_region>/<map>` fragments scoped to one designated PD in a machine-readable reply.

Of course, this ONLY considers what is needed to support a debugger specifically and ignores any other plans for microkit since we don't know what those are. It could be done fairly easily (as compared to cap sharing frames; I'm not trying to imply it is 'easy' to do). This is also only one approach to doing it, there are others of course.

PR #467, opened 2026-04-10
Add iommu support to microkit

Cheng-Li1:
This pull request adds iommu support for Microkit and needs [Rust-Sel4](https://github.com/au-ts/rust-sel4/tree/cheng/iommu_support) with the iommu support to work.

Supported features:
Map the memory region to a device-visible memory address.

Tests done:
Blk example with iommu feature enabled on x86 qemu and hardware(vb-105).

PR #528, opened 2026-06-18
IOMMU: Add SDF Support

### LINKED ISSUE #418: Support IOMMU on x86-64
Tracking issue, currently being worked on at TS. Should be done in the near future, but until then we will disable the IOMMU on x86-64.

This issue will be resolved before the next release.

cazb2:
This commit adds iommu support to microkit sdfs.

Define a constant to represent the a tool enforced limitation on the maximum virtual address suported for an IO address space. This is due to current x86 IOMMU behaviour in the kernel which dynamically detects the maximum virtual address for an IO address space. For more info read: BOOT_CODE bool_t vtd_init_num_iopts(uint32_t num_drhu) in seL4/src/plat/pc99/machine/intel-vtd.c.

Add PciDevice struct and PCIDeviceParseError enum. This types allow us to implement the Display trait which gives a consistent way to output PCI information. The FromStr trait implemented for PCIDevice avoid duplicating logic required to parse the PCI identifier from the xml. When combined with the display trait on the PCIDeviceParseError, we get consistent error messages and avoid repeating code.

Add IOMMUDeviceIdentifier and IOMMUDeviceIdentifierParseError to support parsing the sdf based on the current architecture. This allows the sdf parsing to easily support ARM SMMU and RISCV IOMMU. The Display traits again remove repeated code when outputting IOMMUDeviceIdentifer or IOMMUDeviceIdentifierParseError, particularly useful for error messages.

Add SysIOMapPerms. Representing the perms in an enum completely removes the existence of invalid IOMapPerms once we parse the sdf.

Add SysIOMap. The dual to SysMap, used to represent a mapping in a devices IO virtual address space.

Replace ExecutionContext trait with a Map trait. The previous trait was primarily used in the check_maps function to handle memory regions in virtual machines and normal pds. This Map trait allowed the one check_maps function to handle both SysIOMaps and SysMaps uniformly. The additional information that was provided by the polymorhpic ExecutionContext is just passed as a string to be used in the respective error message. This is because outside of the check_maps function a protection_domain and a virtual_machine are handled as distinct types.

Add error checking cases to check_maps. If the map_start + mr.size overflows alot of the pre-exisiting logic breaks so check for it. Also check that we don't overflow the top of user_virtual memory accouting for the differences between protection_domains, virtual_machines and io address spaces.

Add check_io_maps adapter. Required because check_maps is designed to error check within one address space. Therefore we collect all the mappings that have been made on a per IO address space basis.

capdl/irq.rs waas using the old pci_bus,pci_dev,pci_func form. Since we have a real type now, we updated the expected match form.

### DIFF
--- a/docs/manual.md
+++ b/docs/manual.md
@@ -1093,14 +1093,18 @@ For instance, `<cap_tcb slot="1" pd="alpha">` will place the TCB of PD 'alpha' i
 The `io_address_space` element describes an address space used to isolate a given device.
 
 It supports the following attributes:
+
 * `name`: A unique name for the IO address space
 * `peripheral_id`: A unique identifier. This must match the identifier used by the hardware IOMMU or SMMU to identify devices.
+* `domain_id`: A unique domain identifier, synonymous with an ASID. This is not optional due to current
+                constraints with x86 only providing certain information at runtime.
 
 The `io_address_space` element supports the following elements as children:
 
 * `iomap`: This is used to map a *memory_region* into the IO address space.
 
 The `iomap` element supports the following attributes:
+
 * `mr`: Identifies the memory region to map.
 * `iovaddr`: Identifies the io virtual address at which to map the memory region.
 * `perms`: Identifies the permissions with which to map the memory region. Can be a combination of `r` (read), and `w` (write).
--- a/docs/manual.md
+++ b/docs/manual.md
@@ -109,7 +109,7 @@ This document attempts to clearly describe all of these terms, however as the co
 * [interrupt](#irq)
 * [fault](#fault)
 * [ioport](#ioport)
-* [io address space](#io_address_space)
+* [IO address space](#io_address_space)
 
 ## System {#system}
 
@@ -244,10 +244,10 @@ The mapping has a number of attributes, which include:
 * permissions (read, write and execute)
 
 *Note:* On x86 a memory region can also be *mapped* into one or more
-io address spaces. This type of mapping is known as an *iomap*. It supports
+IO address spaces. This type of mapping is known as an *iomap*. It supports
 a number of attributes, which include:
 
-* the io virtual address at which the region is mapped in the io address space
+* the io virtual address at which the region is mapped in the IO address space
 * permissions (read, write)
 
 **Note:** When a memory region is mapped into multiple protection
@@ -1093,12 +1093,12 @@ For instance, `<cap_tcb slot="1" pd="alpha">` will place the TCB of PD 'alpha' i
 The `io_address_space` element describes an address space used to isolate a given device.
 
 It supports the following attributes:
-* `name`: A unique name for the io address space
+* `name`: A unique name for the IO address space
 * `peripheral_id`: A unique identifier. This must match the identifier used by the hardware IOMMU or SMMU to identify devices.
 
 The `io_address_space` element supports the following elements as children:
 
-* `iomap`: This is used to map a *memory_region* into the io address space.
+* `iomap`: This is used to map a *memory_region* into the IO address space.
 
 The `iomap` element supports the following attributes:
 * `mr`: Identifies the memory region to map.
--- a/docs/manual.md
+++ b/docs/manual.md
@@ -109,6 +109,7 @@ This document attempts to clearly describe all of these terms, however as the co
 * [interrupt](#irq)
 * [fault](#fault)
 * [ioport](#ioport)
+* [io address space](#io_address_space)
 
 ## System {#system}
 
@@ -242,6 +243,13 @@ The mapping has a number of attributes, which include:
 * caching attributes (mostly relevant for device memory)
 * permissions (read, write and execute)
 
+*Note:* On x86 a memory region can also be *mapped* into one or more
+io address spaces. This type of mapping is known as an *iomap*. It supports
+a number of attributes, which include:
+
+* the io virtual address at which the region is mapped in the io address space
+* permissions (read, write)
+
 **Note:** When a memory region is mapped into multiple protection
 domains, the attributes used for different mappings may vary.
 
@@ -366,6 +374,12 @@ delivered to another PD, the fault being handled depends on when the parent PD i
 ## I/O Ports {#ioport}
 
 I/O ports are x86 mechanisms to access certain physical devices (e.g. PC serial ports or PCI) using the `in` and `out` CPU instructions. The system description specifies if a protection domain have access to certain port address ranges. These accesses will be executed by seL4 and the result returned to protection domains.
+
+## IO Address Spaces {#io_address_space}
+
+IO Address Spaces provide a way to isolate device memory accesses within a fixed virtual address space. The isolation provided by the address space is enforced by the underlying hardware IOMMU or SMMU.
+
+IO Address Spaces allow *memory regions* to be mapped to a provided base IO virtual address. These IO virtual addresses will be translated by the hardware IOMMU or SMMU to the underlying physical memory that backs the memory region.
 
 # SDK {#sdk}
 
@@ -1074,6 +1088,23 @@ See the 'cap_sharing' example packaged in your SDK or [on GitHub](https://github
 All capability elements (currently) all support the `pd` attribute, the name of the protection domain that the capability is from.
 For instance, `<cap_tcb slot="1" pd="alpha">` will place the TCB of PD 'alpha' in the CSpace of the current PD.
 
+## `io_address_space`
+
+The `io_address_space` element describes an address space used to isolate a given device.
+
+It supports the following attributes:
+* `name`: A unique name for the io address space
+* `peripheral_id`: A unique identifier. This must match the identifier used by the hardware IOMMU or SMMU to identify devices.
+
+The `io_address_space` element supports the following elements as children:
+
+* `iomap`: This is used to map a *memory_region* into the io address space.
+
+The `iomap` element supports the following attributes:
+* `mr`: Identifies the memory region to map.
+* `iovaddr`: Identifies the io virtual address at which to map the memory region.
+* `perms`: Identifies the permissions with which to map the memory region. Can be a combination of `r` (read), and `w` (write).
+           Defaults to read-write.
 
 ### Page sizes by architecture
 
--- a/tool/microkit/src/sdf.rs
+++ b/tool/microkit/src/sdf.rs
@@ -700,20 +700,12 @@ impl SysMap {
 
 impl SysIOMap {
     fn from_xml(
-        config: &Config,
+        _config: &Config,
         xml_sdf: &XmlSystemDescription,
         node: &roxmltree::Node,
         device: &str,
         identifier: IommuDeviceIdentifier,
     ) -> Result<SysIOMap, String> {
-        if config.arch != Arch::X86_64 {
-            return Err(value_error(
-                xml_sdf,
-                node,
-                "IOMMU is not yet supported on ARM and RISC-V by Microkit".to_string(),
-            ));
-        }
-
         let attrs = vec!["mr", "iovaddr", "perms"];
 
         check_attributes(xml_sdf, node, &attrs)?;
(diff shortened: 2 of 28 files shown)


midnightveil:
For comparison's sake: #467.

tool/microkit/src/sdf.rs
```diff
@@ -466,6 +694,145 @@ impl SysMap {
...
+        xml_sdf: &XmlSystemDescription,
+        node: &roxmltree::Node,
+        device: &str,
+        identifier: IommuDeviceIdentifier,
+    ) -> Result<SysIOMap, String> {
+        if config.arch != Arch::X86_64 {
+            return Err(value_error(
+                xml_sdf,
+                node,
+                "IOMMU isn't supported on ARM and RISC-V".to_string(),
```
midnightveil: at least for ARM, "isn't (yet) supported *by Microkit*".

tool/microkit/src/sdf.rs
```diff
@@ -466,6 +694,145 @@ impl SysMap {
...
+                    if !mapped_mrs.insert(iomap.mr.clone()) {
+                        return Err(value_error(
+                            xml_sdf,
+                            &child,
+                            format!(
+                                "device '{device_name}' maps memory region '{}' more than once",
+                                iomap.mr
+                            ),
+                        ));
+                    }
```
midnightveil: Is there any reason to prevent this?

i.e. is this a hardware limitation?

I don't see why it would be, you can map the same region multiple times for virtual address spaces, why not for IO address spaces?

tool/microkit/src/sdf.rs
```diff
@@ -1694,58 +1973,103 @@ struct XmlSystemDescription<'a> {
 pub struct SystemDescription {
     pub protection_domains: Vec<ProtectionDomain>,
     pub memory_regions: Vec<SysMemoryRegion>,
+    pub iomaps: Vec<SysIOMap>,
     pub channels: Vec<Channel>,
 }
 
-fn check_maps(
+// max_end is the first invalid virtual address
+fn check_maps<'a, M, I>(
```
midnightveil, line 2075: Could you please do this (and any other refactors) in a separate commit; just do a new commit for the refactor-to-be-more-generic, and then later add the new support?

Make this change first: adding support only for the singular map type, but using a generic, and then in the next commit add SysIOMap support.

This just makes it easier to see "is this code supposed to behave the same" vs "is this code doing new/different things".

tool/microkit/src/sdf.rs
```diff
@@ -93,13 +95,177 @@ fn loc_string(xml_sdf: &XmlSystemDescription, pos: roxmltree::TextPos) -> String
...
+        match config.arch {
+            Arch::X86_64 => PciDevice::from_str(s)
+                .map(IommuDeviceIdentifier::X86Pci)
+                .map_err(IommuDeviceIdentifierParseError::Pci),
+            Arch::Aarch64 | Arch::Riscv64 => Err(IommuDeviceIdentifierParseError::UnsupportedArch(
+                config.arch,
+            )),
+        }
+    }
+}
```
midnightveil, line 227: So, I think I know why you've done this: ARM uses SMMU which has Stream IDs, right?

Could you add just a brief comment in the code: especially around the SDF format (here it's more obvious why), why it's called "identifier"? I might also suggest "device:" or "peripheral:" , but maybe that's more confusing the identifier? Identifier is just very vague. 

Maybe peripheral_id?

(note: the name of this struct in Rust is fine, I'm thinking the SDF here)

tool/microkit/src/sdf.rs
```diff
@@ -111,6 +277,113 @@ pub struct SysMap {
     pub text_pos: Option<roxmltree::TextPos>,
 }
 
+#[derive(Debug, PartialEq, Eq, Clone, Copy)]
+pub enum SysIOMapPerms {
+    Read,
+    Write,
```
midnightveil, line 285: Does IOMMU hardware allow write-only? I know for page tables write-only doesn't tend to be allowed.

tool/microkit/src/sdf.rs
```diff
@@ -111,6 +277,113 @@ pub struct SysMap {
...
+    fn mr_name(&self) -> &str {
+        &self.mr
+    }
+
+    fn addr(&self) -> u64 {
+        self.vaddr
+    }
+
+    fn text_pos(&self) -> roxmltree::TextPos {
+        self.text_pos.unwrap()
```
midnightveil: if it's not actually optional don't make it optional :)

tool/microkit/src/sdf.rs
```diff
@@ -2234,11 +2573,26 @@ pub fn parse(
...
+            check_maps(
+                &xml_sdf,
+                &mrs,
+                vm.maps.iter(),
+                &format!("virtual machine '{}'", vm.name),
+                config.vm_map_max_vaddr(),
+            )?;
         }
     }
+    if !iomaps.is_empty() {
```
midnightveil: I think this if check is not necessary.

tool/microkit/src/sdf.rs
```diff
@@ -2283,7 +2637,7 @@ pub fn parse(
                     let pos = mr.text_pos.unwrap();
                     return Err(
                         format!(
-                            "Error: memory region '{}' physical address range [0x{:x}..0x{:x}) overlaps with another memory region '{}' [0x{:x}..0x{:x}) @ {}",
+                            "Error: memory region '{}' physical address range [{:#x}..{:#x}) overlaps with another memory region '{}' [{:#x}..{:#x}) @ {}",
```
midnightveil, line 2735: Cleanups like this please add to a separate cleanup commit.

midnightveil (commented):
OK, so so far this is good :)

There's a few relatively minor changes, but I think I'm happy with this format in the SDF.

I don't know if you've yet to do it, but documentation in the manual would be good.

(also like the rest of the implementation, ha).

tool/microkit/src/sdf.rs
cazb2: According to the kernel source it appears to: src/arch/x86/object/iospace.c:376-382

tool/microkit/src/sdf.rs
cazb2: I think it has to be optional according to someone else's comment on SysMap so I'll return it as an option.

tool/microkit/src/sdf.rs
cazb2: No hardware reason, it just seems like bad design to do that?

tool/microkit/src/sdf.rs
midnightveil: Hm. Do you know where in the Intel manual (which Translation table format) this is?

Linux did remove WO-only support as it gave them as CVE, lol: https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/commit/?id=c848416cc05afc1589edba04fe00b85c2f797ee3 / https://lkml.iu.edu/hypermail/linux/kernel/2102.3/01238.html, but I don't know what first-level or second-level means in this context (like, page table depth..?)

I see 'Format of SS-PTE that Maps 4-KByte Page' in the VT-d spec, which has bit 1: Write, bit 0: Read, though it does say the read is ignored sometimes for ZLR (whatever that is).

It might be fine, I'd just like to double check.

midnightveil: Oh, looks like 'first-level' refers to a newer feature which allows x86 cpus to use the same page table format as vaddr for iovaddr: https://lwn.net/Articles/807079/. 

OK, maybe not an issue then.

cazb2:
> OK, so so far this is good :)
> 
> There's a few relatively minor changes, but I think I'm happy with this format in the SDF.
> 
> I don't know if you've yet to do it, but documentation in the manual would be good.
> 
> (also like the rest of the implementation, ha).

Yep, I thought it would be easier to make a few smaller PRs rather than one big one.

tool/microkit/src/sdf.rs
cazb2: yep, and peripheral_id sounds good

tool/microkit/src/sdf.rs
midnightveil: I don't know either why you'd want to do that, but I also don't think it's worth the extra code to detect and handle it, and maybe there is a reason to do it that I don't know about, which someone would have to work around.

midnightveil:
Ok, last thing:

Can you add this sort of check? https://github.com/seL4/microkit/blob/b5cd889fad2472bb9d15f404b9d323b766726d47/tool/microkit/src/sdf.rs#L1141-L1147

So that until the IOMMU is enabled by default it should be impossible to make these.

(It will be based off the KernelHaveIOMMU define, like for hypervisor).

midnightveil (approved):

midnightveil:
Note: this is pending follow-up PRs for functionality, to be merged as a group.

docs/manual.md
```diff
@@ -242,6 +243,13 @@ The mapping has a number of attributes, which include:
 * caching attributes (mostly relevant for device memory)
 * permissions (read, write and execute)
 
+*Note:* On x86 a memory region can also be *mapped* into one or more
+io address spaces. This type of mapping is known as an *iomap*. It supports
```
dreamliner787-9:
```suggestion
IO address spaces. This type of mapping is known as an *iomap*. It supports
```
Can you please capitalise this since it's an abbreviation, need to be done elsewhere in the manual too, but leave `io_address_space` as is.

cazb2: should be done now

docs/manual.md
```diff
@@ -1074,6 +1088,25 @@ See the 'cap_sharing' example packaged in your SDK or [on GitHub](https://github
 All capability elements (currently) all support the `pd` attribute, the name of the protection domain that the capability is from.
 For instance, `<cap_tcb slot="1" pd="alpha">` will place the TCB of PD 'alpha' in the CSpace of the current PD.
 
+## `io_address_space`
+
+The `io_address_space` element describes an address space used to isolate a given device.
+
+It supports the following attributes:
+* `name`: A unique name for the IO address space
```
midnightveil, line 1097: These need a space before the bullet point lists otherwise they are not rendered correctly. (ditto for others)

tool/microkit/src/sdf.rs
```diff
@@ -470,6 +699,171 @@ impl SysMap {
...
+                format!("duplicate device name '{device_name}'"),
+            ));
+        }
+
+        // In microkit enforcing uniqueness for domain ids is hugely simplifying and enforces better designs.
+        // If we were to allow shared domain ids, we would have to ensure that whenever two devices share a domain
+        // that they do not have any IOVA to physyical address mappings that conflict, and that they share the same
+        // IOPT objects, this because the source identifier aka the PCI BDF is not used in all caches.
+        // See section 6.2.1 and table 17 in the intel VT-D manual.
+        // https://www.google.com/url?sa=t&source=web&rct=j&opi=89978449&url=https://cdrdv2-public.intel.com/671081/vt-directed-io-spec.pdf&ved=2ahUKEwiN_YawksWVAxUDmFYBHfzzCEQQFnoECA0QAQ&usg=AOvVaw3fO9wnba58SqsP2HNKwdfu
```
midnightveil: Please don't link a google link. Link the static Intel page and the full name of the manual.

tool/microkit/src/sdf.rs
```diff
@@ -470,6 +699,171 @@ impl SysMap {
...
+        let device_name = checked_lookup(xml_sdf, node, "name")?;
+        if !device_names.insert(device_name.to_string()) {
+            return Err(value_error(
+                xml_sdf,
+                node,
+                format!("duplicate device name '{device_name}'"),
+            ));
+        }
+
+        // In microkit enforcing uniqueness for domain ids is hugely simplifying and enforces better designs.
```
midnightveil: Please be a bit more formal with your comments. These sentences are very long and run on and could be a bit more brief.

docs/manual.md
cazb2: ah! fixed

midnightveil (commented):
Since you have commits in here that are a refactoring I have to review this at a per-commit level. However, as it stands you have fixup commits that have been added later that mean you change earlier code added. I'd appreciate if you cleaned up the commit history so that each change is standalone and you don't go back and change things you just added. You now have commits that contain changes of several previous commits.

cazb2:
> Since you have commits in here that are a refactoring I have to review this at a per-commit level. However, as it stands you have fixup commits that have been added later that mean you change earlier code added. I'd appreciate if you cleaned up the commit history so that each change is standalone and you don't go back and change things you just added. You now have commits that contain changes of several previous commits.

Right my bad. All new code from when it was reviewed previously is in separate commits from capitalising IO in the manual to the domain ID support. I haven't changed any of the previously reviewed commits but I get what you mean about GitHub being a pain for looking at individual commits.

midnightveil:
Follow-up PRs will happen soon.

PR #532, opened 2026-06-19
tool: create IPC buffer in the tool, not the ELF

midnightveil:
This removes the need for the linker script to define the IPC buffer.

Fixes: #451.

### DIFF
--- a/libmicrokit/src/main.c
+++ b/libmicrokit/src/main.c
@@ -39,9 +39,13 @@ seL4_Word microkit_notifications;
 seL4_Word microkit_pps;
 seL4_Word microkit_ioports;
 
-extern seL4_IPCBuffer __sel4_ipc_buffer_obj;
+#define BIT(n) (1ULL << (n))
+#define MASK(n) (BIT(n) - 1ULL)
 
-seL4_IPCBuffer *__sel4_ipc_buffer = &__sel4_ipc_buffer_obj;
+/* The tool assumes the IPC buffer in the top page of user memory */
+seL4_IPCBuffer *__sel4_ipc_buffer = (seL4_IPCBuffer *)(seL4_UserVSpaceTop & ~MASK(seL4_PageBits));
+_Static_assert(sizeof(seL4_IPCBuffer) <= BIT(seL4_PageBits),
+               "IPC Buffer is expected to need less than one page in size");
 
 extern const void (*const __init_array_start [])(void);
 extern const void (*const __init_array_end [])(void);
--- a/monitor/src/main.c
+++ b/monitor/src/main.c
@@ -29,8 +29,13 @@
 #define BASE_SCHED_CONTEXT_CAP 138
 #define BASE_NOTIFICATION_CAP 202
 
-extern seL4_IPCBuffer __sel4_ipc_buffer_obj;
-seL4_IPCBuffer *__sel4_ipc_buffer = &__sel4_ipc_buffer_obj;
+#define BIT(n) (1ULL << (n))
+#define MASK(n) (BIT(n) - 1ULL)
+
+/* The tool assumes the IPC buffer in the top page of user memory */
+seL4_IPCBuffer *__sel4_ipc_buffer = (seL4_IPCBuffer *)(seL4_UserVSpaceTop & ~MASK(seL4_PageBits));
+_Static_assert(sizeof(seL4_IPCBuffer) <= BIT(seL4_PageBits),
+               "IPC Buffer is expected to need less than one page in size");
 
 char pd_names[MAX_PDS][MAX_NAME_LEN];
 seL4_Word pd_names_len;
--- a/tool/microkit/src/capdl/builder.rs
+++ b/tool/microkit/src/capdl/builder.rs
@@ -31,9 +31,6 @@ use crate::{
     util::{ranges_overlap, round_down, round_up},
 };
 
-// Corresponds to the IPC buffer symbol in libmicrokit and the monitor
-const SYMBOL_IPC_BUFFER: &str = "__sel4_ipc_buffer_obj";
-
 const FAULT_BADGE: u64 = 1 << 62;
 const PPC_BADGE: u64 = 1 << 63;
 
@@ -187,8 +184,8 @@ impl CapDLSpecContainer {
 
     /// Add the details of the given ELF into the given CapDL spec while inferring as much information
     /// as possible. These are the objects that will be created:
-    /// -> TCB: PC, SP and IPC buffer vaddr set. VSpace and IPC buffer frame caps bound.
-    /// -> VSpace: all ELF loadable pages and IPC buffer mapped in.
+    /// -> TCB: Program counter set and VSpace capability bound.
+    /// -> VSpace: all pages from the ELF mapped in.
     /// Returns the object ID of the TCB
     /// NOTE that all ELF frames will just be reference to the original ELF object rather than the actual data.
     /// So that symbols can be patched before the frames' data are filled in.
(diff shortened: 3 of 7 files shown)


midnightveil:
OK, figured out the issue: https://github.com/seL4/seL4/issues/1693. Annoyingly this complicates things, but oh well.

libmicrokit/src/main.c
```diff
@@ -3,6 +3,7 @@
  *
  * SPDX-License-Identifier: BSD-2-Clause
  */
+#include <assert.h>
```
midnightveil: Ok, based on compule failures I should just use _Static_assert

libmicrokit/src/main.c
```diff
@@ -39,9 +40,19 @@ seL4_Word microkit_notifications;
 seL4_Word microkit_pps;
 seL4_Word microkit_ioports;
 
-extern seL4_IPCBuffer __sel4_ipc_buffer_obj;
+// Workaround: https://github.com/seL4/seL4/issues/1693
+#if seL4_UserTop == 0x00ffffffffff || seL4_UserTop == 0x0fffffffffff || seL4_UserTop == 0x7fffffffffff
```
midnightveil: Ok, since per https://github.com/seL4/seL4/issues/1693 we should expect this to become inclusive in future we should make this assume UserTop is inclusive.

tool/microkit/src/sel4.rs
```diff
@@ -310,16 +310,18 @@ pub struct Config {
 }
 
 impl Config {
-    /// Refers to 'seL4_UserTop'
+    /// Refers to 'seL4_UserTop'. Is exclusive.
     pub fn user_top(&self) -> u64 {
         match self.arch {
+            // Almost the same as the true definition, but for an inconsistency
+            // TODO: Resolve https://github.com/seL4/seL4/issues/1693
```
midnightveil: Ok, since per https://github.com/seL4/seL4/issues/1693 we should expect this to become inclusive in future we should make this assume UserTop is inclusive.

tool/microkit/src/capdl/builder.rs
dreamliner787-9: The function doc should be updated when the behaviour changes:

```diff
    /// Add the details of the given ELF into the given CapDL spec while inferring as much information
    /// as possible. These are the objects that will be created:
-    /// -> TCB: PC, SP and IPC buffer vaddr set. VSpace and IPC buffer frame caps bound.
+    /// -> TCB: PC and SP vaddr set. VSpace frame cap bound.
-    /// -> VSpace: all ELF loadable pages and IPC buffer mapped in.
+    /// -> VSpace: all ELF loadable pages mapped in.
```

tool/microkit/src/capdl/builder.rs
```diff
@@ -488,6 +445,27 @@ pub fn build_capdl_spec(
...
+    map_page(
+        &mut spec_container,
+        kernel_config,
+        MONITOR_PD_NAME,
+        mon_vspace_obj_id,
+        mon_ipcbuf_frame_cap.clone(),
+        PageSize::Small as u64,
+        kernel_config.pd_ipc_buffer(),
+    )
+    .expect("should be able to map the IPC buffer as we checked overlaps");
```
dreamliner787-9:
```diff
-    .expect("should be able to map the IPC buffer as we checked overlaps");
+    .expect("should be able to map the IPC buffer as we checked overlaps in sel4.rs");
```

tool/microkit/src/capdl/builder.rs
```diff
@@ -663,7 +657,32 @@ pub fn build_capdl_spec(
...
+        map_page(
+            &mut spec_container,
+            kernel_config,
+            &pd.name,
+            pd_vspace_obj_id,
+            ipcbuf_frame_cap.clone(),
+            PageSize::Small as u64,
+            kernel_config.pd_ipc_buffer(),
+        )
+        .expect("should be able to map the IPC buffer as we checked overlaps");
```
dreamliner787-9:
```diff
-    .expect("should be able to map the IPC buffer as we checked overlaps");
+    .expect("should be able to map the IPC buffer as we checked overlaps in sel4.rs");
```

monitor/src/main.c
```diff
@@ -29,8 +29,19 @@
...
-extern seL4_IPCBuffer __sel4_ipc_buffer_obj;
-seL4_IPCBuffer *__sel4_ipc_buffer = &__sel4_ipc_buffer_obj;
+// Workaround: https://github.com/seL4/seL4/issues/1693
+#if seL4_UserTop == 0x00ffffffffff || seL4_UserTop == 0x0fffffffffff || seL4_UserTop == 0x7fffffffffff
+#define user_top_aligned (seL4_UserTop + 1)
+#elif (seL4_UserTop & ((1ULL << seL4_PageBits) - 1)) != 0
+#error "Expects seL4_UserTop to be aligned"
+#else
+#define user_top_aligned (seL4_UserTop)
+#endif
```
dreamliner787-9: Suggestion: can we fix this in the kernel first? I dislike having workarounds like this, because then we have to go back and fix it later.

dreamliner787-9 (changes requested):
A couple of small things from me but other than that looks good.

monitor/src/main.c
midnightveil: It depends on proof updates to seL4, as the values are used inside seL4 implementation.

I suppose @lsf37 can comment if it would be possible to fix https://github.com/seL4/seL4/issues/1693 before the next release.


---

We could do something hacky where we export a different value than what seL4 uses in the code, and so we make userspace consistent whilst leaving the checks the same. I don't suppose the semantics extraction is clever enough to do constant folding, would it?

e.g. if we changed a seL4_UserTop that is 0x800000000 to 0x7fffffffffff and then made the check inside the vspace map functions do `(>= (seL4_UserTop + 1))`. Whether it's smart enough to constant fold and not affect the proofs?

But I don't know if it's worth doing that as opposed to just fixing it... (it's surely just changing `<` to `<=` in a few places?).

lsf37: The semantics extraction does not do constant folding as such, but a lot of the times it is done implicitly in the proofs, so it would be possible to be lucky on that one.

I currently don't have it on the list for the release, because I wanted to focus on bugs, but it would not be hard to add once we have decided what to do. The tricky part is social: technically this changes the API and is a breaking change. I don't really want to write an entire RFC for this, but a few more opinions would be good.

lsf37: > But I don't know if it's worth doing that as opposed to just fixing it... (it's surely just changing < to <= in a few places?).

In the code, yes. In the proofs it's slightly more than that, because one version needs a proof of absence of overflow and the other does not. It's small, but it's not automatic.

tool/microkit/src/sel4.rs
```diff
@@ -346,30 +348,39 @@ impl Config {
         }
     }
 
+    /// IPC Buffer is located at the highest possible virtual memory page
+    pub fn pd_ipc_buffer(&self) -> u64 {
+        // user_top is inclusive, so we mask off the bits inside the page.
+        // self.user_top() & !util::mask(PageSize::Small.fixed_size_bits(self))
```
midnightveil: Just realised this should use IPCBufferBits. It's the same but its qhat we want.

midnightveil: .. Actually, it's not. Because then it won't be page base.

midnightveil:
This is now just blocked waiting on this update to the microkit-manifest: https://github.com/seL4/microkit-manifest/pull/3

It now uses user_vspace_top everywhere.

PR #533, opened 2026-06-19
docs: trivial fixes

yf13:
Some trivial fixes in manual.md like:

- Add missing break in 10.2 so that list is correctly rendered in PDF version.
- Add missing . after a sentence.

### DIFF
--- a/docs/manual.md
+++ b/docs/manual.md
@@ -1811,6 +1811,7 @@ At build-time, the Microkit tool embeds the capDL specification that describe
 all kernel objects that needs to be created. Then for each kernel object, the spec
 describe what state they need to be in and what capabilities exist to that object
 (i.e. who has access to this kernel object). For example, the spec would specify the:
+
 - starting Instruction Pointer (IP), Stack Pointer (SP) and IPC buffer pointer of a Thread Control Block (TCB),
 - page table structure and mapping attributes of an address space (VSpace),
 - interrupts (IRQ),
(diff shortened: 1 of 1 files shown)


lsf37:
These files are generated. Has the same change been made upstream?

midnightveil:
> These files are generated. Has the same change been made upstream?

Unless I'm mistaken this is the upstream and seL4 docs sources off the microkit manual?

docs/manual.md
```diff
@@ -1177,9 +1177,9 @@ To boot a Microkit image, use a GDB prompt via openOCD per [seL4 Docs](https://d
 
 2. Set the `a0` and `a1` registers for OpenSBI
 
-        # tell OpenSBI where DTB is (there is none)
+        (gdb) # tell OpenSBI where DTB is (there is none)
```
midnightveil: I'd rather these be placed in backticks ``` and indentation removed, I think.

IDK why the markdown parswr would be parsing headings when its indented for a coxe block...

midnightveil: It seems that the manual PDF and GitHub both parse this correctly?

I don't understand this change.

lsf37:
> > These files are generated. Has the same change been made upstream?
> 
> Unless I'm mistaken this is the upstream and seL4 docs sources off the microkit manual?

Wow, yes, I don't know how I hallucinated the docs repo here. Apologies. This is absolutely the right repo for this change.

docs/manual.md
lsf37: Standard markdown specs would definitely not parse this as heading. Indented code blocks are not an exotic feature, but putting it into a fenced code block like @midnightveil suggests would fix any ambiguity if you want to be extra sure.

Adding `(gdb)` seems not good to me, because `(gdb)` is the prompt and the line is supposed to be a comment between prompts.

yf13: It turned out that I was misled by the outline plugin in my text editor:

image

Anyway, I can ignore it now as it looks like a bug of that plugin.

I've updated the commit to keep original form as it is follows markdown rule.

PR #534, opened 2026-06-22
tool: add --viper-output option

zaklogician:
As part of a series of patches which upstream Viper export to Microkit, we add the `--viper-output VIPER_OUTPUT_DIR` command line option to the Microkit tool.

If this option is set, then after the systems is built, the tool will emit `VIPER_OUTPUT_DIR/name.vpr` for each protection domain `name` specified in the system description file. The output consists of a description of the capability table of the given PD in the [Viper](https://www.pm.inf.ethz.ch/research/viper.html) verification language, generated from the same output that the CapDL loader uses. This makes the capability tables available for verification tasks.

### DIFF
--- a/docs/manual.md
+++ b/docs/manual.md
@@ -482,6 +482,12 @@ It is important to note that this is largely just for debugging purposes and the
 highly tied to a specific version and configuration of the kernel. When using this option the kernel
 should be the same version and compiled with the same configuration options.
 
+If the `--viper-output PREFIX` argument is set, then for each protection domain `name` specified in
+the system description file, a file `PREFIX/name.vpr` will be output, containing a description of the
+capability table of the given PD in the Viper verification language. These output files can be used
+for verification purposes, but the exact format may change in future versions of the tool.
+
+
 ## Image format
 
 ### ARM
--- a/example/hello/Makefile
+++ b/example/hello/Makefile
@@ -67,4 +67,4 @@ $(BUILD_DIR)/hello.elf: $(addprefix $(BUILD_DIR)/, $(HELLO_OBJS))
 	$(LD) $(LDFLAGS) $^ $(LIBS) -o $@
 
 $(IMAGE_FILE) $(REPORT_FILE): $(addprefix $(BUILD_DIR)/, $(IMAGES)) hello.system
-	$(MICROKIT_TOOL) hello.system --search-path $(BUILD_DIR) --board $(MICROKIT_BOARD) --config $(MICROKIT_CONFIG) -o $(IMAGE_FILE) -r $(REPORT_FILE)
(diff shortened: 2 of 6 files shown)


zaklogician:
Not sure why the DCO check fails, it looks like all new commits are signed off.

I expect a possible `rustfmt` failure.

midnightveil:
If you click on DCO it tells you why:


> Summary
> 
> Commit sha: [0e4d426](https://github.com/seL4/microkit/pull/534/commits/0e4d4261022da97bf2b2012cad55308451b2912e), Author: Zoltan A. Kocsis, Committer: Zoltan A. Kocsis; Expected "Zoltan A. Kocsis [11808286+zaklogician@users.noreply.github.com](mailto:11808286+zaklogician@users.noreply.github.com)", but got "Ivan Velickovic [i.velickovic@unsw.edu.au](mailto:i.velickovic@unsw.edu.au)".

Though admittedly I don't know why that commit exists still

zaklogician:
Okay, the DCO issue should be fixed now.

midnightveil (approved):
LGTM. Few extra things to fix up but once done will merge.

lsf37:
Not sure why GitHub is suggesting the github no-reply address, but I should point out that the no-reply address is generally not an acceptable DCO sign-off (unless the change is trivial in which case we don't really care). The DCO is a legal copyright declaration, so it should ideally be from the employer email address if it is done under an employer who has copyright claims, or at least some other contactable email address.

zaklogician:
@lsf37 Thanks for letting me know. I will need a source from an official place which states that, otherwise I'll leave it as it is.

lsf37 (changes requested):
https://sel4.systems/Contribute/dco.html

lsf37 (approved):
Looks like the DCO checker is just dumb. I'd recommend using the same address as the committer just to avoid the annoyance in the future, but the DCO requirement as such is fine. I can set it to passing manually.

midnightveil:
I've just force pushed and made the sign-off/committer me because it was complaining about that (Zoltan does not wish to link UNSW email to GitHub for authorship)

If you're happy to press the DCO bypass button again I can make the sign-off Zoltan's again, but I don't know if it matters that much.

lsf37:
> I've just force pushed and made the sign-off/committer me because it was complaining about that (Zoltan does not wish to link UNSW email to GitHub for authorship)
> 
> If you're happy to press the DCO bypass button again I can make the sign-off Zoltan's again, but I don't know if it matters that much.

You signing off on behalf of Zoltan is fine, the DCO explicitly allows that.

lsf37:
@zaklogician not sure what your reason is, but if it is spam protection, an option might be to set up an alias or another address that you can filter on. The requirement is that it is contactable, it doesn't have to be real name (or from the employer, even though that would be preferable). The problem with GitHub's no-reply address is the "no reply" part of it.

PR #531, opened 2026-06-19
Update rust-sel4 commit from TS fork to mainline

dreamliner787-9:
This required a few changes to how the spec is packed into the initialiser ELF from https://github.com/seL4/rust-sel4/pull/345.

Do not merge until:
- [x] https://github.com/seL4/rust-sel4/pull/350 is merged
- [x] rust-sel4 commit is updated in Cargo.toml
- [x] Object allocation simulator in allocation.rs is updated with the new strategy

PR #535, opened 2026-06-22
Fix approach used for making PDs passive

### LINKED ISSUE #91: Fix approach used for making PDs passive at run-time

dreamliner787-9:
Depends on:
- https://github.com/seL4/microkit/pull/531
- https://github.com/seL4/rust-sel4/pull/352

Fixes https://github.com/seL4/microkit/issues/91.

### DIFF
--- a/libmicrokit/include/microkit.h
+++ b/libmicrokit/include/microkit.h
@@ -19,7 +19,7 @@ typedef unsigned int microkit_child;
 typedef unsigned int microkit_ioport;
 typedef seL4_MessageInfo_t microkit_msginfo;
 
-#define MONITOR_EP 5
+/* Index 5 is unused */
 /* Only valid in the 'benchmark' configuration */
 #define TCB_CAP 6
 /* Only valid when the PD has been configured to make SMC calls */
--- a/tool/microkit/src/capdl/builder.rs
+++ b/tool/microkit/src/capdl/builder.rs
@@ -74,16 +74,13 @@ const MON_FAULT_EP_CAP_IDX: u64 = 1;
 const MON_REPLY_CAP_IDX: u64 = 2;
 const MON_BASE_PD_TCB_CAP: u64 = 10;
 const MON_BASE_VM_TCB_CAP: u64 = MON_BASE_PD_TCB_CAP + 64;
-const MON_BASE_SCHED_CONTEXT_CAP: u64 = MON_BASE_VM_TCB_CAP + 64;
-const MON_BASE_NOTIFICATION_CAP: u64 = MON_BASE_SCHED_CONTEXT_CAP + 64;
 
 // Where caps must be in a PD's CSpace
 const PD_INPUT_CAP_IDX: u64 = 1;
 const PD_FAULT_EP_CAP_IDX: u64 = 2;
 const PD_VSPACE_CAP_IDX: u64 = 3;
 const PD_REPLY_CAP_IDX: u64 = 4;
-// Valid only if the PD is passive.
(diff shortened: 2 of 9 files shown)


tool/microkit/src/capdl/builder.rs
```diff
@@ -1043,29 +1028,13 @@ pub fn build_capdl_spec(
 
         // Step 3-15 bind this PD's TCB to the monitor, this accomplish two purposes:
         // 1. Allow PDs' TCBs to be named to their proper name in SDF in debug config.
```
midnightveil, line 1030: Not really relevant, but surely the capDL initialiser should also be doing this?

libmicrokit/include/microkit.h
```diff
@@ -19,7 +19,7 @@ typedef unsigned int microkit_child;
 typedef unsigned int microkit_ioport;
 typedef seL4_MessageInfo_t microkit_msginfo;
 
-#define MONITOR_EP 5
+/* Index 5 is unused */
```
midnightveil, line 22: I think this is probably unnecessary given all the other indices that are unused.

tool/microkit/src/capdl/builder.rs
dreamliner787-9: It actually [already does that](https://github.com/seL4/rust-sel4/blob/c30d0e44fb32c15239049ce402ef2a8c72499aa3/crates/sel4-capdl-initializer/src/initialize.rs#L881), but based on the name of the TCB object rather than the PD name. In Microkit the TCB name is prefixed with `tcb_` then PD name. If this isn't a big problem then we can strip out the debug name entirely from Microkit.

libmicrokit/include/microkit.h
dreamliner787-9: I added that to signal to anyone looking at the code later that the index is free to re-use, since previously it was occupied. But I can remove it

tool/microkit/src/capdl/builder.rs
midnightveil: I think we should probably strip out the Microkit duplication of this work, yes.

dreamliner787-9: Yea I can remove it in another PR

midnightveil:
This needs to be rebased on mainline.

PR #536, opened 2026-06-25
Add bootinfo tag

### LINKED ISSUE #525: x86: add a way for a PD to query the TSC frequency
On x86, the TSC is readable by userspace and you can easily derive its frequency by reading the relevant CPUID leafs. However, in some environments (e.g. virtualisation), those CPUID leafs won't always be enumerated and you won't be able to derive the TSC frequency in userspace.

However, the kernel measures the TSC frequency at boot with the PIT, then exposes it as a BootInfo entry. We should add a way to patch this information into certain PDs to make it easier to use the TSC as a fast and accurate clock source (e.g. for userspace timer drivers or VMMs).

It could look like this:
```xml
<system>
    <protection_domain name="timer_driver" priority="200" >
        <program_image path="timer.elf" />
        <setvar x86_tsc_freq_symbol="tsc_freq" />
    </protection_domain>
</system>
```

terryzbai:
~~This PR is rebased on branch [rust_sel4_update_to_main](https://github.com/au-ts/microkit/tree/rust_sel4_update_to_main), which should be merged soon. The proposed feature is included in the only last 3 commits.~~ (merged)

This PR adds the `bootinfo` tag to map specified type of bootinfo into PD's vspace, and patch the address to pre-declared variable `bootinfo_{bi_type}`. This is useful for the cases where the userspace applications needs to consume bootinfo, e.g., RSDP in ACPI driver, [TSC used as clock source](https://github.com/seL4/microkit/issues/525), etc.

The usage of `bootinfo` tag is added in the example `x86_64_ioport`. It works with [this rust-sel4 PR](https://github.com/seL4/rust-sel4/pull/355) on x86 to show how TSC frequency is mapped and read by a PD.

(On hold, waiting [rust-sel4 PR](https://github.com/seL4/rust-sel4/pull/355) to be merged)

### DIFF
--- a/docs/manual.md
+++ b/docs/manual.md
@@ -255,6 +255,17 @@ a size isn't specified, the memory region will be sized by the length
 of the prefill file, rounded up to the smallest page size or the user
 specified page size.
 
+A *memory region* can also be prefilled with bootinfo by the capDL Initialiser
+via the parameter `prefill_bootinfo` in the [System Description File](#sysdesc).
+If a size isn't specified, the memory region defaults to a single smallest page.
+The supported bootinfo types include:
+
+* `x86_vbe`
+* `x86_mbmap`
+* `x86_acpi_rsdp`
+* `x86_framebuffer`
+* `x86_tsc_freq`
+
 ## Channels {#channels}
 
 A *channel* enables two protection domains to interact using protected procedures or notifications.
@@ -1037,6 +1048,7 @@ It supports the following attributes:
 * `page_size`: (optional) Size of the pages used in the memory region; must be a supported page size if provided. Defaults to the largest page size for the target architecture that the memory region is aligned to.
 * `phys_addr`: (optional) The physical address for the start of the memory region (must be a multiple of the page size).
 * `prefill_path`: (optional) Path to a file containing data that the memory region will be filled with at initialisation.
+* `prefill_bootinfo`: (optional) Bootinfo type that the memmory region will be filled with by the capDL initialiser.
 
 The `memory_region` element does not support any child elements.
 
--- a/example/bootinfo/README.md
+++ b/example/bootinfo/README.md
@@ -1,0 +1,30 @@
+<!--
+     Copyright 2026, UNSW
+     SPDX-License-Identifier: CC-BY-SA-4.0
+-->
+# Example - Hello World
+
+This is a basic example that demonstrates how one can define
+a MemoryRegion prefilled with BootInfo and map it to a PD, so
+the PD can read the BootInfo in userland.
+
+Supported BootInfo includes:
+- x86_vbe
+- x86_mbmap
+- x86_acpi_rsdp
+- x86_framebuffer
+- x86_tsc_freq
+
+As Microkit loader does not pass the DTB through on ARM and RISC-V,
+so only x86 is supported in this example for now.
+
+## Building
+
+```sh
+mkdir build
+make BUILD_DIR=build MICROKIT_BOARD=<board> MICROKIT_CONFIG=<debug/release/benchmark> MICROKIT_SDK=/path/to/sdk qemu
+```
+
+## Running
+
+See instructions for your board in the manual.
--- a/example/bootinfo/bootinfo.system
+++ b/example/bootinfo/bootinfo.system
@@ -1,0 +1,13 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+ Copyright 2026, UNSW
+
+ SPDX-License-Identifier: BSD-2-Clause
+-->
+<system>
+    <memory_region name="bootinfo_x86_tsc_freq" prefill_bootinfo="x86_tsc_freq" />
+    <protection_domain name="bootinfo">
+        <program_image path="bootinfo.elf" />
+        <map mr="bootinfo_x86_tsc_freq" vaddr="0x2_000_000" perms="r" cached="true" setvar_vaddr="bootinfo_tsc_freq" />
+    </protection_domain>
+</system>
--- a/tool/microkit/src/capdl/builder.rs
+++ b/tool/microkit/src/capdl/builder.rs
@@ -11,8 +11,8 @@ use std::{
 };
 
 use sel4_capdl_initializer_types::{
-    object, CapTableEntry, Fill, FillEntry, FillEntryContent, NamedObject, Object, ObjectId, Spec,
-    Word,
+    object, CapTableEntry, Fill, FillEntry, FillEntryContent, FillEntryContentBootInfo,
+    NamedObject, Object, ObjectId, Spec, Word,
 };
 
 use crate::{
@@ -513,8 +513,9 @@ pub fn build_capdl_spec(
                 .paddr()
                 .map(|base_paddr| Word(base_paddr + (frame_sequence * mr.page_size_bytes())));
 
+            let starting_byte_idx = frame_sequence * mr.page_size_bytes();
+
             let frame_fill = if let Some(prefill_bytes) = &mr.prefill_bytes {
-                let starting_byte_idx = frame_sequence * mr.page_size_bytes();
                 let remaining_bytes_to_fill = prefill_bytes.len() as u64 - starting_byte_idx;
                 let num_bytes_to_fill = min(mr.page_size_bytes(), remaining_bytes_to_fill);
 
@@ -535,6 +536,22 @@ pub fn build_capdl_spec(
                         })),
                     }]
                     .to_vec(),
+                }
+            } else if let Some(prefill_bootinfo) = mr.prefill_bootinfo {
+                let remaining_bytes_to_fill = mr.size - starting_byte_idx;
+                let num_bytes_to_fill = min(mr.page_size_bytes(), remaining_bytes_to_fill);
+
+                FrameFill {
+                    entries: vec![FillEntry {
+                        range: Range {
+                            start: 0,
+                            end: num_bytes_to_fill,
+                        },
+                        content: FillEntryContent::BootInfo(FillEntryContentBootInfo {
+                            id: prefill_bootinfo,
+                            offset: starting_byte_idx,
+                        }),
+                    }],
                 }
             } else {
                 FrameFill {
--- a/tool/microkit/src/sdf.rs
+++ b/tool/microkit/src/sdf.rs
@@ -19,8 +19,10 @@
 use crate::sel4::{
     Arch, ArmRiscvIrqTrigger, Config, PageSize, X86IoapicIrqPolarity, X86IoapicIrqTrigger,
 };
+
 use crate::util::{get_full_path, ranges_overlap, round_up, str_to_bool};
 use crate::MAX_PDS;
+use sel4_capdl_initializer_types::FillEntryContentBootInfoId;
 use std::collections::{HashMap, HashSet};
 use std::fmt::Display;
 use std::fs;
@@ -116,6 +118,7 @@ pub enum SysMemoryRegionKind {
     User,
     Elf,
     Stack,
+    BootInfo,
 }
 
 #[derive(Debug, PartialEq, Eq, Clone)]
@@ -141,6 +144,7 @@ pub struct SysMemoryRegion {
     /// stack, ELF, etc.
     pub kind: SysMemoryRegionKind,
     pub prefill_bytes: Option<Vec<u8>>,
+    pub prefill_bootinfo: Option<FillEntryContentBootInfoId>,
 }
 
 impl SysMemoryRegion {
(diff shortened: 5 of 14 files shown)


tool/microkit/src/sdf.rs
```diff
@@ -1419,6 +1452,36 @@ impl CSpace {
...
+
+        let xml_bi_type = checked_lookup(xml_sdf, node, "type")?;
+        let bi_type = match xml_bi_type {
+            "vbe" => FillEntryContentBootInfoId::X86Vbe,
+            "mbmmap" => FillEntryContentBootInfoId::X86Mbmmap,
+            "rsdp" => FillEntryContentBootInfoId::X86AcpiRsdp,
+            "framebuffer" => FillEntryContentBootInfoId::X86FrameBuffer,
+            "tsc_freq" => FillEntryContentBootInfoId::X86TscFreq,
+            _ => return Err(format!("BootInfoMap type: '{xml_bi_type}' is not supported.")),
+        };
```
midnightveil: This should be consistent with the naming scheme of the enum/seL4 Bootinfo items - as in keep the prefix.

tool/microkit/src/sdf.rs
```diff
@@ -1419,6 +1452,36 @@ impl CSpace {
...
+        println!("Detected BootInfo type '{}'", xml_bi_type);
```
midnightveil: Please remember to remove your prints from the code.

tool/microkit/src/sdf.rs
```diff
@@ -1178,6 +1190,26 @@ impl ProtectionDomain {
 
                     cspace = Some(CSpace::from_xml(xml_sdf, &child)?);
                 }
+                "bootinfo" => {
+                    let bi_map = BootInfoMap::from_xml(&mut mem_top, xml_sdf, &child)?;
+                    for existing_bi_map in &bootinfo_maps {
+                        if existing_bi_map.bi_type == bi_map.bi_type {
+                            let pos = xml_sdf.doc.text_pos_at(child.range().start);
+                            return Err(format!(
+                                "Error: duplicate {:?} in protection domain '{}' @ {}",
```
midnightveil: Unless rust-seL4 actively prevents this I don't really see a reason to prevent duplicates. It adds extra code that isn't useful.

tool/microkit/src/sdf.rs
```diff
@@ -644,6 +654,8 @@ impl ProtectionDomain {
         let mut ioports = Vec::new();
         let mut setvars: Vec<SysSetVar> = Vec::new();
         let mut child_pds = Vec::new();
+        let mut bootinfo_maps: Vec<BootInfoMap> = Vec::new();
+        let mut mem_top = config.pd_map_max_vaddr(stack_size);
```
midnightveil: I don't link this very much. The implicit allocation adds a whole bunch of extra cases to our logic about overmapping, which makes things difficult.

Could the bootinfo not just be another `<mr>` that one specifies and then we can place bootinfo "inside" of it?

(You could then in theory map it in RO to multiple PDs without needing to copy it many times - for instance, if we wanted to map in the FDT it would probably be bad to copy it into every PD)

tool/microkit/src/sel4.rs
```diff
@@ -584,3 +584,21 @@ impl X86IoapicIrqPolarity {
         }
     }
 }
+
+#[repr(u64)]
+#[derive(Debug, Copy, Clone, Eq, PartialEq)]
+pub enum BootInfoId {
```
midnightveil: Isn't there already the rust-sel4 `FillEntryContentBootInfoId` which you use?

tool/microkit/src/sdf.rs
midnightveil: We could make the `SysMemoryRegion` have the kind be User/Elf/Stack/Prefill(with a buffer)/Bootinfo (type)` and then these would all be consistently handled

tool/microkit/src/sdf.rs
terryzbai: "No duplicates" assumption is used for patching the address of bootinfo into a PD via `setvar`? and there is no use cases that need duplicate bootinfo I think.

midnightveil: Oh, is the name of the setvar implicit? I really don't like that.

> no use cases that need duplicate bootinfo I think.

That doesn't mean there couldn't be one, and it's more effort to prevent it.

tool/microkit/src/sel4.rs
terryzbai: uh my bad. I forgot to remove them after replacing with `FillEntryContentBootInfoId`.

tool/microkit/src/sdf.rs
terryzbai: you mean the whole `sel4_bootinfo_header_x86_vbe` or `x86_vbe`?

midnightveil: x86_vbe.

tool/microkit/src/sdf.rs
terryzbai: Yes, that's true. I didn't realise that we already had prefilled MR.

tool/microkit/src/sdf.rs
terryzbai: yes, it is implicit. It will be solved if we use prefilled MRs.

> That doesn't mean there couldn't be one, and it's more effort to prevent it.

right, so prefilled MR is indeed a better solution in this case.

midnightveil: Well, we can't use "prefilled" MRs directly, but if we had a mechanism closer to it.

terryzbai: Are you saying we still need a `bootinfo` tag outside of PDS? Can I know what's your concern of adding an optional parameter in prefilled MRs?

midnightveil: Prefilled MRs means an MR prefilled from a file.

This isn't "prefilled" in the same way as other prefilled, but it's a *similar* mechanism.

example/x86_64_ioport/x86_64_ioport.c
```diff
@@ -31,6 +37,10 @@ void init(void)
     microkit_dbg_puts(microkit_name);
     microkit_dbg_puts("\n");
 
+    microkit_dbg_puts("TSC Frequency: ");
+    microkit_dbg_put32(bootinfo_tsc_freq->value);
+    microkit_dbg_puts("MHz\n");
+
```
dreamliner787-9: Can we please split this into a new example called `bootinfo` or something like that? That comes with the benefit of being able to target other archs as well. As this example is just meant to show x86 PIO operations.

dreamliner787-9 (dismissed):

example/x86_64_ioport/x86_64_ioport.c
terryzbai: yes, I have added a separate example for this. Do you have any ideas of how to demonstrate the ability to read DTB on ARM?

midnightveil: we can probably quite easily parae the fdt header and at the very least verify the magic bytes and print the version, without having to write an entire fdt library.

edit: although saying that, microkit loader does not pass through the dtb and always passes 0,0 for the dtb, so.

example/bootinfo/bootinfo.system
```diff
@@ -0,0 +1,13 @@
...
+<!--
+ Copyright 2026, UNSW
+
+ SPDX-License-Identifier: BSD-2-Clause
+-->
+<system>
+    <memory_region name="bootinfo_x86_tsc_freq" prefill_bootinfo="x86_tsc_freq" />
+    <protection_domain name="bootinfo">
+        <program_image path="bootinfo.elf" />
+        <map mr="bootinfo_x86_tsc_freq" vaddr="0x2_000_000" perms="r" cached="false" setvar_vaddr="bootinfo_tsc_freq" />
```
midnightveil: This should be cached="true". There's almost no reason for this to be uncached, as it usually leads to cache coherency issues.

example/bootinfo/README.md
```diff
@@ -0,0 +1,31 @@
...
+a MemoryRegion prefilled with BootInfo and map it to a PD, so 
+the PD can read the BootInfo in userland.
+
+Supported BootInfo includes:
+- x86_vbe
+- x86_mbmap
+- x86_acpi_rsdp
+- x86_framebuffer
+- x86_tsc_freq
+- fdt
```
midnightveil: I might remove this from these lists given it won't work

tool/microkit/src/sdf.rs
```diff
@@ -1540,7 +1549,24 @@ impl SysMemoryRegion {
...
+                "x86_acpi_rsdp" => Ok(FillEntryContentBootInfoId::X86AcpiRsdp),
+                "x86_framebuffer" => Ok(FillEntryContentBootInfoId::X86FrameBuffer),
+                "x86_tsc_freq" => Ok(FillEntryContentBootInfoId::X86TscFreq),
+                "fdt" => Ok(FillEntryContentBootInfoId::Fdt),
+                _ => return Err(format!("BootInfoMap type: '{xml_bi_type}' is not supported.")),
+            }
+        })
+        .transpose()?;
+
+        println!("prefill_bootinfo: {:?}", prefill_bootinfo_maybe);
```
midnightveil: printf to remove

docs/manual.md
```diff
@@ -255,6 +255,17 @@ a size isn't specified, the memory region will be sized by the length
 of the prefill file, rounded up to the smallest page size or the user
 specified page size.
 
+A *memory region* can also be prefilled with bootinfo by the capDL Initialiser
```
midnightveil, line 258: You also need to add it to the memory_region element example, which describes the syntax (see e.g. prefill_path)

tool/microkit/src/sdf.rs
```diff
@@ -1540,7 +1549,24 @@ impl SysMemoryRegion {
             })
             .transpose()?;
 
-        let size = Self::determine_size(xml_sdf, node, &prefill_bytes_maybe, page_size)?;
+        let prefill_bootinfo_maybe = node.attribute("prefill_bootinfo").map(|xml_bi_type| {
```
midnightveil: Do we support both prefill_bytes and prefill_bootinfo both specified?

I think that won't work, right? So we should make that error.

midnightveil (approved):
Approving, but this needs the dependant rust-sel4 changes merged first.

docs/manual.md
```diff
@@ -255,6 +255,17 @@ a size isn't specified, the memory region will be sized by the length
...
+via the paramter `prefill_bootinfo` in the [System Description File](#sysdesc). 
+If a size isn't specified, the memory region defaults to a single smallest page.
+The supported bootinfo types include:
+* `x86_vbe`
+* `x86_mbmap`
+* `x86_acpi_rsdp`
+* `x86_framebuffer`
+* `x86_tsc_freq`
+* `fdt`
```
midnightveil: I think you missed this instance of fdt, btw

tool/microkit/src/sdf.rs
```diff
@@ -1540,7 +1549,30 @@ impl SysMemoryRegion {
...
+                _ => return Err(format!("BootInfoMap type: '{xml_bi_type}' is not supported.")),
+            }
+        })
+        .transpose()?;
+
+        if prefill_bytes_maybe.is_some() && prefill_bootinfo_maybe.is_some() {
+            return Err(value_error(
+                xml_sdf,
+                node,
+                "prefill_path and prefill_bootinfo cannot be both specified".to_string(),
```
midnightveil, line 1588: these two error cases should have tests in tests.rs

midnightveil:
@terryzbai the seL4 PR and rust-seL4 PRs have been merged now; so once you update rust-seL4 (and seL4) then this is good to go, I think.

docs/manual.md
```diff
@@ -255,6 +255,16 @@ a size isn't specified, the memory region will be sized by the length
 of the prefill file, rounded up to the smallest page size or the user
 specified page size.
 
+A *memory region* can also be prefilled with bootinfo by the capDL Initialiser
+via the paramter `prefill_bootinfo` in the [System Description File](#sysdesc).
```
midnightveil: parameter

docs/manual.md
```diff
@@ -255,6 +255,16 @@ a size isn't specified, the memory region will be sized by the length
...
+If a size isn't specified, the memory region defaults to a single smallest page.
+The supported bootinfo types include:
```
midnightveil, line 261: add a space after this otherwise compiled PDF looks broken

tool/microkit/src/capdl/builder.rs
```diff
@@ -536,6 +537,23 @@ pub fn build_capdl_spec(
                     }]
                     .to_vec(),
                 }
+            } else if let Some(prefill_bootinfo) = mr.prefill_bootinfo {
+                let remaining_bytes_to_fill = mr.size - starting_byte_idx;
+                let num_bytes_to_fill = min(mr.page_size_bytes(), remaining_bytes_to_fill);
+
+                FrameFill {
+                    entries: [FillEntry {
```
midnightveil: use `vec![FillEntry { }]` not `[FillEntry {}].to_vec()`

midnightveil (commented):
A few minor fixes but otherwise LGTM.

~~(Note: we still need the manifest seL4 update PR before merging)~~ Done

midnightveil (approved):

PR #538, opened 2026-06-26
Switch to two level cnodes

terryzbai:
This PR makes the CSpace of each PD as a two-level CNodes, which brings more flexibility for cspace management. This is useful for capability sharing.

Now each PD has a smaller root CNode having 64 slots in total. The original Microkit CNode is moved to the root CNode's slot 0 to make sure all existing stuff work. All shared capabilities are moved to the root CNode and addressed by shifting the slot index.

This PR also fixes the capability creation for shared VSpace and CSpace, and adds rules for quick QEMU tests in cap_sharing example.

~~This PR is also based on the branch [rust_sel4_update_to_main](https://github.com/au-ts/microkit/tree/rust_sel4_update_to_main), and should be cherry-picked when being merged.~~(already merged)

### DIFF
--- a/example/cap_sharing/Makefile
+++ b/example/cap_sharing/Makefile
@@ -55,6 +55,8 @@ LIBS := -lmicrokit -Tmicrokit.ld
 
 IMAGE_FILE = $(BUILD_DIR)/loader.img
 REPORT_FILE = $(BUILD_DIR)/report.txt
+SPEC = $(BUILD_DIR)/capdl_spec.json
+KERNEL_32B = $(BUILD_DIR)/sel4_32.elf
 
 all: $(IMAGE_FILE)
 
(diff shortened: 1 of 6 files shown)


example/cap_sharing/Makefile
```diff
@@ -67,5 +69,18 @@ $(BUILD_DIR)/%.o: %.c Makefile | $(BUILD_DIR)
...
+qemu: $(KERNEL_32B) $(IMAGE_FILE)
+	qemu-system-x86_64                      \
+			-cpu qemu64,+fsgsbase,+pdpe1gb,+xsaveopt,+xsave \
+			-m "1G"                         \
+			-display none                   \
+			-serial mon:stdio               \
+		-kernel $(KERNEL_32B) \
+		-initrd $(IMAGE_FILE)
+else
+$(error Unsupported ARCH)
```
midnightveil, line 101: This example does work on other architectures, it's just this QEMU command which doesn't. I was suggesting naming it "qemu_x86".

tool/microkit/src/capdl/builder.rs
```diff
@@ -1162,15 +1173,26 @@ pub fn build_capdl_spec(
...
                 CapMapType::Sc => capdl_util_make_sc_cap(pd_src_shadow_cspace.sched_context),
-                CapMapType::VSpace => capdl_util_make_sc_cap(pd_src_shadow_cspace.vspace),
-                CapMapType::CSpace => capdl_util_make_sc_cap(pd_src_shadow_cspace.cspace),
+                CapMapType::VSpace => capdl_util_make_page_table_cap(pd_src_shadow_cspace.vspace),
+                CapMapType::CSpace => {
+                    let Some(named_object) = spec_container.get_root_object(pd_src_shadow_cspace.cspace) else {
+                        unreachable!("internal bug: couldn't find cnode with given obj id.");
+                    };
+                    let Object::CNode(ref cnode) = named_object.object else {
+                        unreachable!("internal bug: got a non-CNode object id {} with name '{}'", usize::from(pd_src_shadow_cspace.cspace), named_object.name.as_ref().unwrap());
```
midnightveil: This should be reformatted to wrap across multiple lines

midnightveil (commented):
You need to also add the check to prevent people from placing extra caps in the Microkit CNode, yeah? (not permitted slot 0?)

example/cap_sharing/Makefile
terryzbai: uh, i was thinking to make it general, but shouldn't have added the error thing.

terryzbai:
yes,  I have added the check and a test case for it.

midnightveil (approved):

midnightveil:
LGTM once it builds.

PR #542, opened 2026-07-06
Add support for creating SDF structures directly in Microkit tool

sfulham:
Supersedes #455. Adds `new` functions for relevant SDF structures, including validation to ensure resulting System Description is valid.

### DIFF
--- a/tool/microkit/src/sdf.rs
+++ b/tool/microkit/src/sdf.rs
@@ -1638,6 +1638,68 @@ impl SysMemoryRegion {
         }
     }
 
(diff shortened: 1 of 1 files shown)


midnightveil:
Quick comment: these should not duplicate the existing checks in the code, but should replace them / be refactored.

sfulham:
> Quick comment: these should not duplicate the existing checks in the code, but should replace them / be refactored.

Yeah that's the plan, just translating the checks into a way that makes most sense for validation without the XML first

PR #544, opened 2026-07-06
IOMMU: Continue Prior Work

cazb2:
This PR makes a small change to the SDF work previously done. It might be simpler to merge that work to make reviewing new changes simpler. (Only the latest two commits have new content)

It then refactors the memory.rs module in preparation for adding support for IO Address Spaces.

This PR should preserve the current behaviour.


cazb2:
Moved to au-ts fork.

PR #546, opened 2026-07-09
ci: use microkit-manifest for PR SDK builds

midnightveil:
Now we don't have to maintain another SEL4_VERSION variable, only need to use 'Test with' in a PR to
adjust the seL4 build used, and updates to seL4 go through the manifest repository.


lsf37 (approved):
I think this is good, at least what I can see from the workflow side of things.

PR #547, opened 2026-07-09
seL4Config: remove duplicated constants

cazb2:
This commit removes some of the duplicated constants definitions that seL4 already exports.

### DIFF
--- a/build_sdk.py
+++ b/build_sdk.py
@@ -726,32 +726,35 @@ def build_sel4(
 
     # Use the preprocessor to convert the seL4 object size constants to readable JSON
     # for the tool.
-    object_sizes_header = tool_dir / "object_sizes.h"
-    dest = sdk_dir / "board" / board.name / config.name / "object_sizes.json"
-    preprocessor = "clang" if (llvm) else f"{target_triple}-cpp"
-    preprocess_cmd = [
-        preprocessor,
-        "-E",
-        "-P",
-        f"-I{include_dir}",
-        object_sizes_header,
-    ]
-    r = subprocess.run(preprocess_cmd, capture_output=True)
-    if r.returncode != 0:
-        raise Exception(f"Failed creating object_sizes.json: cmd={preprocess_cmd}")
+    sel4_constants = ["object_sizes", "address_space_constants"]
+    for sel4_constant_type in sel4_constants:
+        header_src = tool_dir / f"{sel4_constant_type}.h"
+        json_dst = sdk_dir / "board" / board.name / config.name / f"{sel4_constant_type}.json"
+        preprocessor = "clang" if (llvm) else f"{target_triple}-cpp"
+        preprocess_cmd = [
+            preprocessor,
+            "-E",
+            "-P",
+            f"-I{include_dir}",
+            header_src,
+        ]
+        r = subprocess.run(preprocess_cmd, capture_output=True)
+        if r.returncode != 0:
+            raise Exception(f"Failed creating object_sizes.json: cmd={preprocess_cmd}")
+        sizes = []
+        for l in r.stdout.decode("utf-8").split("\n"):
+            # Preprocessor emits commented lines etc that we want to ignore
+            if ": " in l:
+                assert len(l.split(": ")) == 2
+                obj_name, size = l.split(": ")
+                if "-" in size:
+                    lhs, rhs = size.strip("( )").split("-", 1)
+                    size = str(int(lhs.strip(), 0) - int(rhs.strip(), 0))
+                sizes.append((obj_name, int(size, 0)))
 
-    preprocessor_out = r.stdout.decode("utf-8")
-    object_sizes = []
-    for l in preprocessor_out.split("\n"):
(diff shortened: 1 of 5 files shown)


cazb2:
style check warning can be ignored or we fix the python/style but I was told not too

build_sdk.py
```diff
@@ -726,32 +726,35 @@ def build_sel4(
...
+            raise Exception(f"Failed creating object_sizes.json: cmd={preprocess_cmd}")
+        sizes = []
+        for l in r.stdout.decode("utf-8").split("\n"):
+            # Preprocessor emits commented lines etc that we want to ignore
+            if ": " in l:
+                assert len(l.split(": ")) == 2
+                obj_name, size = l.split(": ")
+                if "-" in size:
+                    lhs, rhs = size.strip("( )").split("-", 1)
+                    size = str(int(lhs.strip(), 0) - int(rhs.strip(), 0))
```
midnightveil, line 752: Hrm, having to do math here is slightly weird. I suppose there's no way around it.

tool/microkit/src/main.rs
```diff
@@ -264,7 +274,8 @@ fn main() -> Result<(), String> {
         invocations_labels,
         device_regions,
         normal_regions,
-        object_sizes,
+        object_sizes: Some(object_sizes),
```
midnightveil: Ok, additional cleanup while we're here: can you make this not optional? It's a little silly

cazb2: The only benefit of it being an option is testing

midnightveil: I don't find that very convincing :sweat_smile:

build_sdk.py
cazb2: Unless we extract from part of the compiled tu

tool/microkit/src/main.rs
cazb2: Understandably

midnightveil:
Your branch is out of date, btw. You need to rebase.

tool/microkit/address_space_constants.h
```diff
@@ -0,0 +1,24 @@
...
+ */
+
+#include <sel4/arch/constants.h>
+#include <sel4/constants.h>
+#include <sel4/plat/api/constants.h>
+#include <sel4/sel4_arch/constants.h>
+
+page_table_index_bits: seL4_PageTableIndexBits
+
+#ifdef seL4_UserVSpaceTop
```
midnightveil: I believe this `#ifdef` should be unconditional.

tool/microkit/src/main.rs
```diff
@@ -197,12 +197,22 @@ fn main() -> Result<(), String> {
         }
     };
 
-    let object_sizes = {
```
midnightveil, line 200: Is there a reason why you removed the existing

```
let object_sizes = {
  // do stuff to parse
};
```

pattern here? I can't see what was wrong with it

It's much nicer IMO.

cazb2: Hmm interesting I don't see the point in the scope, maybe a function would make more sense?

midnightveil: The scope makes a new namespace, so you can define several variables without populating global scope.

Also, it logically groups things.

It's sort of like a function in that way yes.

BTW, clipppy failed, and you need to rebase.

cazb2: on it

cazb2: When I look this commit is alr rebased?

midnightveil: Not according to GitHub. I suspect you have a local copy of main either following au-ts (not seL4) or just hasn't been git pulled. (note; git pull only updates the current branch and your knowledge of remotes. it does not update other branches such as main to follow remotes).

cazb2: hmm I pressed the automatic rebase button on GitHub and it created an empty commit I don't get what the problem was?

midnightveil: It's not empty. Also you did a merge not a rebase (not that it matters, as when I rebase-merge the PR into main it goes away)

midnightveil: The parent of 711c607 was 'tests: add test cases for bootinfo prefilling', which is several commits behind latest main.

midnightveil: Yeah, you likely only have the au-ts remote, as tbe parent of that commit is the current au-ts microkit fork. You need to have an additional seL4 remote following seL4/microkit that main tracks: because we work on a fork.

cazb2: mmmm so a fork can have a different origin main state?

cazb2: I think I understand this a bit now, kinda annoying behaviour

midnightveil: Yes, they're not linked in any way.

midnightveil (approved):

PR #552, opened 2026-07-13
tool: a VM on x86-64 cannot have scheduling params

### LINKED ISSUE #540: x86_64: a VM should not take any scheduling parameters in the SDF
This is because the VMM and VCPU shares a TCB and Scheduling Context. Whereas on ARM they are separate entities, so the SDF parsing code and manual need to be updated to not confuse users.

dreamliner787-9:
Closes https://github.com/seL4/microkit/issues/540.

### DIFF
--- a/tool/microkit/src/sdf.rs
+++ b/tool/microkit/src/sdf.rs
@@ -343,7 +343,7 @@ pub struct VirtualMachine {
     pub vcpus: Vec<VirtualCpu>,
     pub name: String,
     pub maps: Vec<SysMap>,
-    pub sched_params: SchedulingParams,
+    pub sched_params: Option<SchedulingParams>,
 }
 
 #[derive(Debug, PartialEq, Eq)]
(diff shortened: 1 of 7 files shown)


tool/microkit/src/sdf.rs
```diff
@@ -265,14 +265,19 @@ impl Display for CpuCore {
...
+
 #[derive(Debug, PartialEq, Eq)]
 pub struct ProtectionDomain {
     /// Only populated for child protection domains
     pub id: Option<u64>,
     pub name: String,
-    pub priority: u8,
-    pub budget: u64,
-    pub period: u64,
+    pub sched_params: SchedulingParams,
```
midnightveil, line 280: This change to PD isn't actually necessary, right? This is what creates a bunch of changes everywhere else.

dreamliner787-9: That's right, I made that change because in struct VirtualMachine I thought making the priority, budget and period Optional is a bit ugly. So I applied the refactor to ProtectionDomain too.

midnightveil (approved):

PR #558, opened 2026-07-15
Refactor Memory.rs

cazb2:
This PR makes a small change to the SDF work previously completed.

It then refactors the memory.rs module in preparation for adding support for IO Address Spaces. The refactor primarily introduces an address space abstraction that can allow simpler handling of different address space variants.

It also removes the pre-existing duplicated constants that are already defined by seL4. This commit extracts those and includes them in the Config struct that the tool passes around.

This PR preserves the existing behaviour.


midnightveil (approved):
Approved. Was reviewed in https://github.com/au-ts/microkit/pull/7 for easiness.

PR #564, opened 2026-07-17
tool: implement new Node tree pattern for SDF structures

sfulham:
As discussed privately, this is part of some work to make
the microkit tool usable as a software library. In order to
not duplicate effort and consider two distinct interfaces,
we instead make the software library create a node-tree
structure that represents the XML. The library gets a
slightly less "nice" API, but we can use most of the same
code without needing large amounts of structural refactoring
of the Microkit tool: and more importantly, the same
backwards compatibility concerns apply for the SDF XML file
as they do to use of Microkit as a library.

### DIFF
--- a/tool/microkit/src/sdf.rs
+++ b/tool/microkit/src/sdf.rs
@@ -30,6 +30,7 @@ use std::fmt;
 use std::fs;
 use std::num::NonZero;
 use std::ops::Deref;
+use std::ops::Range;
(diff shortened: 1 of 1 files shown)


midnightveil:
Just as an FYI, I'll look into this further after the next Microkit release (2.3.0) on Tuesday.

midnightveil (approved):

dreamliner787-9 (approved):

PR #565, opened 2026-07-17
Add release notes (CHANGES.md) for Microkit 2.3.0

midnightveil:
As per title.

Before merging:

- [x] Need to actually merge the domain scheduler.
- [x] Update seL4 version (what is the new seL4 version number?)
- [x] https://github.com/seL4/microkit/blob/main/platforms.yml
- [x] rerun sDDF CI
- [x] rerun libVMM CI
- [x] versions
- [x] bump manifest seL4


CHANGES.md
```diff
@@ -1,5 +1,123 @@
...
+
+Major:
+
+* x86_64 systems are now built with the IOMMU turned on by default, as we have
+  added new syntax to allow for the IOMMU to be configured (see below).
+  However, previously the IOMMU was disabled in seL4, so systems which did DMA
+  on x86_64 will now *stop working* without specifying the IOMMU setup.
+
+These are all breaking changes, but affect only configurations which were not
+supported by the Microkit, or didn't work previously.
```
Indanz: It's unclear to what changes this sentence refers. Bugfixes and changes to things that were previously unsupported shouldn't be listed under "breaking changes", it would be good to separate them out so the real breaking changes are easier to see. I'd put them with non-feature changes in a subsection "Other changes" or something.

CHANGES.md
```diff
@@ -1,5 +1,123 @@
...
+* Support for the seL4 domain schedule now that the [RFC-20 Runtime Schedules](
+  https://sel4.github.io/rfcs/implemented/0200-domain-schedules.html) has been
+  added to seL4. This is useful for systems which want isolated subsystems
+  with limited information flow between them.
+
+### Bug fixes
+
+* TQMA8XQP1GB loader address has been fixed as it was collided with reserved
+  regions in memory.
+* IOAPIC IRQ polarity was inverted.
```
Indanz, line 99: If the user can specify this, then this is also a breaking change.

CHANGES.md
```diff
@@ -1,5 +1,123 @@
...
+
+### Notes for developers
+
+* There is a new `--override-kernel` option to the tool that allows one to
+  use an SDK built of Microkit but with a different seL4 kernel ELF.
+* We now have basic CI that runs Microkit images on real hardware.
+* We now recommend the use of the [Google repo-manifest 'microkit-manifest'](
+  https://github.com/seL4/microkit-manifest) which tracks the supported version
+  of seL4 to be used for testing the Microkit. It also records the latest build
+  of the main branch which passed CI.
```
Indanz, line 132: This was probably easier for CI integration I assume? Otherwise git submodules would have been better.

CHANGES.md
midnightveil: I don't quite know how to deal with that. Is it a breaking change if something which previously didn't work (or work right) is now disallowed? (etc). It's both a breaking change (something to be aware of) but also not.

CHANGES.md
midnightveil: Yeah, by my definition it should be.

(But yes, "high" became "low" and vice versa...)

CHANGES.md
midnightveil: Yes, better for CI, since that's what everything uses. Also, previous discussions about this indicated that "the" solution for seL4 org repos was google repo.

Also, I'm not convinced submodules are better. They make rebasing so intensely difficult I'd rather not use them at all.

CHANGES.md
Indanz: If something was broken already, a change to that doesn't break it. Breaking change only applies to things that used to work and were supported. In the case of Microkit it's a bit more complicated because it has the system description file. If a change makes it error out instead of silently ignoring the wrong stuff, then technically that's a breaking change, because user stuff that used to work now doesn't. But I probably would still mention it under "other changes" section myself anyway.

The most important thing is that behavioural changes that don't cause an easy to fix error are documented.

CHANGES.md
Indanz: That's exactly the thing you want documented under "breaking changes". Because everyone who worked around the bug by giving the opposite value, needs to know that they can undo that work-around.

CHANGES.md
Indanz: I wouldn't recommend using git submodules for normal repos (that is, mixing normal code with submodules). But if all you want to do is bundling a bunch of external repos in a certain directory layout together, then it seems absolutely fine to me. I don't see the difference with Google's repo then, except you don't need another tool that always wants to update it self at every run. Both just document which commit to use for each repo.

CHANGES.md
joep: Absolutely this is a breaking change and should be included as such. It is going to break our code so....

CHANGES.md
```diff
@@ -1,5 +1,123 @@
...
+support for two new boards.
+
+### Breaking changes
+
+Major:
+
+* x86_64 systems are now built with the IOMMU turned on by default, as we have
+  added new syntax to allow for the IOMMU to be configured (see below).
+  However, previously the IOMMU was disabled in seL4, so systems which did DMA
+  on x86_64 will now *stop working* without specifying the IOMMU setup.
```
dreamliner787-9, line 23: I think it is worth saying that only 4KiB page is supported for IO maps.

dreamliner787-9 (dismissed):

CHANGES.md
midnightveil: Does the page size really matter? I would presume you mostly wouldn't specify one.

CHANGES.md
midnightveil: Included.

CHANGES.md
dreamliner787-9: It's worth specifying IMO, because in the manual we specify something to the effect of "memory region will default to the largest possible page size". But this isn't the case for memory regions with IO maps.

~~Hmmm, as I write this comment I realised that the Microkit tool does not safe guard against this. It will still optimise the page size if it have io maps. Will fix~~ nevermind the tool does check for it I read the code wrong

midnightveil: OK

dreamliner787-9 (approved):

midnightveil:
Merge this after seL4 is released and we update the seL4 in the manifest.

CHANGES.md
```diff
@@ -1,5 +1,136 @@
...
+### Board support
+
+* 'stm32mp2' for the STM32MP2 evaluation board.
+* 'qemu_virt_aarch64_el1' for QEMU/KVM on Linux AArch64 hosts (which don't support
+   nested virtualisation).
+
+### Notes for developers
+
+* There is a new `--override-kernel` option to the tool that allows one to
+  use an SDK built of Microkit but with a different seL4 kernel ELF.
```
lsf37: ?
```suggestion
  use an SDK build of Microkit but with a different seL4 kernel ELF.
```

lsf37 (approved):
Looks good from my side!

PR #569, opened 2026-07-20
manual: document must have CPU features for x86

### LINKED ISSUE #507: Document x86 assumed features
In the manual we should document what optional x86-64 CPU feature bits we assume are available. One concrete example right now is huge pages.

This would involve looking at where seL4 calls `cpuid` and checks for specific bits for our configuration of seL4 that we distribute and documenting them in the manual so users know what to expect.

dreamliner787-9:
This is useful for users of Microkit to check that their environment support all the required features by the kernel. Rather than getting a nasty error at boot time.

Closes https://github.com/seL4/microkit/issues/507.

One thing I'm a bit sketchy on is that the VT-x features table is not rendered very nicely in the PDF:
image

One thing I could do is that I could refactor the table to be grouped by MSR, but I don't know which is better,

### DIFF
--- a/docs/manual.md
+++ b/docs/manual.md
@@ -1729,6 +1729,21 @@ qemu-system-x86_64: Cannot load x86-64 image, give a 32bit one.
 To obtain a bootable ISO image, a Multiboot 2-compliant bootloader must be used. Please refer to your
 bootloader's documentations.
 
+### Required CPU features
+
+The `x86_64_generic` platform's kernel configuration requires these features from your CPU:
+
+| Name                       | cpuinfo feature flag |
+|----------------------------|----------------------|
+| Page Table Attribute (PAT) | pat                  |
+| XSAVE                      | xsave                |
+| x87 state                  | fpu                  |
+| SSE state                  | sse                  |
+| Huge Page                  | pdpe1gb              |
+| fsgsbase                   | fsgsbase             |
+
+To check if your CPU supports a specific feature, boot into Linux and run a command like `grep pdpe1gb /proc/cpuinfo`.
+
 ## x86-64 generic (with VT-x) {#x86_64_generic_vtx}
 
 This board supports x86-64 platforms with generic microarchitecture and virtualisation.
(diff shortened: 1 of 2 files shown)


midnightveil:
I don't think it's necessary to give either the kernel reference nor the exact way to check this, just the list of necessary features. To make it easier you could probably specify the specific microarch (e.g. nehalam or zen2, I made those up) for is the minimum for which it works.

Indanz:
The most user friendly way of listing the required features would be to list the matching /proc/cpuinfo flags.

dreamliner787-9:
> The most user friendly way of listing the required features would be to list the matching /proc/cpuinfo flags.

I thought about this but most of the VT-x features doesn't correspond to a `proc/cpuinfo` flags.

docs/manual.md
```diff
@@ -1644,6 +1644,19 @@ qemu-system-x86_64: Cannot load x86-64 image, give a 32bit one.
 To obtain a bootable ISO image, a Multiboot 2-compliant bootloader must be used. Please refer to your
 bootloader's documentations.
 
+### Required CPU features
+
+The `x86_64_generic` platform's kernel configuration requires these features from your CPU:
+
+| Name                       | /proc/cpuid flag |
```
midnightveil: `/proc/cpuinfo`? I'd also mention in the text (not the table) that you can check with e.g. `cat /proc/cpuinfo | grep xsave`; and just make the table header the "cpuinfo feature flag" or something

docs/manual.md
```diff
@@ -1665,6 +1678,40 @@ vt-x: not supported
...
+
+| Name                             | /proc/cpuid flag |
+|----------------------------------|------------------|
+| Virtual Machine Extensions (VMX) | vmx              |
+
+In addition to the CPU feature flags above, the kernel also requires these Intel VT-x features from your CPU:
+
+| Name                             | /proc/cpuid flag  |
+|----------------------------------|-------------------|
+| Enable VMX outside SMX operation | (Checked via MSR) |
```
midnightveil: I would at least keep the name of the MSR or a reference to the Intel manual here so that people can check this, even if manually.

docs/manual.md
dreamliner787-9: > /proc/cpuinfo?

Oops

midnightveil (approved):

PR #570, opened 2026-07-20
IOMMU: Style Fix ups (minor)

cazb2:
Improve variable naming such that it is consistent between the sdf api and the code.
Remove some boilerplate that can be avoided for now (Hash impl on the PCIDevice).
This implied changing hash map/hash sets that were keyed on this type being replaced, which was a trivial fix, since the name field which we validate as unique in the sdf parsing can be used instead.

Swap order that I placed the IOMMU address space construction in the builder.rs, previously I placed it after handling extra caps,  but in future it is likely the cap sharing will including IO related caps. The diff here is really just the change to the comment ("step 5/6"), but GitHub doesn't handle reorderings well.

### DIFF
--- a/tool/microkit/src/capdl/builder.rs
+++ b/tool/microkit/src/capdl/builder.rs
@@ -24,8 +24,8 @@ use crate::{
     },
     elf::ElfFile,
     sdf::{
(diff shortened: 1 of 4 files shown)


dreamliner787-9:
Could u please rebase your branch? The CI should pass now.

midnightveil (approved):

PR #573, opened 2026-07-22
feat: add pi 5 support

iiPythonx:
This PR essentially takes [the rpi5 branch](https://github.com/seL4/microkit/tree/rpi5) and migrates UART related configuration to it's correct location, while also adding documentation and rebasing.

It's been tested to work with 2GB, 4GB, and also 16GB (with the latter two requiring a custom overlay and board declaration).  
If anything here is incorrect, please let me know.


midnightveil (approved):
This looks good.

Can you put up a PR to [ci-actions](https://github.com/seL4/ci-actions) that adds the `microkit_boards` and `microkit_no_hw_test` declarations to the `platforms.yml` in that repo? Otherwise our CI will complain about not knowing the board.

midnightveil:
I'm not sure what that style checker is suddenly complaining. I'll look into it.

Also, please build & test smp mode.

iiPythonx:
https://github.com/seL4/ci-actions/pull/519

I've also confirmed that SMP functions on the Pi 5 assuming you update the CPU IDs as I just did in https://github.com/seL4/microkit/pull/573/commits/a7d2b882b0937502cc92a0bf216eb3a7d4959cae.

loader/src/aarch64/cpus.c
```diff
@@ -73,6 +73,12 @@ static const size_t psci_target_cpus[4] = {0x00, 0x01, 0x02, 0x03};
 /* BCM2711 does not use PSCI for CPU bring-up. */
 #define SMP_USE_SPIN_TABLE
 static const size_t cpus_release_addr[4] = {0xd8, 0xe0, 0xe8, 0xf0};
+#elif defined(CONFIG_PLAT_BCM2712)
+/* Unlike BCM2711, BCM2712 boots through ARM Trusted Firmware (TF-A BL31) and
+ * supports PSCI CPU_ON. However, because its Cortex-A76 cores use DynamIQ (DSU),
+ * MPIDR target CPU IDs are shifted to Affinity level 1 (0x100 step) rather
+ * than Affinity level 0 like older single-cluster Cortex-A53/A72 boards. */
```
midnightveil: (this might come off as harsh, sorry, not meaning it). These comments I think aren't very necessary. Yes, some boards use Aff1 instead of Aff0 because of the way the clusters are set up. However, these matches the reg values in the Rpi's DTS, so everything is fine.; However, caveat:

Presumably, the MPIDR goes 0x0, 0x100, etc? Or is this incorrect. (If it goes 0x0,0x1, etc then maybe this comment is OK. But it's also not necessary as the code already knows about the distinction between PSCI target CPUs and MPIDRs because the ODROID C4 completely violates the specifications (that target CPU = MPIDR)

iiPythonx: That's fair enough, I'll remove it real quick.

PR #576, opened 2026-07-23
CapMap: Remove the root cnode upper bound

cazb2:
This commit determines the size of the root cnode based upon the largest slot number that is specified in the sdf. It then uses the appropriately sized cnode object in the spec.

### DIFF
--- a/tool/microkit/src/util.rs
+++ b/tool/microkit/src/util.rs
@@ -73,6 +73,16 @@ pub fn ranges_overlap<T: PartialOrd>(left: &Range<T>, right: &Range<T>) -> bool 
     !(left.end <= right.start || right.end <= left.start)
 }
 
+/// Returns the number of bits required to repr the input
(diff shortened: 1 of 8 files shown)


tool/microkit/src/util.rs
```diff
@@ -73,6 +73,16 @@ pub fn ranges_overlap<T: PartialOrd>(left: &Range<T>, right: &Range<T>) -> bool
     !(left.end <= right.start || right.end <= left.start)
 }
 
+/// Returns the number of bits required to repr the input
+pub fn calculate_size_bits<T: Into<u64>>(size: T) -> u8 {
```
midnightveil, line 77: ilog2.

midnightveil:
I want to think about the mechanics of this. The original plan was to expand the sub-cnode, I'm not sure about the root right now.

cazb2:
Does that imply a refactor such that any user specified caps won't live in the root cnode but a subcnode?

PR #577, opened 2026-07-23
MapPerms: Consistency between map and iomap

cazb2:
This commit unifies the way VM rights are encoded in the sdf. Previously normal map perms were represented as a raw u8 enum while the iomap perms were represented as a proper type. 

This uses a base enum for the vm rights and then uses a context specific struct around the base rights depending on if the perms are for a iomap or map entry.

This preserves existing correct behaviour, (I believe seL4 would reject the WriteExecute case where the tool would previously accept this for normal frames leading to a runtime error.)

### DIFF
--- a/tool/microkit/src/sdf/memory_region.rs
+++ b/tool/microkit/src/sdf/memory_region.rs
@@ -23,26 +23,87 @@ use crate::util::get_full_path;
 use crate::util::round_up;
 use crate::{Config, PageSize};
 
-#[repr(u8)]
-pub enum SysMapPerms {
-    Read = 1,
-    Write = 2,
-    Execute = 4,
+// We do not include attributes (execute/cached) here since seL4 does not treat them like vm_rights.
+// Note on seL4 you can't mint a write only frame. However you can request a write only mapping,
+// for the IOMMU on x86.
+// https://github.com/seL4/seL4/blob/5126e718bb68636267e28234c711ac8022b79fe2/src/arch/x86/kernel/vspace.c#L673-L692
+#[derive(Debug, PartialEq, Eq, Clone, Copy)]
+pub enum FrameRights {
+    Read,
+    Write,
+    ReadWrite,
+    None,
+}
+
+impl FrameRights {
+    fn from_bools(read: bool, write: bool) -> Self {
+        match (read, write) {
+            (true, false) => Self::Read,
+            (false, true) => Self::Write,
+            (true, true) => Self::ReadWrite,
+            (false, false) => Self::None,
+        }
+    }
+
+    pub fn read(self) -> bool {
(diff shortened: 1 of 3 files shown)


cazb2:
https://github.com/seL4/seL4/blob/5126e718bb68636267e28234c711ac8022b79fe2/src/arch/x86/kernel/vspace.c#L673-L692

seL4 reference for write only run time error.

midnightveil:
Apologies, you will need to rebase affter: https://github.com/seL4/microkit/pull/581

tool/microkit/src/sdf/memory_region.rs
```diff
@@ -94,15 +140,15 @@ impl Map for SysMap {
...
 
     fn write(&self) -> bool {
-        self.perms & SysMapPerms::Write as u8 != 0
+        self.perms.write()
     }
 
     fn execute(&self) -> bool {
-        self.perms & SysMapPerms::Execute as u8 != 0
+        self.perms.execute()
     }
```
midnightveil: Instead: `map.perms().read`? Since type of `perms` is the same in both cases? Then don't need these extra methods? Or does that get ugly?

tool/microkit/src/sdf/memory_region.rs
```diff
@@ -214,31 +260,34 @@ impl SysMemoryRegion {
...
-            (false, true) => Ok(SysIOMapPerms::Write),
-            (false, false) => Err(()),
-        }
+        let frame_rights = match FrameRights::from_bools(read, write) {
+            FrameRights::None => return Err("Invalid frame right for IOMap".into()),
+            frame_rights => frame_rights,
+        };
+        Ok(SysIOMapPerms(frame_rights))
+    }
+    pub fn read(self) -> bool {
```
midnightveil: Spacing between functions

dreamliner787-9 (approved):
Looks good from me apart from Julia's comments

dreamliner787-9:
> https://github.com/seL4/seL4/blob/5126e718bb68636267e28234c711ac8022b79fe2/src/arch/x86/kernel/vspace.c#L673-L692
> 
> seL4 reference for write only run time error.

Actually can you add this reference into https://github.com/seL4/microkit/pull/577/changes#diff-5fc8f1c29b8f90ae6c6d27410e489ac4d389486ec4c31f3323604cbab3f1cf1fR334

tool/microkit/src/sdf/memory_region.rs
cazb2: I think I made it work lmk if you're happy with the design, it scales better while still allowing the map and the iomap to define their own types around the common FrameRights.

midnightveil (approved):

PR #581, opened 2026-07-27
tool(sdf): split into files

midnightveil:
Cleanup.


dreamliner787-9 (approved):

Indanz:
I was about to suggest to split the big files up to reduce the number of merge conflicts and to make things more manageable.

midnightveil:
Yes, for some reason the Microkit tool has tended to generate a lot of large files. Probably because everything started without structure and there was no definite tipping point to adding more being too much.

Indanz:
It's something you have to be aware of and check every time you add code to an existing function or file: Does it need splitting up?

PR #583, opened 2026-07-29
[draft] libmicrokit: API for 64-bit number debug print

ZGwtao:
This commit once added microkit_dbg_put64 that helps to print out large 64-bit integer number. However, it seems functions like `puthex64` are more preferred and duplicated everywhere in the repo:

- `example/ethernet/eth.c`
- `example/ethernet/pass.c`
- `example/timer/timer.c`
- `loader/src/uart.c`
- `monitor/src/util.c`

Would it be a better way to provide a `microkit_dbg_puthex64` than duplicating all the above implementations?


ZGwtao:
Having discussed with Julia and it seems not putting everything into libmicrokit is a better approach. So, I will close this PR.

PR #584, opened 2026-07-30
tool: refactor `main` to move out kernel configuration and binary generation

sfulham:
Currently kernel configuration and image generation are embedded in the `main` function of Microkit Tool. This means that external applications that wish to use Microkit in their toolchain can't use the tool in their toolchain as a library. This refactors the `main` functions so that these functions can be used by external programs.


midnightveil:
@sfulham You'll need to fix gitlint. Also, could you squash commit 3 into the previous commit that createa that file?

midnightveil:
So gitlint still complaining. Please see: https://sel4.systems/Contribute/git-conventions.html. Ther'es a length limit.

I would prefer it you keep commit 1 and 2 separated though, the kernel config & image generation are better as separate commits

sfulham:
> So gitlint still complaining. Please see: https://sel4.systems/Contribute/git-conventions.html. Ther'es a length limit.
> 
> I would prefer it you keep commit 1 and 2 separated though, the kernel config & image generation are better as separate commits

ah thank you, didn't realise there was a page for the convention!

midnightveil:
You can also run the `gitlint` tool locally yourself. The config is [this file](https://github.com/seL4/seL4_tools/blob/master/misc/.gitlint).

midnightveil (approved):

PR #588, opened 2026-08-06
tool: separate tree parsing and xml parsing

sfulham:
Moves XML SDF specific parsing logic into `parse_xml` and node tree parsing into `parse`.


tool/microkit/src/sdf.rs
```diff
@@ -170,6 +163,25 @@ pub fn parse(
 
     let system: &dyn SdfNode = &system;
 
+    parse(xml_sdf, system, config, search_paths)
+}
+
+pub fn parse(
+    xml_sdf: SystemDescriptionFile,
```
midnightveil: hm, I think I'd rather than this takes a &Path and constructs XmlSystemDescription internally. not much point exposing that in an external interface if it's pointless.

then can leave off the pub(crate) change.

Speaking of I should probably go through and audit what needs to be public vs not..

sfulham: My main thinking with this was that `parse_xml` uses it for `check_no_text` (which to be fair may not *actually* need it) and didn't really want to have to construct it twice if we don't have to, but ultimately it doesn't really matter either way.

midnightveil: Nah, should be fine to construct again.

It's a struct with one member. The in-memory-layout will (almost certainly) be the same as the member itself.

midnightveil (approved):

PR #589, opened 2026-08-10
Support badged SMC caps to restrict SMC function IDs

### LINKED ISSUE #290: Add ARM SMC forwarding granularity
Turns out the kernel has the ability to restrict the SMCs possible depending on the badge of the SMC cap.

We should add the ability for PDs to make use of that.

midnightveil:
Please review commit by commit.

## Summary

Add the ability to provide caps for performing SMC invocations which are 'badged',
only allowing certain function invocations. This allows for exposing certain
functions to specific PDs without granting unilateral rights to do anything, i.e.
following a principle of least privilege. This mechanism [was explicitly
discussed in RFC 9: New Capability for seL4 SMC Forwarding](
https://sel4.github.io/rfcs/implemented/0090-smc-cap.html#for-users).

## Motivation

The existing Microkit SDF syntax allows for protection domains to be given
SMC capabilities. However, this is an unbadged SMC capability, which allows
access to all SMC functions, i.e. a lot of power over the system.

```xml
<protection_domain name="arm_smc" smc="true">
    <!-- elided -->
</protection_domain>
```

PDs are then allowed to invoke the `microkit_arm_smc_call(...)` function with
the `seL4_ARM_SMCContext *args` and `seL4_ARM_SMCContext *resp` structs. This
performs an invocation on `seL4_ARM_SMC_Call(ARM_SMC_CAP, args, resp)`.

Even if all a PD wants to do is call the `PSCI_VERSION` (function `0x84000000`),
they have all the rights to reboot the system, on/offline CPUs, etc.

seL4 provides a simple mechanism to restrict `SMC` capabilities; by providing
a *badged* SMC capability, only function IDs that match the badge are allowed.
For allowing multiple SMC function IDs, one needs multiple badged copies of the
SMC capability.

## Proposal

Using the existing `<cspace>` cap-mapping feature, we add a new node type to the
SDF file, the `cap_smc`.

```xml
<protection_domain name="arm_smc">
    <!-- elided -->

    <cspace>
        <!-- An (ARM) SMC capability that permits SMC calls to 0x84000000 (PSCI_VERSION) -->
        <cap_smc slot="1" function_id="0x84000000" />
    </cspace>
</protection_domain>
```

User code then looks like this:

```c
#define CAP_SMC_PSCI_VERSION  (microkit_cspace_root_slot_to_cptr(1))

seL4_ARM_SMCContext args = { .x0 = PSCI_VERSION_FID, 0 };
seL4_ARM_SMCContext resp;

seL4_Error err;
err = seL4_ARM_SMC_Call(CAP_SMC_PSCI_VERSION, &args, &resp);
if (err != seL4_NoError) {
    /* error - e.g. if invoked with wrong function ID, etc. */
}
```

Specifying a `function_id="0"` is an escape hatch for development, or when you
don't care about security, which provides an SMC capability that is unbadged
and so allows any SMC function call.

As part of this proposal, we remove the `microkit_arm_smc_call()` and `smc="true"`
attribute from the Microkit API. This is because the next release, Microkit 3.0.0,
is planned to be a [breaking change](https://github.com/seL4/rfcs/pull/41). This
is the last commit of this PR and could be dropped instead, leaving in a deprecated
API.

## Rationale and alternatives

We briefly discuss the choices in the design space made in this proposal.

First, we continue allowing unbadged SMC capabilities to be provided, i.e.
`function_id="0"`. We could instead disallow this entirely and have part of this
include policy to never allow this to be the case. We reject this as it can be
useful for development to allow any function ID.

The second point is the decision to expose the seL4 API and capabilities directly,
for calling `seL4_ARM_SMC_Call`. This makes it the responsibility of the user
to manage calling multiple SMC call function IDs themselves. This is instead how
CAmkES works: see [`camkes_get_smc_cap`](
https://github.com/seL4/camkes-tool/blob/camkes-3.13.0/camkes/templates/component.common.c#L87-L104)
and [`vm_smc_handler` of CAmkES-VM](
https://github.com/seL4/camkes-vm/blob/camkes-3.13.x-compatible/components/VM_Arm/src/main.c#L1149-L1161).

```c
seL4_Word smc_function_id = args->r0;
seL4_ARM_SMC smc_cap = camkes_get_smc_cap(smc_function_id);
seL4_ARM_SMC_Call(smc_cap, &args, &resp);
```

We could do this in Microkit, too. This requires providing information to the PD
about which SMC capabilities are where, and then providing a function to distinguish
based on function ID. We choose not to do this, generally we only pass information
to PDs when the user requests, e.g. `setvar`, and it is up to the user to choose
how to encode data about the system in their program; Microkit just provides the
mechanisms.

## Original README

<details>
<summary>Original README</summary>

Note: this contains a breaking change by removing the SMC attribute.
This should be fine, as the 3.0.0 release is planned to be breaking anyway.
This change could be reverted; it is left as a separate commit to make that easy.

The other design choice here was that we do not permit "function_id=0", or some other way to make `badge = 0 <=> unbadged smc cap` which has no allowlisting, instead allowing everything. This is always (in my opinion) a bad design.

```xml
<system>
    <protection_domain name="arm_smc" priority="1">
        <program_image path="arm_smc.elf" />

        <cspace>
            <!-- An (ARM) SMC capability that permits SMC calls to 0x84000000 (PSCI_VERSION) -->
            <cap_smc slot="1" function_id="0x84000000" />
        </cspace>
    </protection_domain>
</system>
```

</details>

### DIFF
--- a/tool/microkit/src/sdf/cspace.rs
+++ b/tool/microkit/src/sdf/cspace.rs
@@ -20,9 +20,33 @@ pub enum CapMapType {
 }
 
 #[derive(Debug, PartialEq, Eq)]
+pub enum CapMapRefDataPdKind {
+    Tcb,
+    Sc,
+    VSpace,
+}
+
+impl From<CapMapType> for CapMapRefDataPdKind {
+    fn from(value: CapMapType) -> Self {
+        match value {
+            CapMapType::Tcb => CapMapRefDataPdKind::Tcb,
+            CapMapType::Sc => CapMapRefDataPdKind::Sc,
+            CapMapType::VSpace => CapMapRefDataPdKind::VSpace,
+        }
+    }
+}
+
+#[derive(Debug, PartialEq, Eq)]
+pub enum CapMapRefData {
+    Pd {
+        pd: Rc<str>,
+        kind: CapMapRefDataPdKind,
+    },
+}
+
+#[derive(Debug, PartialEq, Eq)]
 pub struct CapMap {
-    pub cap_type: CapMapType,
-    pub pd: Rc<str>,
+    pub ref_data: CapMapRefData,
     // The destination "slot" in the CSpace: note that this is "opaque" and
     // can be shifted depending on the location in the CSpace to work as the CPtr,
     // but here it is given as the index into the CNode.
@@ -45,9 +69,19 @@ impl CapMap {
         // At the moment the four cap maps we support all have the 'pd' element,
         // so we can include it here. When that stops being the case we will
         // have to rework this a bit.
-        check_attributes(xml_sdf, node, &["slot", "pd"])?;
 
-        let pd = Rc::from(checked_lookup(xml_sdf, node, "pd")?);
+        let ref_data = match cap_type {
+            CapMapType::Tcb | CapMapType::Sc | CapMapType::VSpace => {
+                check_attributes(xml_sdf, node, &["slot", "pd"])?;
+
+                let pd = Rc::from(checked_lookup(xml_sdf, node, "pd")?);
+
+                CapMapRefData::Pd {
+                    pd,
+                    kind: cap_type.into(),
+                }
+            }
+        };
 
         let slot: u64 = sdf_parse_required_attribute(xml_sdf, node, "slot")?;
 
@@ -69,11 +103,20 @@ impl CapMap {
         }
 
         Ok(CapMap {
-            cap_type,
-            pd,
+            ref_data,
             slot,
             text_pos: node.range().start,
         })
+    }
+
+    pub(crate) fn format_for_slot_collision(&self, xml_sdf: &SystemDescriptionFile) -> String {
+        let loc = loc_string(xml_sdf, self.text_pos);
+
+        match &self.ref_data {
+            CapMapRefData::Pd { pd, kind, .. } => {
+                format!("pd '{pd}'s {kind:?} at '{loc}'")
+            }
+        }
     }
 }
 
--- a/tool/microkit/src/sdf/cspace.rs
+++ b/tool/microkit/src/sdf/cspace.rs
@@ -4,6 +4,7 @@
 // SPDX-License-Identifier: BSD-2-Clause
 //
 
+use std::num::NonZeroU64;
 use std::rc::Rc;
 
 use super::consts::*;
@@ -12,11 +13,14 @@ use super::util::{
 };
 use super::{SdfLocation, SdfNode, SystemDescriptionFile};
 
+use crate::{sel4::Arch, Config};
+
 #[derive(Debug, PartialEq, Eq, Copy, Clone)]
 pub enum CapMapType {
     Tcb,
     Sc,
     VSpace,
+    ArmSmc,
 }
 
 #[derive(Debug, PartialEq, Eq)]
@@ -32,6 +36,7 @@ impl From<CapMapType> for CapMapRefDataPdKind {
             CapMapType::Tcb => CapMapRefDataPdKind::Tcb,
             CapMapType::Sc => CapMapRefDataPdKind::Sc,
             CapMapType::VSpace => CapMapRefDataPdKind::VSpace,
+            CapMapType::ArmSmc => unreachable!(),
         }
     }
 }
@@ -41,6 +46,9 @@ pub enum CapMapRefData {
     Pd {
         pd: Rc<str>,
         kind: CapMapRefDataPdKind,
+    },
+    ArmSmcFunction {
+        function_id: Option<NonZeroU64>,
     },
 }
 
@@ -63,13 +71,10 @@ pub struct CSpace {
 impl CapMap {
     fn from_xml(
         cap_type: CapMapType,
+        config: &Config,
         xml_sdf: &SystemDescriptionFile,
         node: &dyn SdfNode,
     ) -> Result<CapMap, String> {
-        // At the moment the four cap maps we support all have the 'pd' element,
-        // so we can include it here. When that stops being the case we will
-        // have to rework this a bit.
-
         let ref_data = match cap_type {
             CapMapType::Tcb | CapMapType::Sc | CapMapType::VSpace => {
                 check_attributes(xml_sdf, node, &["slot", "pd"])?;
@@ -80,6 +85,22 @@ impl CapMap {
                     pd,
                     kind: cap_type.into(),
                 }
+            }
+            CapMapType::ArmSmc => {
+                check_attributes(xml_sdf, node, &["slot", "function_id"])?;
+
+                if config.arch != Arch::Aarch64 {
+                    return Err(value_error(
+                        xml_sdf,
+                        node,
+                        "cap_smc is only supported on AArch64".to_string(),
+                    ));
+                }
+
+                let function_id: u64 = sdf_parse_required_attribute(xml_sdf, node, "function_id")?;
+                let function_id = NonZeroU64::new(function_id);
+
+                CapMapRefData::ArmSmcFunction { function_id }
             }
         };
 
@@ -116,12 +137,20 @@ impl CapMap {
             CapMapRefData::Pd { pd, kind, .. } => {
                 format!("pd '{pd}'s {kind:?} at '{loc}'")
             }
+            CapMapRefData::ArmSmcFunction { function_id } => {
+                format!(
+                    "smc cap for function '{} at '{}'",
+                    function_id.map(NonZeroU64::get).unwrap_or(0),
+                    loc
+                )
+            }
         }
     }
 }
 
 impl CSpace {
     pub(super) fn from_xml(
+        config: &Config,
         xml_sdf: &SystemDescriptionFile,
         node: &dyn SdfNode,
     ) -> Result<Self, String> {
@@ -131,9 +160,10 @@ impl CSpace {
 
         for child in node.children() {
             cap_maps.push(match child.tag_name() {
-                "cap_tcb" => CapMap::from_xml(CapMapType::Tcb, xml_sdf, &*child)?,
-                "cap_sc" => CapMap::from_xml(CapMapType::Sc, xml_sdf, &*child)?,
-                "cap_vspace" => CapMap::from_xml(CapMapType::VSpace, xml_sdf, &*child)?,
+                "cap_tcb" => CapMap::from_xml(CapMapType::Tcb, config, xml_sdf, &*child)?,
+                "cap_sc" => CapMap::from_xml(CapMapType::Sc, config, xml_sdf, &*child)?,
+                "cap_vspace" => CapMap::from_xml(CapMapType::VSpace, config, xml_sdf, &*child)?,
+                "cap_smc" => CapMap::from_xml(CapMapType::ArmSmc, config, xml_sdf, &*child)?,
                 child_name => {
                     let location = loc_string(xml_sdf, child.range().start);
                     if let Some(type_name) = child_name.strip_prefix("cap_") {
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -439,7 +439,7 @@ checksum = "94143f37725109f92c262ed2cf5e59bce7498c01bcc1502d7b9afe439a4e9f49"
 [[package]]
 name = "sel4"
 version = "0.1.0"
-source = "git+https://github.com/seL4/rust-sel4?rev=dbe6445d56059ed9a757e53c7137892aece1d179#dbe6445d56059ed9a757e53c7137892aece1d179"
+source = "git+https://github.com/seL4/rust-sel4?rev=7a56b4f1ee9084e2b5930b19afa47c080cd0e6cb#7a56b4f1ee9084e2b5930b19afa47c080cd0e6cb"
 dependencies = [
  "cfg-if",
  "sel4-config",
@@ -449,7 +449,7 @@ dependencies = [
 [[package]]
 name = "sel4-alloca"
 version = "0.1.0"
-source = "git+https://github.com/seL4/rust-sel4?rev=dbe6445d56059ed9a757e53c7137892aece1d179#dbe6445d56059ed9a757e53c7137892aece1d179"
+source = "git+https://github.com/seL4/rust-sel4?rev=7a56b4f1ee9084e2b5930b19afa47c080cd0e6cb#7a56b4f1ee9084e2b5930b19afa47c080cd0e6cb"
 dependencies = [
  "cfg-if",
 ]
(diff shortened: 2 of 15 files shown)


Indanz:
The current way of doing SMC is by setting an attribute `smc=true` on the PD description and calling `microkit_arm_smc_call()`.

The proposed way is using the recently added cap export mechanism and letting users call `seL4_ARM_SMC_Call()` directly on that cap.

I agree that this is a better API, as it is more generic and avoids the Microkit wrapper function.

However, the `cap_smc` part makes it SMC specific again. That's probably in line with the other exported caps, but I would have made it more generic and support all boot caps and have a generic badge thing instead of `function_id`, which requires SMC specific code again. I think I made the same comment on other caps exported via `cspace`.

I would not disallow `function_id="0"` for unbadged SMC caps, that is user policy. It can be handy for testing and if a process needs to do multiple SMC calls and one of them is destructive, then it doesn't add much to badge the caps.

tool/microkit/src/sdf/cspace.rs
```diff
@@ -37,15 +69,49 @@ pub struct CSpace {
...
+                // We enforce this restriction as a zero-badged SMC cap allows for
+                // *any* function ID, which gives (essentially) complete rights
+                // over performing SMC calls.
+                let Some(function_id) = NonZeroU64::new(function_id) else {
+                    return Err(value_error(
+                        xml_sdf,
+                        node,
+                        "An SMC function ID must be non-zero".to_string(),
+                    ));
+                };
```
dreamliner787-9: I agree with @Indanz, we should let the user decide what they want. It would be a good idea to make a note in the manual warning that function_id == 0 means granting access to all functions.

dreamliner787-9 (dismissed):

dreamliner787-9:
> That's probably in line with the other exported caps, but I would have made it more generic and support all boot caps and have a generic badge thing instead of function_id, which requires SMC specific code again.

If I understand correctly your proposal is something like this?
```xml
<system>
    <protection_domain name="arm_smc" priority="1">
        <program_image path="arm_smc.elf" />
        <cspace>
            <cap_initial type="smc" badge="0x84000000" slot="1"/>
        </cspace>
    </protection_domain>
</system>
```

Then `type` is just these:
image

This would be interesting and a step toward adding a dynamic escape hatch to Microkit,

tool/microkit/src/sdf/cspace.rs
midnightveil: I've removed this restriction and added a note to the manual.

midnightveil: Done.

dreamliner787-9 (approved):

midnightveil:
FYI, reason why I pinged you both is because this is somewhat closer to #362 than perhaps it should be, utilising the cspace element  which was merged previously without wider discussion.

I don't know if it's worth a proper RFC about whether we should make Microkit abstract away multiple function IDs as part of its API.

Indanz:
> If I understand correctly your proposal is something like this?
> 
> ```
> <system>
>     <protection_domain name="arm_smc" priority="1">
>         <program_image path="arm_smc.elf" />
>         <cspace>
>             <cap_initial type="smc" badge="0x84000000" slot="1"/>
>         </cspace>
>     </protection_domain>
> </system>
> ```

One step further, type can also be something else, like "tcb", "sc", but then it needs a source attribute to find the correct one, like`from="pd name"`. The equivalent for the above would be:

`<cap type="SMC" from="init" badge="0x84000000" slot="1"/>`

Of course, `from="init"` could be the default, so you can leave it out.

(And it's perfectly fine to have aliases like `function_id` so you can use that instead of `badge`. But I wouldn't try to restrict it to just SMC caps, that would add unnecessary code. It can also make it harder for other tools parsing the SDF, so maybe better not.)

> This would be interesting and a step toward adding a dynamic escape hatch to Microkit,

It's not just an escape hatch, it's a more generic way to copy caps.

The advantage is that cap type specific code is contained to the cap lookup part and that the rest of the code can be generic. That means you can add new cap types without changing the XML format.

The `type` attribute isn't used as a real type, it's just a way to specify which cap it is, because here's no other way to do that.

Indanz:
> FYI, reason why I pinged you both is because this is somewhat closer to #362 than perhaps it should be, utilising the cspace element which was merged previously without wider discussion.

I'm in strong favour of #362, but it's not up to me, what do actual users of Microkit think?

> I don't know if it's worth a proper RFC about whether we should make Microkit abstract away multiple function IDs as part of its API.

I wouldn't do that, that doesn't seem within scope of Microkit. Don't forget all you can count on is the SMC Calling Convention, you can't count on it following e.g. PSCI.

midnightveil:
> I wouldn't do that, that doesn't seem within scope of Microkit. Don't forget all you can count on is the SMC Calling Convention, you can't count on it following e.g. PSCI.

I'm specifically thinking about an equivalent to `camkes_get_smc_cap`.

Furao:
I understand wanting to generalize this and provide a very powerful tool for advanced seL4 developers, but I think it would be quite confusing for newer developers, especially if its the only option. 

```xml
<system>
    <protection_domain name="arm_smc" priority="1">
        <program_image path="arm_smc.elf" />
        <cspace>
            <cap_initial type="smc" badge="0x84000000" slot="1"/>
        </cspace>
    </protection_domain>
</system>
```

Presumably there will be some updates to the documentation, but a user needs to understand that they need a cap of type SMC for each smc call they want to allow and that the func_id is represented by the badge (which isn't actually mentioned in the seL4 User manual). Then they need to understand what value to use for slot. 

While the microkit API is somewhat inconsistent about what it provides, it has provided benefit by abstracting some of the more confusing parts of the microkernel.

Do you think we could support both approaches? 

The commented proposal provides the escape hatch which essentially lets you do whatever you need with initial_caps, but is likely only going to be used by advanced developers.

Then Microkit could still provide abstractions on commonly used functionality, such as SMC. As a user, I would hope for a more simple specification such as this:

```xml
<system>
    <protection_domain name="arm_smc" priority="1">
        <program_image path="arm_smc.elf" />
        <allowed_smc_calls>
            <smc_call func_id="0x84000000"/>
            <smc_call func_id="0x84000020"/>
            <smc_call func_id="0x8400003A"/>
        </allowed_smc_calls>
    </protection_domain>
</system>
```

You could also have a more obvious way to specify things like allowing all calls, without needing to understand how that is actually done:
```xml
<system>
    <protection_domain name="arm_smc" priority="1">
        <program_image path="arm_smc.elf" />
        <allowed_smc_calls allow_all="true" />
    </protection_domain>
</system>
```

midnightveil:
OK, thanks for the feedback. I think your proposal makes sense.

I'll propose that this generates the metadata used for a `microkit_get_smc_cap(function_id)` function.

Just to confirm, the other proposal was this, which is explicit about function IDs:

```xml
<protection_domain name="arm_smc">
    <!-- elided -->

    <cspace>
        <!-- An (ARM) SMC capability that permits SMC calls to 0x84000000 (PSCI_VERSION) -->
        <cap_smc slot="1" function_id="0x84000000" />
    </cspace>
</protection_domain>
```

Not cap_initial which does then use badge, which I agree is quite confusing.

midnightveil:
Something to note that if we do provide an `microkit_get_smc_cap` cap function, we need to actually implement it. CAmkES does code generation, which in the end makes [the code do O(n) checks against the values with immediates in the instruction stream - godbolt](https://godbolt.org/z/nz7ra9YGj). For Microkit we'd have to patch in values at runtime and probably do a linear search through an array, unless we want the complexity of a perfect hash function or a binary search. I don't think that's worth the tradeoff, but when it is a Microkit provided function, as opposed to `<cap_smc slot="1" function_id="0x84000000" />` this is policy Microkit imposes. However, we can actually impose policy with Microkit, the question is do we want to?

Indanz:
> Presumably there will be some updates to the documentation, but a user needs to understand that they need a cap of type SMC for each smc call they want to allow and that the func_id is represented by the badge (which isn't actually mentioned in the seL4 User manual).

Yes, SMC is currently not properly documented in the manual, other than the SMC syscall with too little information, see https://github.com/seL4/seL4/issues/1422.

> Do you think we could support both approaches?

The old, simple way can be retained easily enough.

> As a user, I would hope for a more simple specification such as this:
> 
> ```
> <system>
>     <protection_domain name="arm_smc" priority="1">
>         <program_image path="arm_smc.elf" />
>         <allowed_smc_calls>
>             <smc_call func_id="0x84000000"/>
>             <smc_call func_id="0x84000020"/>
>             <smc_call func_id="0x8400003A"/>
>         </allowed_smc_calls>
>     </protection_domain>
> </system>
> ```

Implementing this would be quite tricky in Microkit though: It would need to pass the func_id<->cap mapping to user space via e.g. bootinfo and use that to check if the call is allowed and find the right cap. Maybe this could be done via setvar or something, but that's not exactly user friendly either at this point.

I think it would be much simpler to expand the existing way by adding function_id support instead of just "true", like `smc="func_id"`. That's simple to implement and use for both the user and Microkit, while being backward compatible.

The complicated cspace way can be used for when that's not good enough.

But adding more and more very use case specific, tricky code to Microkit is exactly the kind of thing that I'm trying to push back against. Simple wrappers for simple things is somewhat fine, but now you're proposing to add complicated wrappers for complicated things, without supporting all use cases. That's the road to bloatware, subtle bugs and maintenance hell in 10 year time.

Maybe Microkit is already really close to mapping a read-only page per process with PD-specific metadata, then adding this wouldn't be too bad I suppose...

Furao:
> > Presumably there will be some updates to the documentation, but a user needs to understand that they need a cap of type SMC for each smc call they want to allow and that the func_id is represented by the badge (which isn't actually mentioned in the seL4 User manual).
> 
> Yes, SMC is currently not properly documented in the manual, other than the SMC syscall with too little information, see https://github.com/seL4/seL4/issues/1422.
> 
> > Do you think we could support both approaches?
> 
> The old, simple way can be retained easily enough.
> 
> > As a user, I would hope for a more simple specification such as this:
> > 
> > ```
> > <system>
> >     <protection_domain name="arm_smc" priority="1">
> >         <program_image path="arm_smc.elf" />
> >         <allowed_smc_calls>
> >             <smc_call func_id="0x84000000"/>
> >             <smc_call func_id="0x84000020"/>
> >             <smc_call func_id="0x8400003A"/>
> >         </allowed_smc_calls>
> >     </protection_domain>
> > </system>
> > ```
> 
> Implementing this would be quite tricky in Microkit though: It would need to pass the func_id<->cap mapping to user space via e.g. bootinfo and use that to check if the call is allowed and find the right cap. Maybe this could be done via setvar or something, but that's not exactly user friendly either at this point.
> 
> I think it would be much simpler to expand the existing way by adding function_id support instead of just "true", like `smc="func_id"`. That's simple to implement and use for both the user and Microkit, while being backward compatible.
> 

The most important part for me personally is that I need to support around 20 to 30 function IDs for each VM on the ZCU102.

> The complicated cspace way can be used for when that's not good enough.
> 
> But adding more and more very use case specific, tricky code to Microkit is exactly the kind of thing that I'm trying to push back against. Simple wrappers for simple things is somewhat fine, but now you're proposing to add complicated wrappers for complicated things, without supporting all use cases. That's the road to bloatware, subtle bugs and maintenance hell in 10 year time.

Yes, I see that argument. 

> 
> Maybe Microkit is already really close to mapping a read-only page per process with PD-specific metadata, then adding this wouldn't be too bad I suppose...

midnightveil:
> Maybe Microkit is already really close to mapping a read-only page per process with PD-specific metadata, then adding this wouldn't be too bad I suppose...

Yeah, we have a bunch of separate setvars already for things like this. I was considering something here: https://github.com/seL4/microkit/issues/422.

But we can do this with setvars.

---

I am leaning towards Microkit providing this feature, I think.

Indanz:
> The most important part for me personally is that I need to support around 20 to 30 function IDs for each VM on the ZCU102.

Any that do arbitrary memory reads or writes? Or just mostly harmless pin, clock and power control? That is, does restricting the SMC cap to specific function ids have real value to you?

> I am leaning towards Microkit providing this feature, I think.

If you do, what about the following implementation: Now that you have a multilevel level CSpace, just encode the function id in the guard value.

Function ids are 32-bit, you have 64 bits of CSpace, plenty to encode the function ID. Put the SMC cap in a one-slot big sub-CNode and you're done, stateless SMC cap lookup by function ID. You only need to reserve one top-level CNode slot for SMC badges. Then finding the right cap is a matter of concatenating the function id to the fixed top bits of SMC CSpace path. This is possible because the function id is passed in x0 of `seL4_ARM_SMCContext`.

Only consequence is that a task will fault with a cap fault if it tries to do an SMC call for a function id it doesn't have, but I'd argue that that's a good thing.

midnightveil:
> Function ids are 32-bit, you have 64 bits of CSpace, plenty to encode the function ID. Put the SMC cap in a one-slot big sub-CNode and you're done, stateless SMC cap lookup by function ID. You only need to reserve one top-level CNode slot for SMC badges. Then finding the right cap is a matter of concatenating the function id to the fixed top bits of SMC CSpace path. This is possible because the function id is passed in x0 of seL4_ARM_SMCContext.

Oooh, that's neat (but also somewhat cursed). I guess that won't work for the RISC-V OpenSBI variation, but we can deal with that when we get to it.

Indanz:
> Oooh, that's neat. I guess that won't work for the RISC-V OpenSBI variation, but we can deal with that when we get to it.

For SBI, EID is effectively 31 bits because it is signed and FID is 32 bits, but they all use them in sequential order starting at 0, so it's very unlikely all are needed. You could have two top-level slots for SBI caps:
- One with 1:1 encoding for e.g. 31-bit EID and 25-bit FID, depending on top CNode size.
- One which hashes the FID+EID. Chance of collision is miniscule for a few FID+EID pairs and >50-bits result. But honestly, I think you would get away without this fallback.

Furao:
> > The most important part for me personally is that I need to support around 20 to 30 function IDs for each VM on the ZCU102.
> 
> Any that do arbitrary memory reads or writes? Or just mostly harmless pin, clock and power control? That is, does restricting the SMC cap to specific function ids have real value to you?
> 

See the whole list here: https://xilinx-wiki.atlassian.net/wiki/spaces/A/pages/18842107/Arm+Trusted+Firmware?f=print#SMC-Handling

I mostly need things like device power, clocking, pinmux, etc for Linux device drivers in VMs. However there are ones I also want to block for all or most VMs, such as `PM_FPGA_LOAD`

midnightveil:
Perhaps a stupid question, but presumably your VMMs are trusted? Why don't they do this validation themselves?

Obviously in terms of system containment you want minimal privileges, but it should be easy enough to block a VM from doing SMC calls by just filtering the VMFaults?

Furao:
> Perhaps a stupid question, but presumably your VMMs are trusted? Why don't they do this validation themselves?
> 
> Obviously in terms of system containment you want minimal privileges, but it should be easy enough to block a VM from doing SMC calls by just filtering the VMFaults?

It is not a stupid question and certainly could work. 

However, there are no readily available VMMs that are verified and trusted. We are using libvmm on microkit and if there is a vulnerability in an interface, such as virtio or an emulated device, than that could potentially be used to modify what SMC calls the VMM allows the VM to make. We would avoid this issue completely by using the badged smc caps.

gernotheiser:
VMMs aren't trusted and I don't think they are anywhere near trustworthy, nor are they likely to get there anytime soon. 
The whole idea of the per-VM VMM design is that you don't have to trust the VMM

midnightveil:
Yes, that's what I meant by system containment - for minimal privileges obviously you want to restrict the capabilities themselves if possible (or talk to a minimal server).

I was just asking why you had done that even though the omly available mechanism was unbadged SMC; it seems like the answer is that you just hadn't done it that way, and for now assumed this wasn't an issue (?).

PR #590, opened 2026-08-11
tool(cleanup): use names not IDs for PDs

midnightveil:
As part of this, use Rc<str> instead of String/&str in places so that we can cheaply have owned-strings in HashMap keys without needing to allocate more and more string data.

This will be necessary for the Multikernel version of the Microkit as it will be implemented by building a system/spec for each core, which makes the indices invalid when embedded in e.g. channels or otherwise.

This also means that you can declare channels before PDs, instead they were forced to be placed in the SDF after the PDs due to the way the parsing worked.


tool/microkit/src/sdf.rs
```diff
@@ -242,7 +243,38 @@ pub fn parse(
...
+    if pds.len() > MAX_PDS {
+        return Err(format!(
+            "Error: too many protection domains ({}) defined. Maximum is {}.",
+            pds.len(),
+            MAX_PDS
+        ));
+    }
+
+    for pd in pds.iter() {
+        if pds.iter().filter(|x| pd.name == x.name).count() > 1 {
```
dreamliner787-9, line 264: Personally I would use a HashSet then return error on collision. This linear scan is a bit slow if the user have a lot of PDs when we get around to raising `MAX_PDS`. This is rather inconsequential in the grand scheme of things (capDL steps, then compression etc) so up to you if you want to change it.

dreamliner787-9 (approved):

tool/microkit/src/sdf.rs
midnightveil: Yeah, this could be refactored to be a bit smarter. This code previously existed.

midnightveil: I'll tack on an extra commit.

midnightveil: ... Actually, no, what I really should do is change `pd_flatten` to be a HashMap and then can move the error there, but that gets more complicated.

dreamliner787-9 (approved):

PR #592, opened 2026-08-13
tool(cleanup): add `sdf_parse_attribute()` functions to decrease duplication

### LINKED ISSUE #537: sdf parsing should be consistent with integer parsing
We have an `sdf_parse_number` function that sanely handles an error, but a bunch of the code only does:

```rs
                    let id = checked_lookup(xml_sdf, &child, "id")?
                        .parse::<i64>()
                        .unwrap();
```

Which does:

```
thread 'main' (2402178) panicked at tool/microkit/src/sdf.rs:973:30:
called `Result::unwrap()` on an `Err` value: ParseIntError { kind: InvalidDigit 
```

Which looks ugly.

midnightveil:
Please review commit by commit.

Fixes #537.

This removes a lot of the number/bool/etc parsing logic that exists in every single sdf file, instead separating out into helper functions.

Something that could be maybe done, but hasn't been, is adding support for parsing numbers in a valid range.


tool/microkit/src/sdf/util.rs
```diff
@@ -20,15 +81,7 @@ pub fn sdf_parse_number(s: &str, node: &dyn SdfNode) -> Result<u64, String> {
...
-        Ok(value) => Ok(value),
-        Err(err) => Err(format!(
-            "Error: failed to parse integer '{}' on element '{}': {}",
-            s,
-            node.tag_name(),
-            err
-        )),
-    }
+    T::from_str_radix(final_str, base)
 }
```
dreamliner787-9, line 125: If we only use unsigneds in Microkit, how about we do something like this:
```rust
fn parse_number<T: TryFrom<u64>>(src: &str, radix: u32) ...
```

tool/microkit/src/sdf/util.rs
```diff
@@ -47,6 +47,46 @@ pub fn sdf_required_attribute_as_number<T: IsNum>(
     })
 }
 
+/// Parse an 'attribute' of an `SdfNode` as a boolean.
+/// If the attribute does not exist, return an Optional value.
```
dreamliner787-9:
```Suggestion
/// If the attribute does not exist, return a None value, else, a Some value with the result.
```

I think this should be something like that?

tool/microkit/src/sdf/util.rs
midnightveil: How would that work? You're returning something, not taking something. You still need to know what type to parse as, and if you always parse u64 you then have to convert to another type (e.g. always take i64 to account for negatives).

tool/microkit/src/sdf/util.rs
midnightveil: That's just explaining what an Option is?

tool/microkit/src/sdf/util.rs
dreamliner787-9: > always take i64 to account for negatives

Ah yes I forgot about that.

tool/microkit/src/sdf/util.rs
dreamliner787-9: Yea but the original comment is not accurate from my understanding. 

"If the attribute does not exist, return an Optional value."

Implies that the function only return an Optional if the attribute does not exist. Or you can just remove this line since it is quite obvious what the function is doing.

dreamliner787-9:
From body of [tool(cleanup): impl ParseAttribute on PciDevice](https://github.com/seL4/microkit/pull/592/commits/74c6b494c78f34ea8f4eede2db9f893582e47283)

> Not sure of the best way to deal with IommuDeviceIdentifier
however in this framework. I think if I passed "config" to
parse then it might make more sense, but that means I need
an extra argument for all sdf_parse_attribute calls. Unless
it forms a part of the SystemDescription struct which
currently only contains a file, which might be a good idea.

How about we make the `Config` object a global? Since it is not mutated once created. In context of the wider "refactoring the tool into a library" work, I don't believe this will introduce any issues.

Alternatively, you can just not check the architecture in the parser function, and let the caller deal with architecture specific logic, as the SDF parser always know the target arch.

midnightveil:
I think making it part of the Xml_sdf arg makes sense. It's already passed everywhere including to these functions?

I don't like globals for that. Testing become hard, function no longer pure.

tool/microkit/src/sdf/util.rs
midnightveil: Right. Yeah I can see if I can make that worded better.

midnightveil: How the new wording?

dreamliner787-9: Yea sounds good

dreamliner787-9 (approved):

Issue #597, opened 2026-08-30
`duplicate irq` on x86 is keyed on the vector, not the hardware pin — message is misleading


SuperTarantaMQ:
# `duplicate irq` on x86 is keyed on the vector, not the hardware pin — message is misleading

**Component:** `tool/microkit`
**Version:** SDK 2.1.0, board `x86_64_generic`, config `debug`
**Impact:** a legitimate configuration (two devices, two different IOAPIC pins) is rejected with
an error that names neither the real constraint nor the fix.

## Summary

Two protection domains requesting **different** IOAPIC pins but the **same** `vector` are
rejected as duplicates. The error names the vector, but the word "irq" leads the reader to look
at the hardware line, which is not the thing that collides.

## Reproduction

Two PDs, pins 11 and 10 — genuinely different interrupt sources:

```xml
<system>
  <protection_domain name="probeA" priority="100" stack_size="0x4000">
    <program_image path="a.elf"/>
    <irq id="0" pin="11" vector="43"/>
  </protection_domain>
  <protection_domain name="probeB" priority="99" stack_size="0x4000">
    <program_image path="b.elf"/>
    <irq id="0" pin="10" vector="43"/>
  </protection_domain>
</system>
```

```
Error: duplicate irq: 43 in protection domain: 'probeB' @ min.system:7:3
```

Same result when both PDs use pin 11, which is the case the message reads as describing.

## Why this matters

`vector` uniqueness is **global to the system**, not per pin and not per PD. In a design with
several device-driving PDs, every interrupt must be assigned a distinct vector from `[0, 109]`.
That is a real and reasonable constraint, but it is not discoverable from the manual or from
this message, and it is the opposite of what a reader familiar with IOAPIC routing expects.

## Suggested fix

Name the constraint in the message, e.g.:

```
Error: vector 43 is already assigned (protection domain 'probeA'); x86 interrupt
vectors must be unique across the whole system, independently of the IOAPIC pin
@ min.system:7:3
```

Documenting the global-uniqueness rule in the manual's x86 `irq` section would help just as much.

midnightveil:
This is basically a duplicate of #598 and has the same root cause.

Please don't use AI to contribute.

PR #600, opened 2026-09-15
tool: detect duplicate I/O APIC IRQs and MSIs

### LINKED ISSUE #598: Two `irq` elements on the same IOAPIC pin are accepted silently; the second overwrites the first
# Two `irq` elements on the same IOAPIC pin are accepted silently; the second overwrites the first

**Component:** `tool/microkit`
**Version:** SDK 2.1.0, board `x86_64_generic`, config `debug`
**Impact:** a real configuration mistake produces no diagnostic at all. One of the two channels
never fires, and nothing says why.

## Summary

The tool rejects two IRQs sharing a `vector` (see the companion report), but accepts two IRQs
sharing a `pin` as long as their vectors differ — including within a single protection domain.
Since an IOAPIC redirection entry holds one destination vector per pin, the second declaration
overwrites the first. No warning is emitted.

## Reproduction

```xml
<system>
  <protection_domain name="probe" priority="100" stack_size="0x4000">
    <program_image path="probe.elf"/>
    <irq id="0" pin="11" vector="43"/>
    <irq id="1" pin="11" vector="44"/>
  </protection_domain>
</system>
```

```
$ microkit min.system --board x86_64_generic --config debug --search-path elf -o out.img -r report.txt
MICROKIT|CAPDL SPEC: number of root objects = 47, spec footprint = 6.79 KiB
$ echo $?
0
```

Both IRQ objects appear in the report:

```
# IRQ Details
	- IRQ: 'irq_43_probe'
		* Vector: 43
		* IOAPIC: 0
		* Pin: 11
	- IRQ: 'irq_44_probe'
		* Vector: 44
		* IOAPIC: 0
		* Pin: 11
```

## Suggested fix

Reject — or at minimum warn on — two `irq` elements targeting the same `(ioapic, pin)` pair,
the way the vector collision is already rejected. The check is symmetric to the existing one and
catches a mistake that is otherwise invisible until a channel silently stops working.

dreamliner787-9:
If you create multiple IRQ elements in the SDF with different vectors but the same I/O APIC or PCI device source, the kernel will silently overwrite all of those with the element that is processed last by the capDL initialiser at runtime.

We should guard against this footgun by reporting an error in the tool.

Closes https://github.com/seL4/microkit/issues/598.

Related: https://github.com/seL4/seL4/issues/1748.


tool/microkit/tests/test.rs
```diff
@@ -562,6 +562,15 @@ mod protection_domain {
         )
     }
 
+    #[test]
```
midnightveil, line 565: Could you add a positive (success) test for https://github.com/seL4/microkit/issues/597?

tool/microkit/src/sdf/pci.rs
```diff
@@ -22,6 +23,21 @@ impl Deref for PciDevice {
     }
 }
 
+impl Ord for PciDevice {
+    fn cmp(&self, other: &Self) -> Ordering {
```
midnightveil, line 27: Is this needed because the capDL tool doesn't have Cmp on `object::PCIDevice`?

tool/microkit/src/sdf.rs
```diff
@@ -372,6 +372,46 @@ pub fn parse(
...
+            for sysirq in &pd.irqs {
+                if let SysIrqKind::IOAPIC { ioapic, pin, .. } = sysirq.kind {
+                    if all_ioapic_irqs.contains(&(ioapic, pin)) {
+                        return Err(format!(
+                            "Error: duplicate I/O APIC IRQ chip {}, pin {} in protection domain: '{}' @ {}:{}:{}",
+                            ioapic,
+                            pin,
+                            pd.name,
+                            filename.display(),
+                            pd.text_pos.unwrap().row,
```
midnightveil: `loc_string()`? ditto other

tool/microkit/src/sdf/pci.rs
dreamliner787-9: Yea that's right

tool/microkit/src/sdf.rs
dreamliner787-9: Ah right, I forgot to git pull before working on this and ended up reusing the old style

tool/microkit/src/sdf/pci.rs
midnightveil: Don't want to block this PR, but should we add a comment and fix it upstream?

dreamliner787-9: We could, but I'm not sure if there are negative implications if we add derives that the upstream code themselves don't use.

midnightveil: Nah, it will be dead-code eliminated quite easily.

tool/microkit/tests/test.rs
dreamliner787-9: Addressed offline, we already have tests for duplicate vector, and the issue is just complaining about the error phrasing.

terryzbai:
One I/O APIC IRQ number can be only mapped to one vector, but it is possible to share same I/O APIC IRQ number among multiple PCI devices, which means the same IRQ might need to be mapped to at least two driver PDs.

dreamliner787-9:
> One I/O APIC IRQ number can be only mapped to one vector, but it is possible to share same I/O APIC IRQ number among multiple PCI devices

Sure

> which means the same IRQ might need to be mapped to at least two driver PDs.

I don't think the kernel allow you to do this, `seL4_IRQHandler_SetNotification` section in the manual doesn't specify that you can bind > 1 notification to an interrupt. So you would have to do some Ntfn cap copy or something like that which I don't think will work in Microkit. Plus this problem can be entirely avoided by using MSI in the driver

midnightveil:
Even if the kernel allowed it, you still can't tell which PD the interrupt is for, you need device knowledge for that, which requires either broadcast or some driver-level knowledge: so to replicate it on seL4 you need an intermediary PD. 

(Possibly we should document this pattern somewhere?)

dreamliner787-9:
> (Possibly we should document this pattern somewhere?)

We could make an example?

midnightveil (approved):

PR #603, opened 2026-09-15
Only provide minimal rights for ntfn/ep caps

midnightveil:
This was previously a FIXME in the code before the capDL initialiser migration, and providing all capabilities with full rights allows for all sorts of system rights escapes.


tool/microkit/src/capdl/builder.rs
```diff
@@ -105,6 +105,61 @@ const PD_SCHEDCONTEXT_EXTRA_SIZE_BITS: u64 = PD_SCHEDCONTEXT_EXTRA_SIZE.ilog2()
...
+    read: true,
+    write: false,
+    grant: false,
+    grant_reply: false,
+};
+
+/// Minimal rights for a fault handler *of a TCB*.
+/// Conditions for 'validFaultHandler()' in seL4 source:
+/// We need 'send'/'capCanSend' and a grant or grantreply.
+/// FIXME: why need a grant?
```
midnightveil, line 128: This confuses me.

lsf37: These rights would be consistent with the rights a client needs when it does a Call to a server: it needs send, and it needs at least grantreply if wants to be able to receive a reply (or grant which would subsume grantreply).

If this is the fault handler cap that is stored inside the TCB of the faulting thread, then that makes sense - it's the cap that is used to generate the fault Call and should have the same requirements as a normal Call. Since replying is the mechanism by which the fault handler can restart threads, grantreply is pretty much required.

The fault handler on the other side should only need receive rights and the cap to its reply object, which replygrant on the Call side then authorises to be used for this specific transaction.

midnightveil: Right, I guess that means RIGHTS_EP_SEND needs grant_reply too.

I don't really understand why (see private discussion) but it is at least consistent.

midnightveil: I've updated to make it `RIGHTS_EP_CALL`, and made the faultHandler equal to Call. Hopefully that should be explicit. I still don't quite understand why seL4 uses GrantReply for this, but nevertheless.

dreamliner787-9 (approved):

PR #605, opened 2026-09-17
tool: fix broken PD hierarchy

### LINKED ISSUE #604: Hierarchy Example (and probably hierarchies themselves) are broken
```
thread 'main' (1900678) panicked at tool/microkit/src/capdl/builder.rs:767:58:
no entry found for key
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

dreamliner787-9:
Prior to commit https://github.com/seL4/microkit/commit/c67227fc3d1b9dfd0c1624d53a0222022a8d9d3c, PDs are fed into step 3 of build_capdl_spec()
as a flat vector in topological order. That is, a parent ID always
appears before its children PDs. However, commit https://github.com/seL4/microkit/commit/c67227fc3d1b9dfd0c1624d53a0222022a8d9d3c changed the Vec
into a BTreeMap, so the ordering is now dictated by the Rust data
structure implementation.

Thus it is possible that a child will now be created before its parent.
Causing the code to crash when it tries to find the parent's fault
endpoint.

This is fixed by moving the fault endpoint binding step into a separate
"post processing" step that happens after all the PDs are created.

Closes https://github.com/seL4/microkit/issues/604


tool/microkit/src/capdl/builder.rs
```diff
@@ -1153,7 +1119,59 @@ pub fn build_capdl_spec(
...
+            pd_tcb.slots.push(capdl_util_make_cte(
+                TcbBoundSlot::FaultEp as u32,
+                pd_fault_ep_cap,
+            ));
+            // Stylistic purposes only
+            pd_tcb.slots.sort_by_key(|cte| usize::from(cte.slot));
+        } else {
+            unreachable!("internal bug: build_capdl_spec() got a non TCB object ID when trying to bind the fault Endpoint.");
+        }
+    }
```
midnightveil, line 1224: Where did this come from? It wasn't in the previous code?

dreamliner787-9: Previously in step 3-5 we just put the CTE inside `caps_to_bind_to_tcb` and then everything happen for us in step 3-13. But now this step is in a different scope we need to re-fetch the TCB and insert the fault EP CTE

midnightveil: Right. That's somewhat ugly :/.

But alas...

midnightveil (approved):
Approving as a hotfix for the issue. I think I'd like to restructure this a bit more sensibly but for now let's just fix the issue.

PR #607, opened 2026-09-18
2.3.1

dreamliner787-9:
Add release notes for 2.3.1 minor release.


midnightveil (approved):
Happy with release notes.

I think we need an extra commit updating the VERSION, though, right? And its appearance in the manual and other locations where there is a Microkit version?

dreamliner787-9:
> Happy with release notes.
> 
> I think we need an extra commit updating the VERSION, though, right? And its appearance in the manual and other locations where there is a Microkit version?

Oh true!

